PR #22, opened 2016-11-05
Fix small typo

skade:


casey:
Thanks for spotting and fixing this!

PR #54, opened 2016-11-12
Colors and wide characters

casey:
Adds colored output, as well as fixing an error where wide characters such as tabs and chinese characters that take up more than one space.

PRing them together, since they're a little entangled.


casey:
💛💚💙💜❤

Issue #118, opened 2016-11-15
Allow variadic recipes


casey:
Would look like this:

```make
foo *stuff:
  @echo {{stuff}}
```

```sh
$ just foo a     b    c   d
a b c d
```

Using `+` may be more accurate, since I don't think I want to allow zero arguments, unless a default is supplied.

PR #127, opened 2016-11-18
Add variadic parameters

casey:
Recipes may now have a final variadic parameter:

```make
foo bar+:
  @echo {{bar}}
```

Variadic parameters accept one or more arguments, and expand to a string containing those arguments separated by spaces:

```sh
$ just foo a b c d e
a b c d e
```

I elected to accept one or more arguments instead of zero or more arguments since unexpectedly empty arguments can sometimes be dangerous. 

```make
clean dir:
  rm -rf {{dir}}/bin
```

If `dir` is empty in the above recipe, you'll delete `/bin`, which is probably not what was intended.


casey:
Fixed #118

PR #173, opened 2017-04-21
Fix integration tests when system `sh` is recent bash

casey:
My main machine is now a Fedora box. OS X ships with an ancient
version of bash installed as `sh` due to the great GPL purge[0],
so my `sh` is now much more recent.

This broke some tests which were relying on some non-portable
details of the echo builtin. Switched to using printf, which
should be more consistent across platforms.

[0] http://meta.ath0.com/2012/02/05/apples-great-gpl-purge/

### DIFF
--- a/src/integration.rs
+++ b/src/integration.rs
@@ -806,14 +806,16 @@ fn raw_string() {
   integration_test(
     &[],
     r#"
-export exported_variable = '\\\\\\"'
+export exported_variable = '\\\\\\"\n'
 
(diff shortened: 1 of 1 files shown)


casey:
Travis y u do dis?

Travis uses debian, which comes with the much more reasonable `dash` as the system `sh`, whose printf built-in probably differs in behavior from bash.

casey:
Fixed in another commit.

PR #180, opened 2017-04-22
Run integration tests on multiple shells

casey:
To make sure that I don't break tests for shells other than my
dev-box's `sh` and the `sh` on travis, each integration test case
now runs using `sh`, `dash`, and `bash.


casey:
This was merged by accident and then un-merged.

PR #207, opened 2017-06-04
Enable travis tests on FreeBSD

casey:
Will it blend?


casey:
Welp, mysterious build error, I give up.

PR #217, opened 2017-08-18
Implemented parser.rs (Semi-fixes #161)

### LINKED ISSUE #161: Remove dependency on `sh`
`sh` is `just`'s only significant dependency. I've been thinking for a long time that it would be nice to get rid of it.

I think the best way would be to write a shell library in rust. It would probably be easiest to make it a drop in replacement for [dash](en.wikipedia.org/wiki/Almquist_shell). It could be integrated as a library to avoid having to fork a new binary to run each command.

Some benefits of this would be:

- Faster execution time, since there is no need to fork a new binary to run each command
- Works the same independent of the system-provided `sh`
- Support windows
- Can tokenize and parse recipes ahead of time, and thus emit errors regarding malformed recipes without having to actually run them
- Currently, variable assignment and `cd`ing in recipes don't work as one would expect, since each line is executed by a new shell. If instead some kind of `sh` library were used, these could be made to work with the expected semantics
- Support pipefail
- Can better integrate just variables and shell variables, perhaps allowing both to be accessed with `$NAME`
- Can check for use of undefined shell variables
- Can provide better error messages
- Since the `sh` library will be doing the parsing, multi-line constructs would not need to be backslash escaped

Additionally, I think that such a project would have a lot of utility outside of just. `sh` is one of the most important programs on a system, and writing it in a more secure language like rust would have a lot of benefits. It might be possible to add a lot of aggressive optimizations, features which aren't feasible in C. Also, although the base library would have a very small feature set, and would be a drop in replacement for `dash`, with the ability to add plugins, it could be the starting point for a more feature rich shell.

One difficult constraint is that we currently depend on Cygwin on Windows. Cygwin binaries contain a system call translation layer, which allows them to use unix paths and a unix userland on windows. If we wrote our own shell we would have to figure out how to continue to provide this functionality.

jD91mZM2:
This parser isn't yet hooked up to anything.
It also isn't *fully* compatible with bash.

It is - however - pretty damn similar.
All the tests are confirmed working, A.K.A.

`echo "\"HELLO\", WORLD"lol $TEST. ${LOL!}lol`
`cat < in > out`
`echo '<'`

All of these work as you'd expect them to.

**note**: The first one says `${LOL!}lol`, which converts to `Token::Env("LOL!"), Token::Text("lol")`.
This isn't actually how bash would do it probably, since that doesn't actually allow non-[a-zA-Z0-9] characters even if you specify them inside `{` and `}`. This one does.

**note 2**: This does not throw any errors (yet). Unquoted strings aren't handled. That could easily be done though, if you let me know how you want to handle them.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1,0 +1,154 @@
+#[derive(Debug, Eq, PartialEq)]
+pub enum Token {
+	Text(String),
+	Env(String),
+	Space,
+	Pipe,
+	RedirectIn,
+	RedirectOut
+}
+
+impl Token {
+	fn is_empty(&self) -> bool {
+		match *self {
+			Token::Text(ref buffer) => buffer.is_empty(),
+			Token::Env(ref buffer) => buffer.is_empty(),
+			_ => false,
+		}
+	}
+}
+
+pub fn parse(input: &str) -> Vec<Token> {
+	let mut tokens = Vec::new();
+	let mut buffer = Token::Text(String::new());
+	let mut escape = false;
+	let mut quote  = None;
+
+	'chars:
+	for c in input.chars() {
+		loop {
+			let mut repeat = false;
+			let mut flush  = false;
+			let mut space  = false;
+			let mut newbuffer = None;
+			match buffer {
+				Token::Text(ref mut buffer) => {
+					if escape {
+						buffer.push(c);
+						escape = false;
+						continue 'chars;
+					}
+					match c {
+						'\\' => escape = true,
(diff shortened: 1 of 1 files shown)


jD91mZM2:
Because GitHub didn't recognize #161 in the title, I'm just going to mention it here so that issue gets a "mentioned in #217" automatic message.

EDIT: There we go
image

casey:
Thanks for this! But closing, since the `sh` dependency will need to be replaced by something with the same features.

jD91mZM2:
I thought the pull request could be expanded over-time... Whatever

PR #218, opened 2017-08-18
Retire warn! in favor of eprintln!

casey:


src/app.rs
```diff
@@ -8,18 +8,10 @@ use std::collections::BTreeMap;
...
-  ($($arg:tt)*) => {{
-    extern crate std;
-    use std::io::prelude::*;
-    let _ = writeln!(&mut std::io::stderr(), $($arg)*);
-  }};
-}
-
 macro_rules! die {
   ($($arg:tt)*) => {{
     extern crate std;
```
bb010g, line 13: Why do you need the `extern crate std` if you're already relying on `::std::process` being in scope?

casey: Macros don't have lexical scope like functions, so even if something is in scope where they're defined, it might not be in scope where it's called.

Take that with a grain of salt though, I'm not sure if I was doing it correctly when I did that or just cargo-culting.

PR #222, opened 2017-09-09
Add short options for justfile and working-directory

chrisvittal:
Adds `-f` and `-C` as short options for `--justfile` and `--working-directory` respectively.

I was surprised when these were not options to `just`. It was trivial to add them, so I did.


casey:
I'm definitely on board with `-f`. I'm not sure about `-C` though. I know that make uses `-C`, I guess since it's like `cd`, but that feels a little esoteric if one isn't already familiar with make.

Alternatives could be `-r`, for root directory; `-c`, for change directory, but lowercase so easier to type; or `-d`, for directory. I think I'm leaning towards `-d`.

Thoughts?

chrisvittal:
It's not just make, it's also tar, and a few others. However, just because I expect it does not make it correct, `-d` is the most sensible to me too.

casey:
Merged! Thank you!

PR #253, opened 2017-11-18
 Do not evaluate backticks in assignments during dry runs

aoeu:


casey:
Merged! Thanks, Travis! ^__^

PR #277, opened 2017-12-02
Add functions

### LINKED ISSUE #263: OS and ARCH
It would be awesome if I could get variables that define `OS_NAME=Windows`, `OS_VERSION_MAJOR=10`, and `ARCH=x86_64` on Windows 10 64-bit when launching external apps and making builds. Or to have `just` load `Justfile.win10-x86_64` over `Justfile.win10` over `Justfile.win` over `Justfile` in such a situation. This sort of feature is the main thing I've been missing in any build/task/automation tool. I could set environment variables for all of this. But that is not ideal when the OS changes or files are copied to another system by someone else that is 32-bit vs 64-bit, etc. This is the main reason I need #262 is for managing differences between `os` and `arch` versions in my `Justfile`. A great start would be those types of vars based on what `just` was compiled for. Maybe `JUST_OS=Windows` and `JUST_ARCH=x86_64` being available while running the `Justfile` would be enough.

casey:
– Parse unary (no-argument) functions
– Add functions for detecting the current os, arch, and os family, according to rustc's `cfg` attributes

Fixes #263


skellock:
Awesome job!

casey:
Thanks! ^__^

PR #281, opened 2017-12-04
📝 VS Code Extension

skellock:
Hey. Thanks for `just`. I love it.

I've created an extension for VS Code which adds syntax definitions.

just-demo

I'll be adding a few more things shortly (improved syntax and a task launcher) shortly.  Still figuring out how to make these plugins with VS Code. 🤷‍♂️


casey:
So glad you like just!

Super sweet! This is totally rad. Thank you so much!

PR #284, opened 2017-12-11
Explain binary installs first, remove git install

gregwebs:
I saw someone get confused and think they need to install via cargo.

Not sure what installing git has to do with this project?


casey:
Thanks for the pull request!

Just depends on `sh` for running commands, and those git distributions the easiest ways to get it. I've reworded that part to make it clearer why they're mentioned.

PR #303, opened 2018-02-16
Fix `just` shows escape code instead of colored text on Windows 10 Console

gaeulbyul:
image

On Windows 10, `just` shows escape code instead of colored text.

According to [ansi_term crate](https://crates.io/crates/ansi_term)'s README, Application must enable ANSI support for print colored text on Windows 10 Console.

> Note for Windows 10 users: On Windows 10, the application must enable ANSI support first:
> ```rust
> let enabled = ansi_term::enable_ansi_support();
> ```

This commit add `ansi_term::enable_ansi_support()` into `active` function.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -125,11 +125,15 @@ impl Color {
   }
 
   pub fn active(&self) -> bool {
-    match self.use_color {
+    let mut activated = match self.use_color {
       UseColor::Always => true,
       UseColor::Never  => false,
       UseColor::Auto   => self.atty,
+    };
+    if cfg!(windows) && activated {
+      activated = ansi_term::enable_ansi_support().is_ok();
     }
+    activated
   }
 
   pub fn paint<'a>(&self, text: &'a str) -> ANSIGenericString<'a, str> {
--- a/src/color.rs
+++ b/src/color.rs
@@ -125,13 +125,14 @@ impl Color {
   }
 
   pub fn active(&self) -> bool {
-    let mut activated = match self.use_color {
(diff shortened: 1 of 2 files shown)


casey:
Thank you very much for the PR! I moved the activation code into the run function, so that it's only ever run once.

I'm eager to support Windows users as best I can, so please don't hesitate to report issues as you encounter them.

casey:
I have pushed a new release with the fix: https://github.com/casey/just/releases/tag/v0.3.8

The release will be updated with binaries when they are built on Travis and Appveyor.

gaeulbyul:
image

Hello, I tried released version (v0.3.8) but same issue reappeared again. but, if I input `just --list --color=always`, `just` shows color correctly. Other command shows escape code.

casey:
Hmm, very strange. Was it an issue with the changes that I made to your code? Did your version work consistently? I don't have access to a Windows 10 machine at the moment, so it's hard for me to test.

gaeulbyul:
Yes, my version works well. I don't know why it broke again... 😕

gaeulbyul:
Hello, I tried your win10 branch. `just --list` (without --always option) shows color correctly. but error meesage (`just fail`) still shows escape code.
However, if I put one more `enable_ansi_support()` after `justfile.run()`([src/run.rs:358](https://github.com/casey/just/blob/274c6842f9f1d818f9db52bb99f710021ec5908e/src/run.rs#L358)), error message shows color correctly. putting it before `justfile.run()` does not effect.

PR #317, opened 2018-05-07
Always enable ansi support on windows

casey:
Give up on being clever and just unconditionally enable ansi support on windows. Should finally fix the issue that @gaeulbyul reported in #303.

@gaeulbyul, sorry for taking forever to fix this!


casey:
@gaeulbyul: I've published the new version to crates.io, and windows prebuilts will be here soon: https://github.com/casey/just/releases/tag/v0.3.11

PR #323, opened 2018-06-16
Implement invocation_directory function (fix #312)

### LINKED ISSUE #312: Add function to return the directory `just` was invoked from
I would love to see the option to run the commands in the current directory instead of always from the directory where the justfile is located.
For example, I'm using just in a nested Maven project and sometimes I just want to run the commands on a submodule instead of the whole project, but the justfile is at the top level.
This could be an option, maybe by prefixing the command with a dot (either in the justfile itself or from the shell).

joshuawarner32:
`invocation_directory` retrieves the current directory that `just` was invoked from, before the cwd (current working directory) was changed to the location of the justfile.

The name's not perfect, but I'm not one for bikeshedding.

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -8,6 +8,7 @@ lazy_static! {
     ("os_family",          Function::Nullary(os_family         )),
     ("env_var",            Function::Unary  (env_var           )),
     ("env_var_or_default", Function::Binary (env_var_or_default)),
+    ("invocation_directory", Function::Nullary(invocation_directory)),
   ].into_iter().collect();
 }
 
@@ -29,6 +30,7 @@ impl Function {
 }
 
 pub struct FunctionContext<'a> {
+  pub invocation_directory: Option<&'a str>,
   pub dotenv: &'a Map<String, String>,
 }
 
@@ -90,6 +92,14 @@ pub fn os(_context: &FunctionContext) -> Result<String, String> {
 
 pub fn os_family(_context: &FunctionContext) -> Result<String, String> {
   Ok(target::os_family().to_string())
+}
+
+pub fn invocation_directory(context: &FunctionContext) -> Result<String, String> {
+  if let Some(dir) = context.invocation_directory {
+    Ok(dir.to_string())
+  } else {
+    Err(String::from("error getting invocation directory"))
+  }
 }
 
 pub fn env_var(context: &FunctionContext, key: &str) -> Result<String, String> {
--- a/src/function.rs
+++ b/src/function.rs
@@ -30,7 +30,7 @@ impl Function {
 }
 
 pub struct FunctionContext<'a> {
-  pub invocation_directory: Option<&'a str>,
+  pub invocation_directory: &'a Result<String, String>,
   pub dotenv: &'a Map<String, String>,
 }
 
@@ -95,11 +95,7 @@ pub fn os_family(_context: &FunctionContext) -> Result<String, String> {
 }
 
 pub fn invocation_directory(context: &FunctionContext) -> Result<String, String> {
-  if let Some(dir) = context.invocation_directory {
-    Ok(dir.to_string())
-  } else {
-    Err(String::from("error getting invocation directory"))
-  }
+  context.invocation_directory.clone()
 }
 
 pub fn env_var(context: &FunctionContext, key: &str) -> Result<String, String> {
--- a/src/function.rs
+++ b/src/function.rs
@@ -1,5 +1,9 @@
+use std::path::PathBuf;
+
 use common::*;
 use target;
+
+use platform::{Platform, PlatformInterface};
 
 lazy_static! {
   static ref FUNCTIONS: Map<&'static str, Function> = vec![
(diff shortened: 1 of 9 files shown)


src/function.rs
```diff
@@ -92,6 +94,14 @@ pub fn os_family(_context: &FunctionContext) -> Result<String, String> {
   Ok(target::os_family().to_string())
 }
 
+pub fn invocation_directory(context: &FunctionContext) -> Result<String, String> {
+  if let Some(dir) = context.invocation_directory {
+    Ok(dir.to_string())
+  } else {
+    Err(String::from("error getting invocation directory"))
```
joshuawarner32: I'm not super happy with this error message right now.  The primary reason for this error would be an inability to decode the utf-8 of the original working directory, but it's not clear.  Feedback/thoughts welcome.

casey: I think that a really explicit error message is good, perhaps "invocation directory path not UTF-8, `invocation-directory` cannot be used"

src/run.rs
```diff
@@ -47,6 +47,11 @@ pub fn run() {
   #[cfg(windows)]
   enable_ansi_support().ok();
 
+  let invocation_directory = match env::current_dir() {
+    Ok(pathbuf) => pathbuf,
```
casey: Let's do `pathbuf.to_str().map(str::to_string)` and then use `Option<String>` for the type of the `invocation_directory` value.

src/function.rs
casey: Also, you can use [Option::ok_or_else](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else) to turn `None` into an error.

casey (changes requested):
Thank you very much for the PR! I added some comments inline.

We'll need a test for this. It should invoke the just binary with a recipe that prints out `invocation_directory()`, and then asserts that it prints the correct directory.

Since it needs to call the binary directly, it should be an integration test, so something like `tests/invocation_directory.rs`. If you look in `tests/integration.rs`, you'll see how to get the binary and path to the binary and invoke it.

Let me know if you have any questions!

src/function.rs
joshuawarner32: Yeah - I don't disagree.  The only thing that gives me pause is that _where the error occurred_ is pretty far (in code) from where this string is - which makes me think that they could fall out of sync.  For example, one day in the future there could be a different reason that `invocation_directory` can't be used, and it would be easy to forget to update this message.

What do you think about just having `invocation_directory` be passed along as a `Result<String, String>`?  That way, the error string is constructed close to the error itself.

Alternatively, `just` could refuse to run when invoked from a non-utf8 path, but that seems rather draconian.

joshuawarner32:
Hmm... I'm at a loss with what to do with the windows test.  It seems to be a fundamental mismatch between windows paths and cygwin (or msys?) paths.  Here are a few options I can think of:
* Bow to the cygwin gods, and use conditional compilation (or just normal `if`s) to conform to the style of paths that `sh` will expect on that system.
* Only enable the test for *nix systems, where pathing like this is a little more sane
* Pick a different shell to run this test under, that's going to deal better with windows paths.  Perhaps `cmd.exe`?

casey:
Hmmm, there's a utility called `cygpath` which `just` already uses to generate shebang paths on windows. (See `src/platform.rs:55`.) Could we use it here to translate the windows-style path to the cygwin path that the test will output, when we're running on windows?

tests/invocation_directory.rs
```diff
@@ -0,0 +1,65 @@
...
+fn test_invocation_directory() {
+  let tmp = TempDir::new("just-integration")
+    .unwrap_or_else(
+      |err| panic!("integration test: failed to create temporary directory: {}", err));
+
+  let mut justfile_path = tmp.path().to_path_buf();
+  justfile_path.push("justfile");
+  brev::dump(justfile_path, "default:\n @echo {{invocation_directory()}}");
+
+  let mut dotenv_path = tmp.path().to_path_buf();
```
casey: I think the dotenv stuff can be left out of this test.

joshuawarner32:
Apologies for the CI noise on this PR; I don't have a great place to test on windows so I've been using CI.

I pivoted to providing the cygwin "unix" path on windows (rather than the native `C:\...` style path), which makes the tests pass and generally will make `sh` happier I think.

I think there's an argument to be made to expand `invocation_directory()` into the native path, or perhaps an _escaped_ native path (i.e. `C:\\...`, doubling all the backslashes - which would _also_ have made the tests pass in this case).  In the future, that could be specified as e.g. `invocation_directory("native")` or `invocation_directory("escaped")`.  However, I think that's out of scope for this PR, and definitely worth getting feedback from people who use `just` on Windows before heading that direction.

Anyway, do you have further thoughts/comments @casey?

casey (approved):

casey:
That seems like the right thing to do, on windows recipes expect to be run in cygwin, so the unix path will be more immediately usable. I added the `cd` back into the test, as an extra test that the directory is in fact usable within recipes.

Nice work, and thanks for tackling what turned out to be much trickery issue than it first appeared!

PR #329, opened 2018-07-05
Provide prebuilt binary install command

casey:
@cledoux mentioned wanting an easy shell command to install `just` on a new machine, and I finally got around to seeing how much work it would be. It turns out that since binaries are built with the excellent trust CI templates, it's already done for us.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -52,6 +52,13 @@ On Windows, `just` works with the `sh` provided by https://git-scm.com[Git for W
 
 Pre-built binaries for Linux, MacOS, and Windows can be found on https://github.com/casey/just/releases[the releases page].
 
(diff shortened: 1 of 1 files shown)


cledoux:
This is awesome!

Just a note, I had to manually provide the `--target` option to get it to work.  The target was defaulting to `x86_64-unknown-linux-gnu`, but there is no `x86_64-unknown-linux-gnu` release. The release that exists is `x86_64-unknown-linux-musl`.  Note the change from `gnu` to `musl` (took me a bit to notice this difference myself).

This might be worth mentioning in the documentation.

cledoux:
Thanks for taking the time to look into this!

casey:
Sweet, thanks for trying it, and thanks for the note! I'll add that to the docs.

Issue #351, opened 2018-09-14
Distribute on Scoop


casey:
[scoop](https://scoop.sh) looks like a good option for easily getting `just` to windows users. I haven't used it, but it looks similar to homebrew and is used by [gutenberg](https://github.com/Keats/gutenberg) to distribute their binaries.

bugabinga:
A starting point for the impatient ;) 

`just.json`
```json
{
    "version": "0.3.12",
    "license": "CC0-1.0",
    "url": "https://github.com/casey/just/releases/download/v0.3.12/just-v0.3.12-x86_64-pc-windows-gnu.zip",
    "homepage": "https://github.com/casey/just",
    "hash": "E83E12E48AAE9EB004AB3059515B1A3EF7DC1A1E93E62AFAB54187A947D29551",
    "bin": "just.exe"
}
```

bugabinga:
I´ve refined the manifest a little bit since then.

Turns out, _scoop_ supports (semi-)automatically tracking releases on GitHub.
I have tested an auto-upgrade from 1.3.10 to 1.3.12.

https://github.com/casey/scoop/pull/1

```
{
    "version": "0.3.12",
    "license": "CC0-1.0",
    "homepage": "https://github.com/casey/just",
    "bin": "just.exe",
    "checkver": "github",
    "architecture": {
        "64bit": {
            "url": "https://github.com/casey/just/releases/download/v0.3.12/just-v0.3.12-x86_64-pc-windows-gnu.zip",
            "hash": "e83e12e48aae9eb004ab3059515b1a3ef7dc1a1e93e62afab54187a947d29551"
        }
    },
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": "https://github.com/casey/just/releases/download/v$version/just-v$version-x86_64-pc-windows-gnu.zip"
            }
        }
    }
}
```

casey:
Fixed by the excellent @matthewjberger in #359!

PR #359, opened 2018-10-02
Update readme with scoop installation instructions.

matthewjberger:
I recently added a manifest for installing just with the scoop package manager: https://github.com/lukesampson/scoop/pull/2638.

I updated the installation instructions for just with a link to scoop and a command for installing just on windows.


casey:
Awesome, thank you so much! Another step in the `just` world domination master plan ^_^

matthewjberger:
You're welcome! :smile: I would have tagged #351 but I actually wasn't aware of it until a few minutes ago

PR #360, opened 2018-10-11
Fix another-recipe output in README

uu1t:


casey:
Awesome, thank you very much!

Issue #156, opened 2017-02-27
Fuzz test just


casey:
Probably using https://github.com/rust-fuzz/cargo-fuzz

Also add convenient recipe to fuzz test

casey:
An interesting twist to this would be to do grammar-directed fuzz testing, where we generate text which should be valid according to the grammar.

casey:
Fixed in #362. Thanks Geoff!

PR #362, opened 2018-10-13
Fuzz testing

RadicalZephyr:
Taking a stab at #156 here.  This is my first foray into fuzz testing, but cargo-fuzz was pretty straightforward to use.

In order to use cargo-fuzz I had to make the code available as a library crate, I re-exported all of the types that I needed for the fuzzer and a few more that seemed like they might be convenient if someone wanted to use Just as a library.

ALSO! While I was writing this, it found a crash!  I'll file a bug when I figure out what the output means for reproducing it.


casey:
Sweet, thank you so much for this!

I removed the extra `pub use`s, since I'm nervous about the additional stability guarantees implied by making just a lib crate. I'm definitely not totally opposed, but I'd like to understand the use cases, and then only make public those parts which support those use cases. I opened #365 for discussion. Definitely comment if you're interested in using just as a library!

Super sweet that it caught something! I just merged a fix.

Issue #341, opened 2018-08-16
Print recipe signature in case of missing arguments


casey:
If an argument is missing, just should print out the recipe signature including the names of all arguments.

casey:
Fixed by @ladysamantha with #369. Thank you!

PR #369, opened 2018-10-28
Print recipe signature if missing arguments

ladysamantha:
references issue #341 

Wasn't sure what format would be the best, but I took a stab at it and and printed out a simple usage doc on top of the error that was already there. Let me know what more I can do to help/make the code cleaner! I'm still fairly new to rust, so if this seems like a naÏve approach, please forgive me :laughing:

### DIFF
--- a/src/justfile.rs
+++ b/src/justfile.rs
@@ -99,6 +99,7 @@ impl<'a> Justfile<'a> where {
           if !argument_range.range_contains(argument_count) {
             return Err(RuntimeError::ArgumentCountMismatch {
               recipe: recipe.name,
+              parameters: recipe.parameters.iter().map(|p| p.name).collect(),
               found: tail.len(),
               min: recipe.min_arguments(),
               max: recipe.max_arguments(),
@@ -307,11 +308,13 @@ a return code:
     {
       ArgumentCountMismatch {
         recipe,
+        parameters,
         found,
         min,
         max,
       } => {
         assert_eq!(recipe, "a");
+        assert_eq!(parameters, ["b", "c", "d"]);
         assert_eq!(found, 2);
         assert_eq!(min, 3);
         assert_eq!(max, 3);
@@ -328,11 +331,13 @@ a return code:
     {
       ArgumentCountMismatch {
         recipe,
+        parameters,
         found,
         min,
         max,
       } => {
         assert_eq!(recipe, "a");
+        assert_eq!(parameters, ["b", "c", "d"]);
         assert_eq!(found, 2);
         assert_eq!(min, 3);
         assert_eq!(max, usize::MAX - 1);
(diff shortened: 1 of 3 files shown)


casey:
Thank you for the PR! I think this has a stray 2nd commit. You might want to create a separate branch for the pull request, so that pushes to master don't land on the PR after you've opened it.

src/runtime_error.rs
```diff
@@ -71,12 +71,13 @@ impl<'a> Display for RuntimeError<'a> {
...
       },
-      ArgumentCountMismatch{recipe, found, min, max} => {
+      ArgumentCountMismatch{recipe, ref parameters, found, min, max} => {
         if min == max {
           let expected = min;
-          write!(f, "Recipe `{}` got {} argument{} but {}takes {}",
+          write!(f, "Recipe `{}` got {} argument{} but {}takes {}\n",
                     recipe, found, maybe_s(found),
                     if expected < found { "only " } else { "" }, expected)?;
+          write!(f, "Usage:\n just {} {}", recipe, parameters.join(" "));
```
casey: Nit: Usage should be lowercase, since `error:` is also lowercase.

casey (changes requested):
Parameters have a `Display` implementation, so that should be reused to format optional and 
 variadic parameters, and to color the usage string. I use the `{:#}` formatter for colored output. Since this message is being printed to stderr, it should use `{:#}` of stderr is attached to a terminal, and `{}` otherwise. Check out `run.rs` for how to do this.

src/runtime_error.rs
casey: Also, let's go with four spaces of indentation in front of the word `just`.

src/justfile.rs
```diff
@@ -308,13 +308,12 @@ a return code:
...
       ArgumentCountMismatch {
         recipe,
-        parameters,
+        parameters: _,
         found,
         min,
         max,
       } => {
         assert_eq!(recipe, "a");
-        assert_eq!(parameters, ["b", "c", "d"]);
```
casey: For the tests, let's leave in the assertion that the names are right. You can collect into a vec of the recipe names and compare with that.

casey (changes requested):

ladysamantha:
> Thank you for the PR! I think this has a stray 2nd commit. You might want to create a separate branch for the pull request, so that pushes to master don't land on the PR after you've opened it.

D'oh! That actually makes sense, I'm just so used to not using branches at work (company policy...) so I forget sometimes to use them lol

Did you want the usage printed on all of the conditions in `ArugmentCountMismatch`?

casey (changes requested):
Sweet, thanks for making those changes. We should check stderr instead of stdout, and I think we should print the usage string in all cases. After that this should be good to go.

casey:
Awesome, thank for your contribution!

ladysamantha:
Thank you! Happy to contribute :blue_heart:

Issue #326, opened 2018-06-23
Support cmd.exe on windows


casey:
Currently, just uses `sh` to execute all recipes on windows. This requires users to have installed sh, usually from cygwin or git-bash.

I would like to make this dependency optional. It should be possible for a user to create a justfile that uses `cmd.exe` to execute recipes.

I'm not sure the best way to indicate that a justfile uses `cmd.exe` for recipe execution, but my thought it to have just search for a file called `justfile.com`, and if it finds it use `cmd.exe` to execute all recipes within. It may be desirable to allow individual recipes to use `cmd.exe`, or to give some kind of syntax within a justfile, aside from the name, for using `cmd.exe` so all that is up in the air.

I don't know enough about windows to be able to do this, so I'm looking for help! If you're a windows dev and would like to implement this, do reach out!

joshuawarner32:
The use case that I'm most interested in here is being able to set up a single justfile that works across windows/mac/linux.  My (initial, unsubstantiated) thought is that, for _most_ commands, I'll want them to work on both `sh` and `cmd`.  Now, happily some simple commands are already fairly portable across both of those, but many commands may need "implementations" for both `sh` and `cmd`.

Now, we _could_ have these split out into different justfiles (distinguished by extension, like you suggested) - but it would make a lot more sense to me if they were actually packed into a single file, with a concise syntax to specify different "implementations" of the same command, for different shells.

Here's an example syntax (not a serious suggestion, just for discussion purposes):

```
# List the changes that have occurred to .rs files since origin/master
[sh] changes:
  git diff --stat origin/master..|grep -E '\.rs\b'
[cmd] changes:
  git diff --stat origin/master..|find '.rs' # I don't know cmd/batch that well, so this could be totally wrong... but for discussion purposes, ...
```

This syntax can naturally be extended to specify that a single implementation happens to work across both shells:

```
[sh|cmd] diff:
  git diff --stat origin/master..
```

This feature does have some overlap with the ability to specify an interpreter through a shebang-style line, but the key difference I'm proposing is that a single "command" (from the user's perspective) can have multiple different implementations for different platforms.

What I'm getting at, is that I'd like to be able to suggest a _single_ workflow/command to run for both windows and mac users, while also making it obvious which commands are not (yet) translated for my platform.  So, for example, `just --list` would show all commands that are specified for either platform, but clearly show that some of them won't work on my platform.

```
$ just --list
Available recipes:
    build            # Build a final executable (available on all platforms)
[...]

PR #384, opened 2019-02-16
Refactor the --shell internals

naufraghi:
This PR contains a little internal rework in the direction of #326:

- add a `Shell` enum around the shell str
- manage `ShLike` shells with the previous `-cu` argument
- allow a generic `Custom` shell without extra argument

The pattern for multi-platform scripts is:

```makefile
build:
    just _build-{{os()}}

_build-linux:
   ...

_build-windows:
   ...
```

And I had to put in place a wrapper around `cmd.exe` to translate `-cu` into `/C`. This PR adds the extra argument only for _sh like_ shells.

An upcoming PR may introduce some new features, like the support for `cmd` with the required options.
A subsequent one may add another option to allow to specify and shell command and some extra arguments, something like:

```
--shell: as it is (with automatic arguments for known shells)
--shell-args: multiple arguments, may override an automatic argument
```

I'm not a Windows user, but I have some CI running on Windows where I can test something.

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -135,7 +135,7 @@ pub fn run() {
       Arg::with_name("SHELL")
         .long("shell")
         .takes_value(true)
-        .default_value(DEFAULT_SHELL)
+        .default_value(DEFAULT_SHELL.name())
         .help("Invoke <SHELL> to run recipes"),
     )
     .arg(
@@ -423,7 +423,10 @@ pub fn run() {
     evaluate: matches.is_present("EVALUATE"),
     highlight: matches.is_present("HIGHLIGHT"),
     quiet: matches.is_present("QUIET"),
-    shell: matches.value_of("SHELL").unwrap(),
+    shell: matches
+      .value_of("SHELL")
+      .map(|s| Shell::new(s))
+      .unwrap_or(DEFAULT_SHELL),
     verbosity,
     color,
     overrides,
--- a/src/assignment_evaluator.rs
+++ b/src/assignment_evaluator.rs
@@ -12,7 +12,7 @@ pub struct AssignmentEvaluator<'a: 'b, 'b> {
   pub overrides: &'b Map<&'b str, &'b str>,
   pub quiet: bool,
   pub scope: &'b Map<&'a str, String>,
-  pub shell: &'b str,
+  pub shell: &'b Shell<'b>,
 }
 
 impl<'a, 'b> AssignmentEvaluator<'a, 'b> {
@@ -22,7 +22,7 @@ impl<'a, 'b> AssignmentEvaluator<'a, 'b> {
     dotenv: &'b Map<String, String>,
     overrides: &Map<&str, &str>,
     quiet: bool,
-    shell: &'a str,
+    shell: &'b Shell<'a>,
     dry_run: bool,
   ) -> RunResult<'a, Map<&'a str, String>> {
     let mut evaluator = AssignmentEvaluator {
@@ -140,11 +140,11 @@ impl<'a, 'b> AssignmentEvaluator<'a, 'b> {
     raw: &str,
     token: &Token<'a>,
   ) -> RunResult<'a, String> {
-    let mut cmd = Command::new(self.shell);
+    let mut cmd = self.shell.command();
 
     cmd.export_environment_variables(self.scope, dotenv, self.exports)?;
 
-    cmd.arg("-cu").arg(raw);
+    cmd.arg(raw);
 
     cmd.stderr(if self.quiet {
       process::Stdio::null()
--- a/src/common.rs
+++ b/src/common.rs
@@ -34,5 +34,6 @@ pub use recipe::{Recipe, RecipeContext};
 pub use recipe_resolver::RecipeResolver;
 pub use runtime_error::{RunResult, RuntimeError};
 pub use shebang::Shebang;
+pub use shell::Shell;
 pub use token::{Token, TokenKind};
 pub use verbosity::Verbosity;
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -1,6 +1,6 @@
 use common::*;
 
-pub const DEFAULT_SHELL: &str = "sh";
+pub const DEFAULT_SHELL: Shell = Shell::ShLike("sh");
 
 pub struct Configuration<'a> {
   pub dry_run: bool,
@@ -8,7 +8,7 @@ pub struct Configuration<'a> {
   pub highlight: bool,
   pub overrides: Map<&'a str, &'a str>,
   pub quiet: bool,
-  pub shell: &'a str,
+  pub shell: Shell<'a>,
   pub color: Color,
   pub verbosity: Verbosity,
 }
--- a/src/justfile.rs
+++ b/src/justfile.rs
@@ -69,7 +69,7 @@ impl<'a> Justfile<'a> where {
       &dotenv,
       &configuration.overrides,
       configuration.quiet,
-      configuration.shell,
+      &configuration.shell,
       configuration.dry_run,
     )?;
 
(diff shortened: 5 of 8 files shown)


naufraghi:
I realized that this PR may break scripts for someone that is using a wrapper with `--shell cmd` like I did in my tests, but I may have done like the user below suggested:

> As an aside I have partially got this working for us by creating a small passthru program called sh.exe that translates -cu to -command and then calls Powershell - works for us!
>
> _Originally posted by @justdan96 in https://github.com/casey/just/issues/326#issuecomment-400352269_

src/run.rs
```diff
@@ -423,7 +423,10 @@ pub fn run() {
     evaluate: matches.is_present("EVALUATE"),
     highlight: matches.is_present("HIGHLIGHT"),
     quiet: matches.is_present("QUIET"),
-    shell: matches.value_of("SHELL").unwrap(),
+    shell: matches
+      .value_of("SHELL")
+      .map(|s| Shell::new(s))
+      .unwrap_or(DEFAULT_SHELL),
```
naufraghi, line 429: This should never happen, we have the `default_value`, I can restore the `unwrap()` we had before.

casey:
Hi Matteo, thank you for the PR!

I like the direction this is going, but I think it would be good to avoid special-casing `sh`-like shells, and have all shells be treated uniformly.

In the long run, I would like to support per-justfile settings, so you could do:

```
just.shell = ["cmd.exe", "-command"]

foo:
  dir
```

and not need to pass in a command-line argument every time, but it's probably not a good idea to wait for a perfect solution instead of landing one that works now.

I thought about a number of different ways to approach this, which you've probably considered, and many seem to have problems. I think the third approach, a new flag that takes a string that will be split on whitespace, is the most promising.

What do you think?

The name `--shell-args` is not great, but I can't immediately think of anything better.

### split value of `--shell` on whitespace

This would be a breaking change for users who use `--shell bash`, since just would try to run their commands with `bash COMMAND`, instead of `bash -cu COMMAND`.


### add a --shell-arg flag that accepted multiple values to pass arguments to the shell

This wouldn't work because many of the arguments you would want to pass start with `-`, and clap would intercept them and interpret them as arguments to `just`.


### new flag that supersedes `--shell`

The flag could be called `--shell-args`, and would expect a whitespace-separated string containing the shell and arguments.

`just --shell bash` → `just --shell-args "bash -cu"` 

`just` would complain if it got `--shell` and `--shell-args`, and `--shell` would keep its old behavior.

naufraghi:
1. _per file shell setting_: seems a good idea, but the single multi-platform file is not covered (but one can use multiple files, with `just windows/build` instead of `just _build-windows`). This may solve another problem I found, that `--shell` option is not propagated to child `just` process, one may define ad extra source in the environment, like `JUST_PARENT_SHELL`, or define a function to have back the current `--shell` value to propagate manually, or have a per file `shell` option as you suggested with a different layout, too many options :smile:),
2. _shell with options_: I was thinking that in the long run, most shells will have a builtin set of correct extra arguments (found by name), and for the uncovered ones, a `--shell-args "-some --extra /argument"` may cover most of the cases (escape masters may pass white-spaces and quotes too, maybe, but for extreme cases a thin shell wrapper is always there as a backup solution).
3. _shell special-casing_: I think this will be hard, in my current WIP branch, I've this, because `cmd` is _different_ (https://github.com/casey/just/pull/386/commits/2930293e72c5245344a70afc889b682a1cf15e4d):

```rust
      #[test] fn sh()   { integration_test("sh",   $text, &[$($arg)*], $stdout, $stderr, $status, false); }
      #[test] fn dash() { integration_test("dash", $text, &[$($arg)*], $stdout, $stderr, $status, false); }
      #[test] fn bash() { integration_test("bash", $text, &[$($arg)*], $stdout, $stderr, $status, false); }
      #[test] fn cmd() {
        //! Not all tests are compatible with cmd, for example:
        //! - `printf` is missing
        //! - `/bin/echo` is missing
        //! - `$VARIABLE` are spelled as `%VARIABLE%`
        //! - single quotes have no special meaning in cmd (to be added)
        let justfile = $text;
        let may_fail = justfile.find("printf").is_some() || justfile.find("/bin").is_some() || justfile.find("$").is_some();
        integration_test("cmd.exe", justfile, &[$($arg)*], $stdout, $stderr, $status, may_fail);
      }
 ```

This is a trash code, but the idea of the problem is here.

TIL: you can test `cmd` from Linux using the wine `cmd.exe`

edit: `--shell-args` is just the first name I found, I'm not defending it in any way :smile:

**Comment by casey** [comment]
> 1. _per file shell setting_: seems a good idea, but the single multi-platform file is not covered (but one can use multiple files, with `just windows/build` instead of `just _build-windows`). This may solve another problem I found, that `--shell` option is not propagated to child `just` process, one may define ad extra source in the environment, like `JUST_PARENT_SHELL`, or define a function to have back the current `--shell` value to propagate manually, or have a per file `shell` option as you suggested with a different layout, too many options 😄),

Eventually I'd like to fix this with attributes that conditionally enable recipes based on the OS:

```
#[cfg(os = "windows")]
show-files:
  dir

#[cfg(os = "linux")]
show-files:
  ls
```

Another fuzzy is to conditionally enable submodules, :

```
#[cfg(os = "windows")]
mod build "windows.just"

#[cfg(os = "linux")]
mod build "linux.just"

use build::*
```

(Which is all pretty shamelessly cribbed from rust.)

> 2. _shell with options_: I was thinking that in the long run, most shells will have a builtin set of correct extra arguments (found by name), and for the uncovered ones, a `--shell-args "-some --extra /argument"` may cover most of the cases (escape masters may pass white-spaces and quotes too, maybe, but for extreme cases a thin shell wrapper is always there as a backup solution).

Unfortunately `--shell-args "-some --extra /argument"` won't work. Even though the argument to `--shell-args` is quoted, just can't see that, so it doesn't know not to interpret it as another option (the string `"-some --extra /argument"`).

I definitely agree that most shells will have a few obviously correct arguments that most people should use. However, I think that having `just` add additional arguments to shells would confuse users more than it would help.

However, it is an annoying burden that users who want to use just with different shells will have to figure out how to invoke their shells, when this is something that just could know.

As a compromise, we could have a `--shell-args` flag that takes the command line and any arguments: `--command "sh -cu"`, but then also have another flag that switches between a few convenient presets:

`--shell-type sh` → `sh -cu`
`--shell-type cmd` → `cmd /Q /C`

> 3. _shell special-casing_: I think this will be hard, in my current WIP branch, I've this, because `cmd` is _different_ ([2930293](https://github.com/casey/just/commit/2930293e72c5245344a70afc889b682a1cf15e4d)):
> 
> ```rust
>       #[test] fn sh()   { integration_test("sh",   $text, &[$($arg)*], $stdout, $stderr, $status, false); }
>       #[test] fn dash() { integration_test("dash", $text, &[$($arg)*], $stdout, $stderr, $status, false); }
>       #[test] fn bash() { integration_test("bash", $text, &[$($arg)*], $stdout, $stderr, $status, false); }
>       #[test] fn cmd() {
>         //! Not all tests are compatible with cmd, for example:
>         //! - `printf` is missing
>         //! - `/bin/echo` is missing
>         //! - `$VARIABLE` are spelled as `%VARIABLE%`
>         //! - single quotes have no special meaning in cmd (to be added)
>         let justfile = $text;
>         let may_fail = justfile.find("printf").is_some() || justfile.find("/bin").is_some() || justfile.find("$").is_some();
>         integration_test("cmd.exe", justfile, &[$($arg)*], $stdout, $stderr, $status, may_fail);
>       }
> ```

Oh yikes, I didn't think about getting the integration tests running on windows T_T. We can cross that bridge when we get to it.

This is kind of another reason that I'm loathe to add special-casing for different shells. Before we do that, I'd really like to make sure that they're getting test coverage.

casey:
See #386.

PR #386, opened 2019-02-17
Dev/add windows cmd support

naufraghi:
I'm using an PR to trigger the CI on windows.

TODO:

- [ ] Merge #384 and rebase
- [ ] add integration tests

### DIFF
--- a/src/assignment_evaluator.rs
+++ b/src/assignment_evaluator.rs
@@ -12,7 +12,7 @@ pub struct AssignmentEvaluator<'a: 'b, 'b> {
   pub overrides: &'b Map<&'b str, &'b str>,
   pub quiet: bool,
   pub scope: &'b Map<&'a str, String>,
-  pub shell: &'b str,
+  pub shell: &'b Shell<'b>,
 }
 
 impl<'a, 'b> AssignmentEvaluator<'a, 'b> {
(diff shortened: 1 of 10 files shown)


casey:
I'm sorry to close this, since a lot of care and attention went in to it, but I'm leaning towards an alternative refactoring of the shell infrastructure.

I deprecated `=` syntax in assignments, in favor of `:=`. This makes extending the parser with new assignment-like constructs much simpler.

So, I'm going to introduce a `set SETTING := VALUE` syntax, that allows justfile-wide settings. The first such setting that I'll introduce will be something that lets you do: `set shell := ["sh", "-c"]` which sets the shell for all recipes and backticks in the entire file.

I think that this is much better than the current `--shell` flag and associated infrastructure for a few reasons:

- Often, justfiles will be written with one particular shell or interpreter in mind, and allowing users to pass in a shell on the command line doesn't make much sense. If a justfile is written with `cmd.exe` commands and windows paths, passing in `--shell sh` would be bad.

- The `--shell` syntax requires setting an alias for it to be ergonomic, which is annoying.

I hope that makes sense and sounds reasonable. Apologies again for letting this sit for so long and not merging it.

naufraghi:
Don't worry, I almost forgot about it :smile:, it was still an interesting exercise.

Only two things:

1. there was a commit (https://github.com/casey/just/pull/386/commits/4c1e8fbd9f0de1e53978a884bc9470a543739ec9) correcting a typo, maybe it has already been corrected in the master
2. if with the new approach it is possible to use `just` even under windows, even in a sub-optimal way, my use case is still covered, so thanks again for your tool!

casey:
Nice, I didn't catch that. I opened #418 which I'll merge to pick up that commit.

PR #393, opened 2019-04-08
Allow outer shebangs so justfiles can be used as scripts

### LINKED ISSUE #367: Executable Justfiles
## Motivation

I use Make to make simple script tools. I find it's much easier to use (and maintain) than my previous solution -- bash switch blocks. Recently I've been porting these Makefiles to Just. But, Just has one weakness, I can't make a single file executable like I can with shell scripts. This means that when I want to distribute my tools to others I have to explain how to them how to use them, how it's like a Makefile, and how it's not. Being similar to a Makefile, creates friction, as I find that many devs don't like seeing Makefiles for things other than building code. To me the structure of a Makefile is ideal for writing composable scripts. For more complex tooling I even have multiple Makefiles in a directory tree acting like modules, separating out different common utilities into each file. If you think that this sounds like madness, take a look at [BuildRoot](https://github.com/buildroot/buildroot). It's a production codebase that's almost entirely written in Makefiles!

I also spend some amount of time using and maintaining a large recursive-make build script. This may cause some to shudder, but that's just how it is. One day, I'd like to port this to Just, as I think it has the potential to smooth out many of the developer pain points. The similarity in syntax makes this especially appealing over other alternatives that would require a larger rewrite. However, I think Just lacks some really killer features that would make this work more justifiable (pun intended).

## Features

Just has many advantages over Make as pointed out by its docs, but I suggest here some features that could make it even more useful for writing declarative tooling scripts and handling complex builds. Though they may seem unrelated, they are really two sides of the same coin.

### Top Level Shebang

This is very simple, and at most would require some low-level changes to how Just handles file inputs. What I want to support is having a line like the following at the top of the file.

```
#!/usr/bin/env just <args> --justfile
```

This should look and act just like a regular shebang you may see in a single-file python/js/shell script. The `<args>` is a placeholder for whatever args Just might need to support this feature. For example, the way Just is written right now it would likely need `--working-directory` to be supplied; though it could easily default to the CWD.

Currently this kind of usage is explicitly blocked with an error:

```
error: `#!` is reserved syntax outside of recipes
  |
1 | #!/usr/bin/env just --working-directory ./ --justfile
  | ^
```

I understand this is likely done to avoid confusion with the shebangs allowed in recipes, but it shouldn't be too hard to disambiguate, especially if you only allow shebangs on the first line. I think dropping this error is probably about all you need to make this work.

### Alternate Naming Convention

If a justfile is to be executable, it makes sense to allow for it to be named something other than `Justfile`. Here I'd suggest the `*.just` extension. Supporting this may be more about editor support more than anything, but within Just I'd at least hope that it would support opening `*.just` files by default if no `Justfiles` or `justfile` is present in in the CWD. 

Ideally, I'd hope that multiple `*.just` files could be selected from based on target, logically acting like one large `Justfile`. Collisions between targets in separate files could be disambiguated with a fully qualified (Make compatible) target like `path/to/filename@target`. See below for more.

###  Inter-file Targets

The dream-usage I'm aiming for is that one could execute a top-level `*.just` file that could call out to other "child" `*.just` files by specifying target alone. This would be similar to the recursive make idiom, except with a few *huge* advantages.

1. **Composition:** Targets of another Justfile can be specified as dependent targets of a recipe alone. This allows them to be interleaved with other targets in the same file, leading to concise and introspectable composition.

```
get-build-set THING: 
    getters/thing.just@get {{THING}} \
    build {{THING}} \
    setters/thing.just@set {{THING}} \
```

2.  **Target Inlining:** Using something like`$(MAKE)` to execute dependent targets makes those actions opaque and increases process overhead. In a large build, the cost of creating all those extra processes can be rather burdensome. Hoisting the target up into the recipe makes it much easier to introspect the dependency graph and perform optimizations. 

3. **Introspection:** While you'd probably only want to supply targets from the CWDs file/s by default when `--list` or `--summary` is used, this behavior could be changed with a `--recursive` flag. The upshot of this being that the entire graph can be inspected and debugged by the programmer, without having to execute any recipes. Anyone who's spent time working with a recursive-make build knows how much of a lifesaver this would be.

casey:
This was long overdue, since it was really just a matter of removing code.

With this change, justfiles can be used as scripts. Just write a justfile with a shebang line on top:

```make
#!/usr/bin/env just --justfile

foo:
  echo foo
```

and you can do:

```sh
$ ./justfile
echo foo
foo
```

You can use a shebang like this to avoid changing directories:

```
#!/usr/bin/env just --working-directory ./ --justfile
```

Fixes #367.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -655,27 +655,28 @@ $ just foo/
 
 === Just scripts
 
-Just can be used as an interpreter for standalone, executable scripts.
+Just can be used as an interpreter for standalone executable scripts.
 
-To do so, add a suitable shebang at the top of a justfile and make it executable:
+To do so, add a suitable shebang to the top of a justfile and make it executable:
 
 ```sh
 $ cat > script <<EOF
 #!/usr/bin/env just --justfile
+
 foo:
   echo foo
 EOF
 $ chmod +x script
-$ ./justfile
+$ ./script foo
 echo foo
 foo
 ```
 
-When a script with a shebang is run, the system supplies the path to the script as an argument to the command in the shebang. So, with a shebang of `#!/usr/bin/env just --justfile`, the command will be `/usr/bin/env just --justfile PATH_TO_SCRIPT`, which allows `just` to location the `justfile`. The shebang line itself starts with a `#`, so will be ignored as a comment.
+When a script with a shebang is run, the system supplies the path to the script as an argument to the command in the shebang. So, with a shebang of `#!/usr/bin/env just --justfile`, the command will be `/usr/bin/env just --justfile PATH_TO_SCRIPT`, which allows `just` to locate correct `justfile`. The shebang line itself starts will be ignored as a comment.
 
 Note that when `just` is invoked with `--justfile` and not `--working-directory`, the working directory will be the directory containing the justfile, not the directory in which the command was invoked.
 
-To use the current working directory as the working directory instead, use `#!/usr/bin/env just --working-directory . --justfile`.
+To use the current working directory instead, use `#!/usr/bin/env just --working-directory . --justfile`.
 
 
 == Frequently Asked Questions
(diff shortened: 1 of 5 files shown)


casey:
I'm very curious what uses people put this feature to, so do let me know if you wind up using it!

gl-yziquel:
I'm using it to make more custom and convenient wrappers around the himalaya command line interface tool for emails. It's extremely convenient to be able to hack around and improve on the user experience of a raw CLI. Makes CLI composable, in some sense.

PR #401, opened 2019-04-12
Bump version: 0.3.13 -> 0.4.0

casey:
Technically the addition of aliases was a breaking change, since a recipe named alias with no arguments will be parsed differently, so bumping the minor version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,17 @@
 # Changelog
 
-All notable changes to this project will be documented in this file.
+## [v0.4.0] - 2019-04-12
+### Added
+- Add recipe aliases by @ryloric (#390)
+- Allow arbitrary expressions as default arguments (#400)
+- Add justfile summaries (#399)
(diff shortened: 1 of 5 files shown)


casey:
[v0.4.0](https://github.com/casey/just/releases/tag/v0.4.0) is released!

It comes with a few goodies, in particular

- Recipe aliases by @ryloric, #390:

```
# recipe `build` can also be invoked with `b`
alias b = build

build:
  echo 'Building!'
```

- Arbitrary expressions as default arguments, #400:

```
foo bar=`cat baz`:
  echo {{bar}}
```

- Allow outer shebang lines so justfiles can be used as scripts (#393):

```
#!/usr/bin/env just --justfile

foo:
  echo 'Foo!'
```

- Allow `--justfile` without `--working-directory` by @smonami (#392), which was needed for the above.

- And since last release, a Chinese translation of the readme was made by @chinanf-boy (#377)

Thanks so much to everyone who contributed!

PR #406, opened 2019-04-16
Replace regex-based lexer with character-at-a-time lexer

### LINKED ISSUE #241: Rewrite Lexer
The lexer is by far the messiest code in `just`. Rewrite it using a character-at-a-time lexer, instead of regular expressions.

- [ ] parser invokes lexer
- [ ] new "Position" type, which stores an offset and a source file
- [ ] make eyes bleed less when reading it

### LINKED ISSUE #251: Crater for Just
There are a lot of justfiles on github, and it would be nice to leverage them to make sure that existing justfiles aren't broken by future changes to`just`.

I'm imagining a tool that would, for a new version of just:

- download all files on named `justfile` or `Justfile`
- check that all files are tokenized identically under the new and old versions
- check that all files are parsed identically under the new and old versions
- check that all files produce identical dry-runs under new and old versions

Caldera would be an appropriate name. It will probably need a blacklist for those files which are called `justfile` but which aren't actually justfiles.

casey:
Fixes #241, which I've been threatening to do for a long time.

The current lexer uses regexes and is god awful.

The new lexer processes text mostly one character at a time, making decisions about which tokens to emit along the way. It's more verbose than the old lexer, but the new code is much easer to read, understand, and modify.

Also, the new lexer is 4x faster than the old lexer, when tested against a corpus of justfiles collected from github. In release mode, the change is more dramatic, with the new lexer being 15x faster.

I suspect that the speed increase is partially due to the old lexer trying to lex tokens by matching regexes in a sequence, which led to a lot of wasted work, whereas the new lexer is usually able to make a decision about which token to emit next by looking at the next character.

Since this is such a massive change, I'm testing it using a new tool called Janus, which downloads all justfiles on github and feeds them to multiple versions of just, looking for differences in behavior. Janus is of course inspired by Rust's crater, and once I finally release it we can close #251.

So far the results from Janus are encouraging. The just+new lexer produces slightly better error messages in a few cases, as well as being able to parse a previously unparsable justfile.

The only change that I need to investigate before landing the rewrite is a change in the handling of windows newlines at the end of recipe lines. For example, it looks like a text that the old lexer would extract from the line `echo foo\r\n` would be  `echo foo\r`, whereas the new lexer correctly recognizes `\r\n` as a unit, and extracts `echo foo` as the line text.

Although the new lexer's behavior is correct, I'm slightly concerned that there might be cases where the new behavior might cause a shebang recipe to fail when previously it would have succeeded.


casey:
An additional change in behavior is that the new lexer includes trailing space as part of a recipe line. I.E if a recipe line contains `echo foo  `, those two spaces will be extracted and included. I think this is an improvement, since we definitely want to be able to execute [whitespace](https://en.wikipedia.org/wiki/Whitespace_(programming_language)) shebang recipes, and the old lexer would break those.

casey:
Since there were so few justfiles that saw changes (2 out of 498), and those that did won't change behavior, I'm going to merge this.

Issue #374, opened 2018-11-20
prompting for a password


slisznia:
My recipe needs to deploy to prod and take a password silently. Without using ENV vars, is it possible to prompt and capture a password into a just-variable? I don't believe "read" command is useful here since it takes a bash variable to set the result into.

i.e.
PASS = \`prompt without echoing and return\`

casey:
I don't think there's a good way to do this with just variables.

However, you can do this with env variables in a shebang recipe:

```
pw:
	#!/usr/bin/env bash
	echo Please enter your password:
	read -s pw
	echo $pw
```

Shebang recipes are executed as scripts, so variables work. You could also create a `.env` file containing the variables. Does that help?

slisznia:
I saw the inline invocation of other languages through shebangs but how do you make the value of $pw (from your example) available outside of the shebang-recipe? The password needs to be reused by possibly many other recipes, so ideally it's stored in a variable.

casey:
Ah, I see. Yeah, that's tough. You could read the password and then re-invoke just:

```
pw:
	#!/usr/bin/env bash
	echo Please enter your password:
	read -s pw
	echo $pw
	just pw=$pw deploy
```

Does putting it in a `.env` file work?

slisznia:
No it doesn't - I tried that before posting this "issue" in the first place. I'm seeing "error: Backtick failed with exit code 1"

Neither does wrapping the password capture in a bash-helper-program like the prompt_for_password.sh  work. I'm out of ideas for now...

casey:
Hmm, that's unfortunate. One possibility is that backtick commands aren't able to read from stdin, although I'm not sure why that would be the case.

slisznia:
That seems to be the case exactly - it doesn't like the stdin program inside the backticks.

casey:
I opened #376 to track that issue, which is definitely a bug.

casey:
Fixed in #409!

PR #409, opened 2019-04-17
Pass stdin handle to backtick process

### LINKED ISSUE #376: var = `read foo && echo foo` doesn't work
In #374, @slisznia wanted a way to read a password from a user into an env variable. Unfortunately, for some strange reason, the following doesn't work:

```ocaml
var = `read foo && echo foo`
```

The following works, and I'm not sure what just is doing differently when it runs backtick commands:

```rust
use std::process::Command;

fn main() {
    let mut cmd = Command::new("sh");

    cmd.arg("-cu").arg("read foo && echo $foo");

    cmd.status().unwrap();
}
```

I'm not sure why this is the case. I won't have time to dig into this for the near future, so if someone wants to take a crack at it they are most welcome!

casey:
Backtick processes are launched (via the `brev` crate) using `Command::output`, which by default prevents stdin from being inherited.

This diff passes stdin through. This allows backtick commands to read from stdin, and so fixes #376 and #374.

In particular, this allows:

```make
password := `read PASSWORD && echo $PASSWORD`

ssh:
  echo {{password}}
```

To support this change, integration tests are modified to take an additional string argument, which is written to the `just` child process's stdin. Existing tests are passed `""`, and a few tests have been added that read from stdin and echo it, to make sure that it actually works.

cc: @slisznia

### DIFF
--- a/src/assignment_evaluator.rs
+++ b/src/assignment_evaluator.rs
@@ -141,9 +141,11 @@ impl<'a, 'b> AssignmentEvaluator<'a, 'b> {
   ) -> RunResult<'a, String> {
     let mut cmd = Command::new(self.shell);
 
+    cmd.arg("-cu").arg(raw);
+
     cmd.export_environment_variables(self.scope, dotenv, self.exports)?;
 
-    cmd.arg("-cu").arg(raw);
+    cmd.stdin(process::Stdio::inherit());
 
     cmd.stderr(if self.quiet {
       process::Stdio::null()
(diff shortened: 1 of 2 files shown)


casey:
It should be noted that the order that backticks execute in is unspecified and may change in the future.

Currently, dependencies between variables determine backtick execution order, not the order they appear in source. This justfile:

```make
b := a + `echo bar`
a := `echo foo`

foo:
  echo {{b}}
```

will run `echo foo` before `echo bar`, even though `echo foo` appears a line after `echo bar`. However, this is an implementation detail, and may change.

So, if multiple top-level backticks read from stdin, the order that they read from stdin is undefined.

Issue #379, opened 2018-11-29
🔥 Assignments and aliases conflicts with default arguments


casey:
`:=` has been added as a synonym for `=` in assignments, aliases and exports, and `=` is deprecated in those same contexts and will be removed in a future update, most likely `v0.5.0`.

Since this is a disruptive change, I think it's a good idea to explain the rationale.

Currently, `=` is used in assignments, aliases, exports, and default arguments:

```
alias b = build
foo = "bar"
export var = "val"

build module="rob":
  ./build --module {{module}} 
```

The just language doesn't have keywords in the conventional sense, so the parser uses lookahead to disambiguate conflicting constructs.

However, lookahead is limited and messy, and some conflicts cannot be resolved. So for example, these constructs fail to parse:

```
# the parser thinks that this is an export
export foo = "bar":
  echo {{foo}}

# the parser thinks that this is an alias:
bar = "baz"
alias foo = bar:
  echo {{foo}}
```

Also, there are additional constructs that I'd like to add that will produce further conflicts, like settings: `set setting = value`.

In order to fix this, there are a few options:

A. Use more lookahead to disambiguate expressions
B. Use a token other than `=` to introduce default arguments to recipes
C. Use a token other than `=` in assignments, exports and aliases

Using more lookahead, as in A, gets messy very quickly, especially as it's extended to longer and more complex constructs, which would limit extension of the grammar in the future, so I think ruling that out is reasonable.

B is reasonable, but I don't think that there is a good alternative to `=` for default argument values. `:` is already used to end recipe argument lists.

On the other hand, `:=` is a good alternative for assignments, exports, aliases, and a future `set setting := value` notation. `:=` is used in `make`, which `just` is most similar to, and is also used for assignment in Go and Algol. So, I think C is the best choice.

My current plan is to release this as `v0.4.3`, and follow up after some time has passed with a version, probably `v0.5.0` that removes the old meaning of `=`. I don't know what the most popular ways of getting just are, but I'd like to give v0.4.3 a chance to appear in package managers so that nobody is in the unfortunate situation of needing to run the same justfile on `< 0.4.3` and `>= 0.5.0`, and not having a construct available that will work on both.

casey:
The reason this happens is because both `export`, `alias`, default arguments use `=`. I'd like to introduce `set SETTING = VALUE` items, which will extend the issue to recipes named `set` as well.

[...]

PR #413, opened 2019-04-18
Deprecate `=` in assignments, aliases, and exports in favor of `:=`

casey:
This PR adds `:=` as a synonym for `=` in assignments, aliases and exports, and adds a deprecation warning `=` in those same contexts.

The deprecation warning looks like this:

```
warning: `=` in assignments, exports, and aliases is being phased out on favor of `:=`
Please see this issue for more details: https://github.com/casey/just/issues/379
```

Since this is a disruptive change, I'd like to go over the rationale, just in case anyone is curious.

Currently, `=` is used in assignments, aliases, exports, and default arguments:

```
alias b = build
foo = "bar"
export var = "val"

build module="rob":
  ./build --module {{module}} 
```

The just language doesn't have keywords in the conventional sense, so the parser uses lookahead to disambiguate conflicting constructs.

However, lookahead is limited and messy, and some conflicts cannot be resolved. So for example, these constructs fail to parse:

```
# the parser thinks that this is an export
export foo = "bar":
  echo {{foo}}

# the parser thinks that this is an alias:
bar = "baz"
alias foo = bar:
  echo {{foo}}
```

Also, there are additional constructs that I'd like to add that will produce further conflicts, like settings.

In order to fix this, there are a few options:

A. Use more lookahead to disambiguate expressions
B. Use a token other than `=` to introduce default arguments to recipes
C. Use a token other than `=` in assignments, exports and aliases

Using more lookahead, as in A, gets messy very quickly, especially as it's extended to longer and more complex constructs, which would limit extension of the grammar in the future, so I think ruling that out is reasonable.

B is reasonable, but I don't think that there is a good alternative to `=` for default argument values. `:` is already used to end recipe argument lists.

On the other hand, `:=` is a good alternative for assignments, exports, aliases, and a future `set setting := value` notation. `:=` is used in `make`, which `just` is most similar to, and is also used for assignment in Go and Algol. So, I think C is the best choice.

My current plan is to release this as `v0.4.3`, and follow up after some time has passed with a version, probably `v0.5.0` that removes the old meaning of `=`. I don't know what the most popular ways of getting just are, but I'd like to give v0.4.3 a chance to appear in package managers so that nobody is in the unfortunate situation of needing to run the same justfile on `< 0.4.3` and `>= 0.5.0`, and not having a construct available that will work on both.

I was thinking of supplying a `--fix` flag, as I mentioned in #379, but just's current parse-and-dump functionality re-orders constructs and strips comments, so using it to implement fixup functionality would be harder than I thought.


aoeu:
@casey I noticed this referencing something else in the README, I like the change.  Any idea what version of `just` it will be in? (I'm on `just v0.4.2` via `cargo`, FWIW.)   I'm curious since I'm trying to get some people started on using  `just` and some files I've written, they're likely going to be using `homebrew` or `cargo`, and I want to make sure I'm introducing justfiles with the new syntax (but need to make sure they'll be using compatible versions).

casey:
Whoops, I meant to cut a release with this sooner. I'll cut v0.4.3 with the change and push it out today, probably later tonight. I'm not sure how long homebrew lags behind crates.io.

casey:
I just cut v0.4.3. It's available on Cargo now, and prebuilt binaries will appear on the [release page](https://github.com/casey/just/releases) as they're built by CI.

I don't maintain the homebrew package, but it should be pretty easy to update.

PR #421, opened 2019-04-26
Mention Make’s “phony target” workaround in the comparison

roryokane:
I think mentioning `.PHONY` is necessary to be fair to Make. Someone might decide that using phony targets in Make is the best solution for them (e.g. due to portability), despite Make having worse error messages and so on, and we shouldn’t hide this option from them. But I tried to mention the reasons one might prefer Just’s implementation of phony targets.


casey:
I think that's totally reasonable. Thank you for the PR!

PR #436, opened 2019-05-12
Ignore justfile name case while searching

shevtsiv:


casey:
I notice that this is a draft, so let me know when you'd like me to take a look.

shevtsiv:
This is a draft because I haven't covered this features with tests yet, but I plan to leave this code in the first commit almost unchanged, so feel free to review this part, while I still working on test coverage. Once tests are done I'll mark this request as ready for review, so it can be reviewed then as well.

src/run.rs
```diff
@@ -273,44 +273,37 @@ pub fn run() {
...
-          }
-          Err(error) => {
-            if error.kind() != io::ErrorKind::NotFound {
-              die!("Error fetching justfile metadata: {}", error)
-            }
-          }
+    match search(
+      env::current_dir()
+        .as_ref()
+        .expect("Error getting current directory"),
```
casey: Just should never panic if there is an error that isn't due to a bug, it should always exit with an error code and a useful error message.

`.expect` panics on `Err`, so it can't be used.

Instead, you can do something like:

```
let current_dir = match env::current_dir() {
  Ok(current_dir) => current_dir,
  Err(io_error) => die!("Error getting current dir: {}", io_error),
};

match search(&current_dir) {
...
}
```

src/run.rs
```diff
@@ -273,44 +273,37 @@ pub fn run() {
...
-      }
-
-      if let Err(error) = env::set_current_dir("..") {
-        die!("Error changing directory: {}", error);
-      }
-    }
-
-    if matches.is_present("EDIT") {
-      edit(name);
+        SearchError::MultipleCandidates { candidates } => {
```
casey: In the interest of keep the size of the run function down, let's move `SearchError` into its own file, `src/search_error.rs`, and add an `impl Display for SearchError {...}`, then all the error formatting can be done in the `impl` block.

Then, in main, we can do:

```
let justfile = match search(&current_dir) {
  Ok(justfile) => justfile,
  Err(search_error) => die!("{}", search_error),
};
```

casey:
Cool, just added some comments. I'm not sure where search should go, but I think it should probably be in its own file, like `src/search.rs`

shevtsiv:
@casey I've added some test coverage for the justfile search function. Although this is work in progress, I would be grateful for review. Currently tests are located in the same file that search function and I'm not sure if it is a good idea. Also, I don't know if we should remove the old integration tests, because it is tests though and maybe it would be better to leave them as well for confidence that everything works properly. 
Another spot I'm worried about is `#[cfg(unix)]` above the `multiple_candidates` test function. I've used this attribute to disable this test running on Windows since it has case insensitive file system and there is no differences between `JustFile` and `justfile`, for example. But I'm not sure if it is a right way because I've also seen manual checking for the file system case sensitivity in the other tests(`test_capitalization_priority`).

casey:
Great! I'm traveling the moment, so forgive me if it takes a few days before I get the chance to do a proper review.

Leaving the current integration tests in place is probably a good idea. They won't hurt, and they'll catch cases where the search code is correct, but it isn't being called correctly.

I'll be back home on Sunday, so I should be able to give you proper feedback soon after that.

src/search.rs
```diff
@@ -0,0 +1,146 @@
...
+      use caseless::default_caseless_match_str;
+      if default_caseless_match_str(name, "justfile") {
+        files.push(entry.path());
+      }
+    }
+  }
+  if files.len() == 1 {
+    Ok(files.pop().unwrap())
+  } else if files.len() > 1 {
+    Err(SearchError::MultipleCandidates { candidates: files })
```
casey: If you rename `files` to `candidates`, then you can do: `SearchError::MultipleCandidates { candidates }`.

src/search.rs
```diff
@@ -0,0 +1,146 @@
+use crate::search_error::SearchError;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+pub fn search(directory: &Path) -> Result<PathBuf, SearchError> {
```
casey: I would rename from `search` to `justfile`, and always call it with the full path name, i.e. `search::justfile`.

src/search.rs
```diff
@@ -0,0 +1,146 @@
+use crate::search_error::SearchError;
```
casey: You can add `search_error::SearchError` to `common.rs`, and do `use crate::common::*:`

src/search.rs
```diff
@@ -0,0 +1,146 @@
...
+  } else if let Some(parent_dir) = directory.parent() {
+    search(parent_dir)
+  } else {
+    Err(SearchError::NotFound)
+  }
+}
+
+#[cfg(test)]
+mod test {
+  use crate::search::search;
```
casey: You can do `use super::*;` to bring all items in the parent module into scope.

src/search.rs
```diff
@@ -0,0 +1,146 @@
...
+  use crate::search_error::SearchError;
+  use std::fs;
+  use tempdir::TempDir;
+
+  #[test]
+  fn not_found() {
+    let tmp = TempDir::new("just-test-justfile-search")
+      .expect("test justfile search: failed to create temporary directory");
+    let path = tmp.path().to_path_buf();
+    match search(path.as_path()) {
```
casey: You can avoid the conversion to a path buf and do `search(tmp.path()`

src/search.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+  }
+
+  #[test]
+  #[cfg(unix)]
+  fn multiple_candidates() {
+    let tmp = TempDir::new("just-test-justfile-search")
+      .expect("test justfile search: failed to create temporary directory");
+    let mut path = tmp.path().to_path_buf();
+    path.push("justfile");
```
casey: I think you'll need to create the files for this test to work.

src/search.rs
```diff
@@ -0,0 +1,146 @@
...
+    match search(path.as_path()) {
+      Err(SearchError::NotFound) => {
+        assert!(true);
+      }
+      _ => panic!("No justfile found error was expected"),
+    }
+  }
+
+  #[test]
+  #[cfg(unix)]
```
casey: Instead of doing `#[cfg(unix)]` let's come up with a way to determine if we're running on a case insensitive filesystem, and thus the test should be skipped.

We can do:

1. Create a file called `justfile`
2. Try to open it using the name `JUSTFILE`
3. If 2. succeeds, we're on a case insensitive filesystem and the test should be skipped.

casey (changes requested):
Sorry for the delay!

src/search.rs
shevtsiv: Sorry, but I'm not sure I understood this moment correctly. If we want to call `justfile` function with full path name(search::justfile()), then we need to import `search` mod here explicitly: `use crate::search;` Thus, should I do `use super::*;` to import `SearchError` only?

casey: Totally reasonable question! Inside of `src/common.rs`, there's a use statement which contains modules that are used directly, you can add `search` to that use statement, like so:

```
// modules
pub(crate) use crate::{color, doc, hint, mode, search};
```

Then, if you do `use crate::common::*;` at the top of `src/search.rs`, the `use super::*;` will bring everything into scope, including the `search` module.

shevtsiv: Thank you for your answer! Fixed it in the last commit.

shevtsiv:
It seems `test_capitalization_priority` test is failing now. I suppose it's because we do not support prioritization anymore. If I am right, should I just delete this test?

casey:
Yup, that's right, we can just delete that test.

casey:
It looks like the tests are green! Is this ready for another round of review?

shevtsiv:
> It looks like the tests are green! Is this ready for another round of review?

I think so!

casey:
Looks good!

Pushed a few additional commits to the branch:

- Rebased onto current master
-  I found `str::eq_ignore_ascii_case` in the standard library, so I removed the dependency on the caseless crate
- Organized imports in common.rs
- The old code would also change the working directory to the parent directory of the found justfile. The new code changed this, but I somehow didn't have a test for this which would have caught this. I added a test and changed the code to change the working directory, as before.
-Put the `justfile` string in a constantly, for slighly better readability
- "JuStFiLe" is, I think, more correctly referred to as [spongebob case](https://knowyourmeme.com/memes/mocking-spongebob), so I changed the name of the test.
- Reworked how some of the search errors are formatted

I also updated the docs to reflect the new functionality. Once the tests pass I'll merge this in.

casey:
Merged! Thank you very much for the contribution, and for bearing through the process of getting it merged!

shevtsiv:
Thank you for help, it was cool experience!

casey:
You are very welcome 🙇‍♂️

PR #440, opened 2019-05-15
Replace colored_diff with pretty_assertions

casey:
Colored diff has a bug which causes it to [panic in certain cases](https://github.com/CAD97/colored-diff/issues/3). For now, replace colored_diff with pretty_assertions.


soenkehahn (approved):
LGTM.

Issue #442, opened 2019-05-15
Make `--show ALIAS` also show the alias


casey:
If we have the following justfile:

```make
alias b := build
build: 
    echo 'Building!'
```

And do `just --show b`, we'll see the recipe, but not the alias:

```
$ just --show b
build: 
    echo 'Building!'
```

We should also show the alias:

```
$ just --show b
alias b := build
build: 
    echo 'Building!'
```

PR #443, opened 2019-05-15
Make `--show` work with aliases

### LINKED ISSUE #420: Make `--show` work with aliases
Given the justfile:

```
alias b = build

build:
  echo 'building!'
```

The following should work:

```
$ just --show b
```

casey:
Fixes #420. I opened issue #442 for also displaying the alias itself, which should be done in a follow-up PR.


casey (commented):
LGTM

PR #455, opened 2019-07-06
Remove deprecated tempdir dependency

NickeZ:
I'm trying to package just in debian and this dependency would be a blocker for that.

edit: This won't pass CI until a new release of brev has been published.


casey:
I think the failure on AppVeyor is an issue with rust's use of the MinGW toolchain on Windows. I just pushed a commit changing the toolchain in the appveyor config to MSVC toolchain (with static linking to the C runtime, since we had some issues with dynamically linked binaries before), so hopefully that works.

casey:
That fixed it. Merged, and thank you for the PR!

PR #458, opened 2019-07-08
Fix minimal versions

NickeZ:
Just doesn't actually compile with clap 2.0.0. I bumped it to 2.33.0 which is the current one in debian.

You can run the following command to generate a lockfile with minimal-versions:

```
cargo +nightly generate-lockfile -Z minimal-versions
```


casey:
Merged, thank you for the PR!

How did you notice that Just doesn't build with 2.0.0? Do you test with minimal versions as part of the packaging process?

NickeZ:
The special thing about Debian is that all dependencies have to be individually packaged. And obviously the Debian repositories can't be 100% up to date with crates.io. This means that quite often it won't be possible to build executables according to their checked in `Cargo.lock` file. So I figured out that there is an experiment going on to create "minimal versions lockfiles" instead. You can see [here](https://github.com/kpcyrd/cargo-debstatus/pull/3) in the screenshots how many fewer crates I need to package to get `just` into debian if I use the minimally acceptable versions instead of the most recent acceptable versions (which is `cargo update`s default).

casey:
Gotcha, that makes sense. I created #459, and I'll try to work out a process so as to prevent `Cargo.lock` getting out of date in the future.

In the screenshot, what does "invalid in debian" mean?

NickeZ:
It means that there is currently no version in Debian that satisfies what is required.

For example `just` requires `ctrlc` 3.1.2, but there is only 3.1.1 available in Debian. In this case there are two obvious solutions. Try downgrading the requirement of `just` to 3.1.1. Or try upgrading the packaged version of `ctrlc` to 3.1.X. In this case I think I would upgrade `ctrlc` because it only has ~70 dependencies and the latest version is 3.1.3.

casey:
Okay, interesting. This makes me think that I should attempt to lower all of just's dependency requirements in `Cargo.toml` to the minimum version that builds and passes tests, but which is compatible with the latest version. I.e if Just depends on `ctrlc` at 3.1.2 right now, and 3.1.2 is the latest release, I should ideally lower it to 3.0.0, assuming that I don't depend on a feature that was introduced in 3.1.0.

Is that reasonable?

All things being equal, I would like packaged versions of just to be built with the latest versions of dependencies, but it seems like enforcing that via `Cargo.toml` will possibly create a lot of work for packagers.

NickeZ:
Right, if you can lower it to 3.0.0 and it still builds that would simplify a lot for packagers. Because eventually one of 3.1.3, 3.1.2, 3.1.1, 3.1.0, 3.0.3, 3.0.2, 3.0.1, 3.0.0 will be packaged by debian and/or fedora and they probably won't even package the same version. (I assume that fedora has a similar packaging policy as debian.)

Obviously the distributions want to use as recent versions as possible, but there is simply more work to do than maintainers can keep up with. I'm not an expert in this, but I guess that to push out a new `ctrlc` version all dependencies has to be recompiled and retested, so also shortage of build machines could introduce delays.

NickeZ:
I suspect a good practice is to never depend on `PATCH` versions greater than 0. In your case it should be enough to depend on `3.1.0` to get the latest possible features. If everyone follows semver appropriately.

casey:
That seems reasonable. I opened #461, which minimizes required versions as much as possible.

PR #463, opened 2019-07-19
Generate basic man page with help2man

### LINKED ISSUE #456: Use `help2man` to generate interim man page
Eventually, I'd like to have a voluminous man page that includes:

- Documentation of command line syntax (autogenerated by clap)
- The contents of the current readme (generated using asciidoctor's roff backend)
- The just file grammar

However, this is waiting on a few things, mostly for me to get to it, and for clap to land man page generation support.

In the mean time, @NickeZ who's working on packaging `just` for Debian, mentioned that we could use the `help2man` utility to generate a serviceable man page from the output of `just --help`.

This seems like a good interim step. In order to make sure that the man page is consistent across packages, we should build it and check it in to the repository as `man/just.1`

casey:
Generates a basic man page with `help2man`. The generated man page doesn't look great by default, so this also adds the `help4help2man` feature, which makes Just print a help message that produces a better man page.

Fixes #456.


Alhadis:
You're better off writing it by hand, mate. Seriously.

There's no such thing as a quality generated man page. 😉

PR #465, opened 2019-08-21
Add compile target for linux-gnu

ghost:
This allows projects using the default travis Rust runner to work
out of the box.

Otherwise, you will see quite confusing errors like these:
https://travis-ci.org/google-apis-rs/generator/jobs/574675350#L197

"gzip: stdin: not in gzip format"
which leads to a closed issue in Trust indicating the required target
does not exist: https://github.com/japaric/trust/issues/68 .

The easiest fix would be to provide the target here.

Additionally, I think it will help the adoption 'just'.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
     # Linux
     #- env: TARGET=i686-unknown-linux-gnu
     #- env: TARGET=i686-unknown-linux-musl
-    #- env: TARGET=x86_64-unknown-linux-gnu
+    - env: TARGET=x86_64-unknown-linux-gnu
     - env: TARGET=x86_64-unknown-linux-musl
 
     # OSX


casey:
Thank you for opening this PR!

It looks like the trust download script is failing because:
- the script uses `rustc -Vv | grep host` to get the target binary to download
- on many linux systems (including on travis) the host returned by rustc is `x86_64-unknown-gnu-linux`
- Just's linux binary releases is `x86_64-unknown-linux-musl`

Just doesn't include an `x86_64-unknown-gnu-linux` release because:
- the `-gnu` target dynamically links to libc, which can cause problems
- it's an extra option which might confuse users looking for which binary to download for their system

I'm definitely keen to get the trust download scripts working, but I'd prefer not to add an extra release binary if it can be avoided. (Especially because, if I recall correctly, we had some issues caused by linking dynamically to libc.)

It looks like an [issue](https://github.com/japaric/trust/issues/120) regarding the install script using rustc was recently opened in the trust repo, and I added a [comment](https://github.com/japaric/trust/issues/120#issuecomment-524506448). Hopefully this can be fixed upstream, instead of adding a new release binary.

Until the trust script is improved, does the following command work for your use case?

```
curl -LSfs https://japaric.github.io/trust/install.sh | \
  sh -s -- --git casey/just --target x86_64-unknown-linux-musl --to DESTINATION_DIRECTORY
```

I'm going to close this PR for now, but let me know if the workaround above doesn't work for you.

ghost:
Hi @casey , thanks a lot for that fantastic community work of yours! This reply is downright fantastic - super helpful, and in a writing style I envy absolutely!

The suggested workaround will work for me, and your proposed solution is the way to go :)!

casey:
Awww, thank you for the kind words, I really appreciate it! Getting positive feedback like is part of what makes working on open source software rewarding for me ^_^

Also, glad that my suggestion works for you! Hopefully the next iteration of the trust install script will find `-musl` and `-gnu` binaries.

Issue #438, opened 2019-05-15
Format multiline strings in integration tests with indoc


casey:
With [indoc](https://github.com/dtolnay/indoc), instead of:

```rust
integration_test! {
  name: alias_shadows_recipe,
  justfile: "bar:\n  echo bar\nalias foo := bar\nfoo:\n  echo foo",
  args: (),
  stdin:  "",
  stdout: "",
  stderr: "error: Alias `foo` defined on `3` shadows recipe defined on `4`
  |
3 | alias foo := bar
  |       ^^^
",
  status: EXIT_FAILURE,
}
```

We can do:

```rust
integration_test! {
  name: alias_shadows_recipe,
  justfile: "bar:\n  echo bar\nalias foo := bar\nfoo:\n  echo foo",
  args: (),
  stdin:  "",
  stdout: "",
  stderr: indoc!("
    error: Alias `foo` defined on `3` shadows recipe defined on `4`
      |
    3 | alias foo := bar
      |       ^^^
  "),
  status: EXIT_FAILURE,
}
```

For extra magic, we can pass all stdout/stderr strings to `indoc!()` in the `integration_test!()` macro.

casey:
Fixed by #481.

PR #477, opened 2019-10-04
Start to use indoc! for integration tests

aquarhead:
I decided to take a look at #438 

Starting with a single change just to make sure things make sense. I tried to use `indoc!` in the `integration_test` macro rule, but as it stand it requires the input to be a string literal.

I'm thinking maybe we should just add a function for this conversion rather than pulling in many dependencies for a seemingly simple thing?

Waiting for some feedback before more work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -220,6 +220,7 @@ dependencies = [
  "target 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempfile 3.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "unicode-width 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "unindent 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -494,6 +495,11 @@ version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "unindent"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "utf8-ranges"
 version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


casey:
Thanks for getting started on this!

A few thoughts:

- I landed #479, which simplifies the integration tests slightly by only running them with `bash`. Currently, the tests run with `bash`, `dash`, and whatever's installed as `sh`. Removing them makes the tests run faster, and should make this PR a little simpler.

- The functionality of `indoc`  is available in the `unindent` crate as a function that doesn't require a string literal, so we could use that instead.

- There might be a complication with `unindent`: `unindent::unindent()` strips trailing newlines, but Just output and the expected output strings often have trailing newlines, so `unindent` will cause the tests to fail.

For the last issue, we could write our own version of unindent (like you mentioned) that additionally doesn't strip trailing newlines. What do you think?

aquarhead:
Ah~ I didn't notice the `unindent` crate 🙄 That crate's layout is a little bit... messy.

I will check out the `unindent` function and see whether it fits us, or else I will reimplement it. I'll also rebase to the latest master.

aquarhead:
Also, if we're not using `dash` anymore we should just remove the related steps in CI? https://github.com/casey/just/blob/633f0ccaa6da83e13b9dbd87b1cdd9f371191bcc/.travis.yml#L35-L38 Could save about 5min I think

aquarhead:
So I switched to use `unindent`, added a helper function to ensure the formatted string ends with a newline (if not empty string).

One "major" change is that I switched several `printf` to `echo`, in several backtick and raw string related tests. In these test the `stdout` are driven by users not by `just`, and `printf` will not append newline by default. An example is this: https://github.com/casey/just/pull/477/files#diff-2a2d8cb84eb7f679ed0fdc62e67e9071R423-R428

This may not be the best solution, since now there's this implicit relationship between using `echo` in test and the expected `stdout`. An alternative solution (that I actually tried/implemented) is to have another option for constructing the `Test` struct, like `stdout_by_user` to indicate that the `stdout` is driven by the user and not something formatted by `just`. (I'd think `stderr` are always controlled by `just`?) I'm not so sure whether it's worth this extra effort though.

All in all, this is again a WIP (although very close to finish), if the current pass feels good then I will take a final pass to ensure all strings are converted (so many of them xD)

Also, should we also convert all `justfile` strings to use this format? Since quite often we compose multiline `justfile`s.

casey:
> Also, if we're not using `dash` anymore we should just remove the related steps in CI?

Good catch! I opened #482 to fix this.

I merged #481, which includes a custom implementation of `unindent()` that allows for gradual migration of test strings. It only modifies the string if common indentation was found, which makes all the tests pass without modification. It also strips a maximum of one leading and one trailing blank line, which allows trailing blank lines where necessary, which isn't possible with the `unindent::unindent`.

Sorry for going ahead and writing it myself, I started thinking about the problem, and realized that an implementation that would allow gradual migration would be pretty fiddly, so I went ahead and wrote it.

I only changed a few strings, so new tests can use indented strings, and old tests can be migrated as needed. Changing the old strings will probably be pretty tedious, so definitely don't change them unless you're feeling masochistic 😅

aquarhead:
No problem, do you still see any value in this PR? or should I just close it?

casey:
I pulled the string formatting changes out into a separate diff, and merged it in #489.

I tried to keep the commit as authored by you, since it was all your changes, but the squash and merge clobbered the original authorship information, so I credited you in the [commit message](https://github.com/casey/just/commit/64ed94c8ac77fc40eaa22cf71fea625caf40c9d0).

Issue #387, opened 2019-03-18
function to get Justfile location path


camerondavison:
I created a super simple `Justfile`
```
info:
	touch file.empty
```
then created a nested project.
```
mkdir -p a/b/c
```
then ran 
```
just --justfile Justfile --working-directory .
just --justfile Justfile --working-directory a
just --justfile Justfile --working-directory a/b
just --justfile Justfile --working-directory a/b/c
```
and got
```
$ tree
.
├── Justfile
├── a
│   ├── b
│   │   ├── c
│   │   │   └── file.empty
│   │   └── file.empty
│   └── file.empty
└── file.empty
```
I am going to assume that this is expected, but given https://github.com/casey/just/issues/312 and the `invocation_directory` function it seems like there is not a great way to do something like
```
ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
```
that you can do in make to allow paths that are absolute, but relative to the Justfile.

I would like to make paths absolute so that I do not depend on someone calling `--justfile` with the correct `--working-directory`

casey:
This should be easy to support. If you wouldn't mind elaborating on your use case, that would be great, to make sure the feature actually does what you want to do.

Functions are defined in `functions.rs`. They get passed a `FunctionContext`, which contains the information they need to function. (Haha.) It would be pretty easy to add a function that returns the directory the `justfile` was found in.

camerondavison:
the use case is I have something like
```justfile
clean:
     rm -rf build
```
I would prefer to write
```justfile
clean:
     rm -rf "{{source_dir()}}/build"
```
or even better
```justfile
project_dir = source_dir()
build_dir = project_dir + "/build"

clean:
     rm -rf "{{build_dir}}"
```

casey:
That makes sense.

Currently, just requires `--working-directory` when `--justfile` is set. This should be changed so that you can pass `--justfile` without `--working-directory`, issue #340. There's actually an open PR, #355, that just needs an integration test before it lands.

Why is the justfile being called with `--justfile` and invoked by letting `just` find the justfile itself and setting the working directory? Would not requiring `--working-directory` with `--justfile` be a good fix?

camerondavison:
My thinking is that people "could" call it that way. And I would never want to accidentally delete something in the wrong working directory. Which is why I always like to make sure the path is absolute. This can be done by making it relative to the justfile but then knowing the absolute location of the justfile.

casey:
[...]

PR #478, opened 2019-10-05
Add `justfile_directory()`

aquarhead:
This is for #387

I'm in the process of adding test for it.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -387,6 +387,25 @@ build:
     cd {{invocation_directory()}}; ./some_script_that_needs_to_be_run_from_here
 ```
 
+==== Justfile Directory
+
+- `justfile_directory()` - Retrieves the parent directory of the justfile being executed.
+
+For example, to call `rustfmt` on files under the same directory of the justfile, use the following rule:
+
+```
+rustfmt:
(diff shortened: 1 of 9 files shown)


aquarhead:
Added some test in similiar fashion of `test_invocation_directory()` However it currently fails on OS X like this:

```
---- test_justfile_directory stdout ----
bad stdout:
got:
"/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/just-test-tempdirIJaX0l\n"
expected:
"/private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/just-test-tempdirIJaX0l\n"
```

My guess is that this happens because the temporary path (in OS X) is observed differently from child process (the executed `just` process) and in the test process (or `fs::canonicalize` I guess?).

Don't really have a good idea to deal with this yet, any suggestions?

casey:
Thanks for working on this!

It looks like on macOS, `/var` is a symlink that points to `/private/var`. Is one of the paths being canonicalized and the other isn't? `/var` will canonicalize to `/private/var`, so if it isn't being canonicalized in the test code, but is being canonicalized inside of Just, that could explain it.

casey:
I'm going to close this for now, please feel free to re-open it if you pick it up again!

PR #493, opened 2019-10-13
Add repository attribute in Cargo.toml

SOF3:
docs.rs shows the GitHub page iff `package.repository` is provided in Cargo.toml

Otherwise it just shows "Homepage", and doesn't display stars, issues, etc.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ description = "🤖 Just a command runner"
 authors     = ["Casey Rodarmor <casey@rodarmor.com>"]
 license     = "CC0-1.0"
 homepage    = "https://github.com/casey/just"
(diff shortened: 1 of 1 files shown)


casey:
Thank you for the PR!

Are you sure that the `.git` extension should be there? For example, in Cargo's [Cargo.toml](https://github.com/rust-lang/cargo/blob/master/Cargo.toml#L10), the repository key is set to the  GitHub URL without the `.git` extension.

SOF3:
Perhaps not using the `.git` is better, I suppose. For GitHub though, both are equivalent, because cloning without `.git` will automatically redirect, and accessing repo webpage with `.git` will automatically redirect as well

casey:
Sweet, merged!

PR #496, opened 2019-10-16
Rename: Parser -> Analyzer

casey:
This change is in preparation for a future change, which will replace
the current `Parser`, which both parses and performs multiple kinds of
analysis, with a separate parser, which will produce an AST, and
one or more analyzers, which perform the other consistency checks that
are performed by the current parser.


casey:
Closing, this will be included in a separate PR.

Issue #289, opened 2017-12-15
Allow dependencies that take parameters


fuine:
I'm not sure if this is possible, but consider the following case:
```
run stuff:
    cd {{stuff}} && python main.py

visualize stuff:
    python src/visualize.py {{stuff}}

run_visualize stuff:
    just run {{stuffl}}
    just visualize {{stuff}}
```

Is there a way to create `run_visualize` recipe without subsequent executions of `just` from the main `just` process?

casey:
Is there a good default value you could give to `stuff`? If so, you could turn it into a variable:

```
stuff = "foo"

run:
    cd {{stuff}} && python main.py

visualize:
    python src/visualize.py {{stuff}}

run_visualize: run visualize
```

You could then override it on the command line like so:

`just stuff=bar run`

fuine:
I'm afraid I can't, that's the whole problem, I thought about using the default value but I want to make sure that this particular value is always passed by user

casey:
Hmm, darn, this is a tricky one.

The only solution I can think of is one which would be pretty weird. Basically to allow dependencies that take the exact same number of arguments, and pass the arguments of the dependee through, which would allow:

```make
run stuff:
    cd {{stuff}} && python main.py

visualize stuff:
    python src/visualize.py {{stuff}}

run_visualize stuff: run visualize
```

This would be preeeeetty magical though. An additional constraint could be that they must have the same name, which would mean that the above would still work, but the following wouldn't:

```make
run bar:
    cd {{bar}} && python main.py

visualize foo:
    python src/visualize.py {{foo}}

run_visualize stuff: run visualize
```

Since `foo` != `stuff` and `bar` != `stuff`.

I can't decide whether this is super weird or not.

What do you think? Also, anyone else, would you use this or has this been an issue before?

fuine:
I'm not really sure this is a good idea, because that would break backward compatibility, consider
```
foo arg: bar baz
```
what if `bar` and `baz` don't take any arguments? I think a better option would be to allow the user to provide the arguments somehow so maybe something like
```
foo arg1 arg2: (bar arg1 arg2) (baz arg2)
```
What do you think?

fuine:
When I think about it now, maybe it's a better idea to introduce something like that within the body of the recipe? I'd think that a lot of the times you want to run the recipes somewhere in the middle of different operations like so:
```
foo arg1 arg2:
    # some bash commands here
    bar arg1 arg2  # run recipe bar here
    # more bash commands
[...]

PR #555, opened 2019-12-03
Allow passing arguments to dependencies

casey:
I've started working on passing arguments to dependencies, see #289, a longstanding feature request.

Open questions to resolve:

## Syntax

- 1. Lisp style:

```
bar name options:
  ./build {{name}} {{options}}

foo: (bar "arg" "baz")
```

- 2. Fortran style:

```
bar name:
  ./build {{name}}

foo: bar("arg", "baz")
```

Lisp can be pretty incomprehensible, for example:

```
foo: (bar 'a' + 'b' 'c' + 'd')
```

As opposed to:

```
foo: bar('a' + 'b', 'c' + 'd')
```

However, I think in some sense lisp style is more natural, since recipe invocation on the command line doesn't require parenthesis or commas. At the moment I'm leaning towards lisp style.

## Repeat dependency invocations

Just, like Make, will only run a recipe once per invocation, even if it appears multiple times in a dependency tree. For example, given the following, `just bob` will run `bar` only once, even though `foo` and `baz` both depend on it:

```make
bar:

foo: bar

baz: bar

bob: foo baz
```

However, the situation is complicated by arguments. Consider:

```
build target:
  ./build {{target}}

release: (build "foo") (build "bar")
  ./release
```

Clear, the user intends to run `build` twice, once for "foo" and once for "bar".

I think a reasonable approach would be to run recipes once per recipe and combination of arguments. So given the following justfile, `just release` would run `build "foo"`, `build "bar"` and `build "baz"`:

```
build target:
  ./build {{target}}

build-foo: (build "foo") (build "bar")

build-baz: (build "baz") (build "bar")

release: build-foo build-baz
```

I can't immediately think of any counter-examples where this behavior wouldn't be desirable, so barring any counter examples, this is what I'll implement.

## To Do

- [x] error on undefined variables in arguments
- [x] allow recipes with parameters in dependencies
- [x] error if wrong number of arguments are passed to dependency
- [x] test assignments can be used in arguments
- [x] test parameters can be used in arguments
-  [x] test functions can be used in arguments
-  [x] test backticks can be used in arguments
- [x] dependency memoization is correct (recipe with same args run once, different args run multiple times)
- [x] update readme


clarfonthey:
I also think the lisp style is the way to go.

runeimp:
I'm good with either but I'd say the "lowest common denominator" might be the Fortran style. Most languages I've dealt with handle functions in the same manner that you have Fortran listed. Were as the LISP style is similar to calling functions in a subshell. So if your coming from anything other than LISP (and Scheme?) development or shell scripting the Fortran style will be more natural.

casey:
@clarfon and @runeimp, thanks for the feedback! I decided to go with the lisp-style, just because it felt more natural, and to make it visually distinct from function invocations.

As for duplicate recipe invocations, Just will run every combination of recipe and arguments at most once.

For example in:

```make
foo: (bar "A") (bar "A")
bar x:
```

`just foo` will run `bar` once with the argument "A".

PR #606, opened 2020-03-13
Update zsh completion file

heyrict:
I just modified the clap-generated completion file for `zsh` to make it working.

Note that completion for `--set` calls `just --evaluate` to get the variables, so it can slow down the console a bit if you have heavy evaluation logic.

### DIFF
--- a/completions/just.zsh
+++ b/completions/just.zsh
@@ -13,7 +13,6 @@ _just() {
         _arguments_options=(-s -C)
     fi
 
-<<<<<<< HEAD
     local common=(
         '--color=[Print colorful output]: :(auto always never)' 
         '(-f --justfile)'{-f+,--justfile=}'[Use <JUSTFILE> as justfile.]' 
@@ -58,7 +57,7 @@ _just() {
 
             if [[ ${lastarg} = */* ]]; then
                 # Arguments contain slash would be recognised as a file
-                _arguments -s -S $common '::ARGUMENTS -- Overrides and recipe(s) to run, defaulting to the first recipe in the justfile:_files' 
+                _arguments -s -S $common '*:: :_files'
             else
                 # Show usage message
                 _message $args_str
--- a/completions/just.zsh
+++ b/completions/just.zsh
@@ -75,15 +75,6 @@ _just() {
     return ret
 }
 
-(( $+functions[_just_variables] )) ||
-_just_variables() {
-    local variables; variables=(
-        ${(s: :)$(_call_program commands just --variables)}
-    )
-
-    _describe -t variables 'variables' variables
-}
-
 (( $+functions[_just_commands] )) ||
 _just_commands() {
     local commands; commands=(
@@ -91,6 +82,15 @@ _just_commands() {
     )
 
     _describe -t commands 'just commands' commands "$@"
+}
+
+(( $+functions[_just_variables] )) ||
+_just_variables() {
+    local variables; variables=(
+        ${(s: :)$(_call_program commands just --variables)}
+    )
+
+    _describe -t variables 'variables' variables
 }
 
 _just "$@"
--- a/completions/just.zsh
+++ b/completions/just.zsh
@@ -5,7 +5,7 @@ autoload -U is-at-least
 _just() {
     typeset -A opt_args
     typeset -a _arguments_options
-    local context curcontext="$curcontext" state line ret=1
+    local ret=1
 
     if is-at-least 5.2; then
         _arguments_options=(-s -S -C)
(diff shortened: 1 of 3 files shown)


casey:
Thank you for doing this!

I'd like to avoid running `just --evaluate` during completions. If I add functionality to just to print out a space-separated list of variable names, would that allow removing the call to `just --evaluate`, in favor of `just --variables`? For example:

```
x := "hello"
y := "bye"
```

```
$ just --variables
x y
```

The existing ZSH completions file is auto-generated, so to merge this I'll have to write a little code inside of the `just --completions` subcommand which modifies the generated zsh completions file with the additions in this PR. That will be easier if the modifications are simple insertions and deletions. So, for example, avoiding modifying or reformatting the block that starts with:

```
local context curcontext="$curcontext" state line
    _arguments "${_arguments_options[@]}" \
```

Would be ideal, and the fewer changes to the generated existing code, the easier that will be. Would it be possible to refactor this to introduce slightly less churn to the generated code?

casey:
I just landed #608, which adds a `just --variables` subcommand that prints out a space-separated list of variable names. Unfortunately, this also caused the generated completion scripts to change, which introduced conflicts with this PR.

heyrict:
Oh, sorry. I thought the completion files are maintained manually since they are added to VCS. Should I modify the `bin/generate-completions` script instead? I'm not familiar with editing files with command line except with `vi -c`. 

Anyway, here are the list of generated _arguments that should be changed:

- `'-s+[..]' ` should be changed to `'-s+[..]: :_just_commands'`, so is with `--show=[..]`
- `'*--set=[..]' ` should be changed to `'*--set[..]: :_just_variables' `
- Fallback to complete files `'::ARGUMENTS...:_files'` should be moved afterwards and the array should be assigned to a variable `$common` as it is used twice. I thought this is the hardest part.

casey:
> Oh, sorry. I thought the completion files are maintained manually since they are added to VCS.

No worries, it's a somewhat convoluted setup. The completion scripts are generated by clap, the command line argument parsing library that just uses. The `just` binary has a subcommand, `just --completions SHELL`, which prints the completions from clap to stdout. I also decided to save the completions to the `completions` directory, so that packagers will be more likely to know that they exist, and won't have to know to run `just --completions` to get them.

> Should I modify the `bin/generate-completions` script instead? I'm not familiar with editing files with command line except with `vi -c`.

This is the implementation of `just --completions`: https://github.com/casey/just/blob/1dbc7653904e9094faeeb0b8745c73ba325571c7/src/config.rs#L462

So I'll have to edit that function to modify the generated completions file with your changes.

I think the best thing to do is to modify this PR to have fewer diffs, and then I'll modify the completions function so that it produces the right output.

For example, there's this diff:

```diff
-    local ret=1
+    local context curcontext="$curcontext" state line ret=1
 
     if is-at-least 5.2; then
         _arguments_options=(-s -S -C)
     else
         _arguments_options=(-s -C)
     fi
 
-    local context curcontext="$curcontext" state line
```

I think these changes could be avoided, since they move around existing code.

> Anyway, here are the list of generated _arguments that should be changed:
> 
> * `'-s+[..]' ` should be changed to `'-s+[..]: :_just_commands'`, so is with `--show=[..]`
> * `'*--set=[..]' ` should be changed to `'*--set[..]: :_just_variables' `
> * Fallback to complete files `'::ARGUMENTS...:_files'` should be moved afterwards and the array should be assigned to a variable `$common` as it is used twice. I thought this is the hardest part.

That makes sense. What I'll probably do is create an `arguments_wrapper` function that looks like this:

```
arguments_wrapper() {
  _arguments $@ # pass arguments to _arguments
  common=$@
}
```

And then replace the call to `_arguments` with a call to `arguments_wrapper`, to minimize the amount of code that needs to change.

heyrict:
@casey Thank you for your response! I've just refactored the PR to include minimal changes to the completion file. Hopefully it will be easier to add.

> What I'll probably do is create an arguments_wrapper function that looks like this

Well, sorry I didn't get the point how this function would work. I left the `$common` variable as it was.

casey:
I just pushed some diffs to your branch which modify the completions subcommand, now in `src/subcommand.rs`, to modify the ZSH completions script that it generates. The completions file shouldn't have changed too much, although I modified it a little bit to make it easier to create.

Is there any way to test ZSH completion scripts?

Also, I've never actually used ZSH completions. How do I tell zsh that I want to load the completion script?

heyrict:
Thanks for your fast work!

> Is there any way to test ZSH completion scripts?

Well, I don't know any automated approach to test a completion script. Most of the time I test the script by inserting lines e.g. `echo $somevar >> debug.txt` and `tail -F debug.txt`

> How do I tell zsh that I want to load the completion script?

Afaik zsh uses `fpath` variable to get the location where completion scripts lies. Completion script for `just` should be placed in `{compdir}/_just` where `{compdir}` is a member of `fpath`.

By convention completion scripts are placed in `.zsh/completions` or so. I'll use `/tmp/completions` here to minimize the effect done to your existing configs.
```zsh
mkdir -p /tmp/completions
cp /path/to/just/completions/just.zsh /tmp/completions/_just
```

After that, simply paste the following to your zsh or `.zshrc` will make zsh completion work.

```zsh
fpath+=/tmp/completions
# The styles for completion descriptions and messages, zsh will hide them by default.
zstyle ':completion:*:descriptions' format "%U%B%d%b%u"
zstyle ':completion:*:messages' format "%F{green}%d%f"
autoload -Uz compinit
compinit -u
```

casey:
> Thanks for your fast work!

You bet!

> Well, I don't know any automated approach to test a completion script. Most of the time I test the script by inserting lines e.g. `echo $somevar >> debug.txt` and `tail -F debug.txt`

I searched a little bit, and that's my impression too, i.e. that it doesn't seem to be common to test zsh completion scripts. I'd like to add that eventually, since generating completion scripts from clap and then modifying them in code feels error-prone, but that can wait.

> After that, simply paste the following to your zsh or `.zshrc` will make zsh completion work.

Sweet, worked for me. Pretty cool! It's awesome to see the recipes names and variable values complete. Also rad that it prints the whole recipe in green.

Thanks so much for this! Better completions scripts is a frequent request, so this will definitely make zsh users happy.

Issue #445, opened 2019-05-15
Suggest recipe aliases


casey:
Currently, just will only suggest recipes when it doesn't recognize an argument. It should also suggest aliases:

```
$ just bd
error: Justfile does not contain recipe `bd`.
Did you mean `b`, an alias for `build`?
```

casey:
Fixed in #624!

PR #624, opened 2020-04-19
Show aliases in suggestions

Celeo:
https://github.com/casey/just/issues/445


src/justfile.rs
```diff
@@ -28,16 +28,32 @@ impl<'src> Justfile<'src> {
...
+          Some(alias.target.name.lexeme()),
+        )
+      })
+      .collect();
+    available.append(&mut aliases);
+    available.sort_by(|a, b| a.1.cmp(&b.1));
+    if let Some(&(suggestion, distance, alias_for)) = available.first() {
       if distance < 3 {
-        return Some(suggestion);
+        return if let Some(target) = alias_for {
```
Celeo: It would be simpler (here, at least) to just conditionally add the "an alias of Y" string to aliases, and then not have to change this function's return type (and cascade that change) but that runs into fighting the borrow checker for creating a new `String` from `format!` inside this method, and then returning that, which doesn't adhere to the lifetime of the struct.

If you don't mind returning a `String` from this method, then that's a simple update.

casey: It's a little more cumbersome, but I often write types that capture things like this. Here's a possibility:

```
struct Suggestion<'src> {
  name:   &'src str,
  target: Option<&'src str>,
}

impl<'src> Display for Suggestion<'src> {
  fn fmt(&self, f: &mut Formatter) -> fmt::Result {
    write!(f, "Did you mean `{}`", self.name)?;
    if let Some(target) = self.target {
      write!(f, ", an alias for `{}`", target)?;
    }
    write!(f, "?")?;

    Ok(())
  }
}
```

And then suggest returns an `Option<Suggestion<'src>>`, and anything that gets a suggestion can just `eprintln("{}", suggestion)`.

src/justfile.rs
```diff
@@ -28,16 +28,32 @@ impl<'src> Justfile<'src> {
     self.recipes.len()
   }
 
-  pub(crate) fn suggest(&self, name: &str) -> Option<&'src str> {
-    let mut suggestions = self
+  pub(crate) fn suggest(&self, name: &str) -> Option<(&'src str, Option<&'src str>)> {
+    let mut available: Vec<(&str, usize, Option<&str>)> = self
```
casey: This can be cleaned up by chaining iterators, something like:

```
  let mut suggestions = self
    .recipes
    .keys()
    .map(|name| {
      (edit_distance(name, input), Suggestion {
        name,
        target: None,
      })
    })
    .chain(aliases.iter().map(|name, target| {
      (edit_distance(name, input), Suggestion {
        name,
        target: Some(target),
      })
    }))
    .filter(|(distance, _suggestion)| distance < 3)
    .collect::<Vec<Suggestion>>();

  suggestions.sort_by_key(|(distance, _suggestion)| distance);

  suggestions.first()
```

casey:
Thanks for the PR! Added some comments inline.

src/justfile.rs
Celeo: Ah, thanks for putting me onto `.chain`!

src/common.rs
```diff
@@ -48,23 +48,66 @@ pub(crate) use crate::{
 
 // structs and enums
 pub(crate) use crate::{
-  alias::Alias, analyzer::Analyzer, assignment::Assignment,
```
Celeo: It's a big formatting change, but it's what I got from running the exact formatting commands you have in the GitHub Actions config file. 🤷‍♂️

casey: That's really strange, are you running `cargo +nightly fmt --all`?

What do `cargo +nightly --version` and `cargo +nightly fmt --version` show?

I get the following:

```
: cargo +nightly --version
cargo 1.44.0-nightly (ebda5065e 2020-04-16)
: cargo +nightly fmt --version
rustfmt 1.4.14-nightly (a5cb5d26 2020-04-14)
```

Celeo: Yup!

```plain
just on  issue_445 is 📦 v0.5.10 via 🦀 v1.42.0 
❯ cargo +nightly fmt --all

just on  issue_445 is 📦 v0.5.10 via 🦀 v1.42.0 
❯ cargo +nightly fmt --all -- --check

just on  issue_445 is 📦 v0.5.10 via 🦀 v1.42.0 
❯ echo $?
0

just on  issue_445 is 📦 v0.5.10 via 🦀 v1.42.0 
❯ git status
On branch issue_445
Your branch is up to date with 'origin/issue_445'.

nothing to commit, working tree clean
```

casey: Hmm, super weird, and how about the versions from `cargo +nightly --version` and `cargo +nightly fmt --version`? (I can see cargo  is at 1.42, but the nightly version should be different.)

Celeo:
```plain
❯ cargo +nightly --version
cargo 1.44.0-nightly (ebda5065e 2020-04-16)

~ 
❯ cargo +nightly fmt --version
rustfmt 1.4.14-nightly (a5cb5d2 2020-04-14)
```

casey: Ah, okay, I figured it out, it's because of the braces in `justfile::{Justfile, Suggestion}`. When an item in a use block has braces, rustfmt switches to putting each item on its own line. You can put suggestion in its own module, i.e. `src/suggestion.rs`. I usually just put everything in its own module, since it's slightly easier to jump to.

Celeo: Ahhh, good find!

src/justfile.rs
```diff
@@ -28,19 +28,31 @@ impl<'src> Justfile<'src> {
...
+        (edit_distance(alias_name, name), Suggestion {
+          name:   alias_name,
+          target: Some(alias.target.name.lexeme()),
+        })
+      }))
+      .filter(|(distance, _suggestion)| distance < &3)
+      .collect::<Vec<(usize, Suggestion)>>();
+    suggestions.sort_by_key(|(distance, _suggestion)| *distance);
+
+    if let Some((_distance, suggestion)) = suggestions.first() {
```
casey: This is totally fine, but you could also do:

```
suggestions.into_iter().map(|(_distance, suggestion)| suggestion).next()
```

I think that `into_iter().next()` will give you the value, and not a reference to the value, so you don't need to derference with `*`.

src/justfile.rs
```diff
@@ -28,19 +28,31 @@ impl<'src> Justfile<'src> {
     self.recipes.len()
   }
 
-  pub(crate) fn suggest(&self, name: &str) -> Option<&'src str> {
+  pub(crate) fn suggest(&self, name: &str) -> Option<Suggestion> {
```
casey: This is totally fine, but if you use a different name for the parameter, then you can more easily assign to `Suggestion::name`. using the `Foo{field}` shorthand.

So if it's:

```rust
pub(crate) fn suggest(&self, input: &str) -> Option<Suggestion>
```

Then later on you can do:

```rust
recipe
.keys()
.map(|name| {
 (edit_distance(input, name), Suggestion { name, target: None})
})
```

And then later on:

```rust
aliases
.iter()
.map(|name, alias| {
 (edit_distance(input, name), Suggestion { name, target: Some(alias.target.name.lexem())})
})
```

casey (commented):
A couple of very minor comments. Also, this could use an integration test that checks the output is correct.

If you look in `tests/integration.rs`, you can find a test called `show_suggestion`, which you can make a copy of that adds an alias and checks that it suggests the alias.

casey (changes requested):
I forgot to actually hit the `Request changes` radio button, doing that now, which forces me to leave this content-free comment :P

src/justfile.rs
Celeo: Oh good point. I'd thought of doing the map before, but I was worried about needlessly applying the function to the entire list when I just needed the first element. I'd forgot that Rust uses lazy processing there, so it's not a problem.

casey: It's still wild to me that many uses of closures in rust are as efficient as they are. In my head I still think of things like closures, map, etc, as being spooky, inefficient things that you can only use if you don't care about performance.

Celeo: Haha definitely! I've been using JS too much.

Celeo:
Heya,

Anything else I can do for this?

casey:
Just an integration test! 

> A couple of very minor comments. Also, this could use an integration test that checks the output is correct.
> 
> If you look in `tests/integration.rs`, you can find a test called `show_suggestion`, which you can make a copy of that adds an alias and checks that it suggests the alias.

This will mostly be copy pasta from the existing `show_suggestion` test, but should check that the actual text output when an alias is suggested is correct.

Celeo:
Sure thing! I remember seeing that request earlier, but must've hidden it or something.

casey:
Merged! Thank you very much for this!

By the way, I don't actually have a changelog generator set up this repo,, so no commit metadata necessary.

And, I'm still merging with the "squash and merge" button on Github, since I don't (yet!) have a policy of signing all commits on this repo (that's a new thing I'm trying out for Intermodal).

Celeo:
Great, happy to contribute!

PR #625, opened 2020-05-02
Adds dynamic recipe completions for fish

tyehle:


casey:
This is great, thanks for this!

Will `complete -c just -f` disable file completions after recipes? Just recipes can have arguments, which might be paths to files.

tyehle:
Yeah it does. Changed it to only disable files before any subcommand.

casey:
Also, will this interfere with options like `--justfile`, which take a file path?

tyehle:
Haha, yeah that is an issue and is not fixed. I'll take a look at the fish docs a bit more to see if there is something to take care of it.

casey:
What do you think about actually landing this with `--force-files` set? This would be conservative in the sense that we could be sure that it would never break path completion where it was wanted, at the cost of sometimes enabling path completion where it wasn't wanted.

I think this would still be a huge improvement on the existing completions, and tightening them up could be left for later.

tyehle:
Makes sense and works for me

tyehle:
Looking at this again I realized that:
1. Recipes were being disabled after the first one, which shouldn't happen
2. There is an easy way to not suggest files for only the first argument and then enable them after that. I think this is a pretty good usable conservative approach.

Both of these issues should be fixed now

casey:
Awesome, merged. Thank you very much!

Issue #628, opened 2020-05-24
Make slash continuations resolve to a single space


casey:
Recipe slash continuations preserve leading whitespace on the line after the slash. When commands with line continuations are echoed, they contain this extra leading whitespace, which makes them harder to read.

For example, this recipe line:

```
help2man \
  --name 'save and run commands' \
  --manual 'Just Manual' \
  --no-info \
  target/debug/just \
  > man/just.1
```

Gets echoed as:

```
help2man        --name 'save and run commands'  --manual 'Just Manual'  --no-info       target/debug/just     > man/just.1
```

It would be nice to echo whitespace, followed by a line continuation, followed by leading whitespace on the next line as a single space.

This is unlikely to break any recipes, but I think this should be considered a breaking change. It won't change commands, but if line continuations are used within strings, then this would change the amount of whitespace inside those strings.

rjsberry:
It would be great if we could also partially preserve indentation.

Given the recipe:

```make
say-hello:
  docker run alpine \
    sh -c ' \
      speak=hello; \
      echo $speak \
    '
```

I would expect this to be echoed as:

```
$ just say-hello
docker run alpine \
 sh -c ' \
  speak=hello; \
  echo $speak \
 '
hello
```

Rather than:

```
$ just say-hello
docker run alpine \
 sh -c ' \
 speak=hello; \
 echo $speak \
 '
hello
```

casey:
Thanks for the feedback! I think I personally would prefer it to be echoed as 

```
docker run alpine sh -c ' speak=hello; echo $speak'
```

Just to avoid taking up too much space in the terminal. I think it's definitely arguable either way. I'm curious what other people think.

Is there a difference between the last two example outputs?

rjsberry:
My mistake - I had interpretted the original issue as preserving the newline in the recipe. If that is not the case, trimming the the whitespace on the single line of output is certainly an improvement.

The difference in the example outputs is simply the indentation inside the command string run by `sh`.

casey:
> My mistake - I had interpretted the original issue as preserving the newline in the recipe. If that is not the case, trimming the the whitespace on the single line of output is certainly an improvement.

Ahh, gotcha.

> The difference in the example outputs is simply the indentation inside the command string run by `sh`.

Ooop, I see it now!

casey:
Fixed in #635. This is breaking change, so next release will bump the version to `0.6.0`.

PR #632, opened 2020-06-04
Make line continuations resolve to a single space

rjsberry:
See discussion in #628 

* Moves recipe directive parting into a separate structure
* Exposes methods to access both the raw evaluated str and pretty str


rjsberry:
I've kept the formatting separate for what we show when the recipe is invoked, and the string that is actually used to run the recipe. We can collapse these into one, but this is definitely a breaking change.

casey:
Thanks for the PR!

I want to avoid there being a difference in what's displayed versus what's actually executed. If there's a discrepancy, then I think it would get very confusing to reason about and debug.

I just opened #635, which is basically just the minimum possible diff to implement the feature. The code is gross and could use a refactor, as you've done, but I wanted to do it with the minimum amount of churn possible. 

I'm super sorry to close in favor of my own PR. I know that's a super annoying response to a PR.

rjsberry:
Sure - I intended this PR as a starting point rather than a request for immediate merge.

> I want to avoid there being a difference in what's displayed versus what's actually executed.

Completely agree. I was unsure of your desired design from the issue so I went with a conservative impl. FWIW I was fishing for feedback and would have been happy to update the implementation.

PR #635, opened 2020-06-09
Strip leading spaces after line continuation

casey:
Make line continuations strip leading spaces on the next line.

This changes what is passed to the shell, so this is a breaking change.
However, I don't think that this will break any recipes.

### DIFF
--- a/src/evaluator.rs
+++ b/src/evaluator.rs
@@ -145,11 +145,20 @@ impl<'src, 'run> Evaluator<'src, 'run> {
     })
   }
 
-  pub(crate) fn evaluate_line(&mut self, line: &Line<'src>) -> RunResult<'src, String> {
+  pub(crate) fn evaluate_line(
+    &mut self,
+    line: &Line<'src>,
+    continued: bool,
+  ) -> RunResult<'src, String> {
     let mut evaluated = String::new();
-    for fragment in &line.fragments {
+    for (i, fragment) in line.fragments.iter().enumerate() {
       match fragment {
-        Fragment::Text { token } => evaluated += token.lexeme(),
+        Fragment::Text { token } =>
+          if i == 0 && continued {
+            evaluated += token.lexeme().trim_start();
+          } else {
+            evaluated += token.lexeme();
+          },
         Fragment::Interpolation { expression } => {
           evaluated += &self.evaluate_expression(expression)?;
         },
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -89,7 +89,7 @@ impl<'src, D> Recipe<'src, D> {
     if self.shebang {
       let mut evaluated_lines = vec![];
       for line in &self.body {
-        evaluated_lines.push(evaluator.evaluate_line(line)?);
+        evaluated_lines.push(evaluator.evaluate_line(line, false)?);
       }
 
       if config.dry_run || self.quiet {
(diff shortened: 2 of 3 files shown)


casey:
I am doing a Janus run to try to figure out if there are recipes that are likely to be broken. What I'll be looking for are line continuations without a space before the slash, and with leading whitespace on the next line. Like this:

```
foo:
  echo bar\
    baz
```

This currently runs `echo bar  baz`, but after the change would run `echo barbaz`.

casey:
I could only find one justfile on github that seems to be affected, and [opened a PR](https://github.com/jsim2010/paper/pull/166) with a fix.

camerondavison:
Just my two cents. After this broke my justfiles (they are private so no way you would have known) I found this ticket.

I think that it would have been nice to make this work like Make in that it stripped all the spaces equal to the indent. I think that is what happens for the shebang polyglot code.

casey:
> Just my two cents. After this broke my justfiles (they are private so no way you would have known) I found this ticket.

Ahh, sorry this broke your justfiles. Did they have slashes without spaces before them? Like this:

```
foo:
  echo bar\
   baz
```

I think I assumed that having no space before line continuation slashes is less common than it actually is.

> I think that it would have been nice to make this work like Make in that it stripped all the spaces equal to the indent. I think that is what happens for the shebang polyglot code.

Do you mean that in this justfile:

```
foo:
  echo bar \
   baz
```

The first line of foo would be:

```echo bar  baz```

I.e. two spaces between `bar` and `baz`?

**Comment by camerondavison** [comment]
Thanks for reaching out. Yes. This is my example
Makefile
```
echo:
	set -x && echo test\
	this
echo2:
	set -x && echo test\
	 this
```
```
❯ make echo
set -x && echo test\
	this
+ echo testthis
testthis
```
```
❯ make echo2
set -x && echo test\
	 this
+ echo test this
test this
```
and justfile `just v0.6.1`
```
echo:
    set -x && echo test\
     this
echo2:
    #!/bin/bash
    set -x && echo test\
     this
```
```
❯ just echo
set -x && echo testthis
+ echo testthis
testthis
```
```
❯ just echo2
+ echo test this
test this
```

**Comment by casey** [comment]
Thanks for the examples. Is there a reason that you prefer not stripping leading spaces from the line after a line continuation?

My thinking was that you often want to indent the line after a line continuation, to visually distinguish it, but you don't necessarily want to include the whitespace, especially if there is a space before the slash.

In this example, if you're using two space indents, you probably don't want two spaces between `foo` and `bar`:

```
echo:
  set -x && echo foo\
    bar
```

If you like putting a space before the slash, as in this example, you can't indent the next line or else you'll get extra spaces:

```
echo:
  set -x && echo foo \
  bar
```

camerondavison:
I can for sure change my behavior. I appreciate the conversation. I think that I never liked putting the space before the slash because then i would not be able to indent it in make without adding an extra space. So I guess the change allows one to put the space before the slash without adding extra spaces. I just got used to putting the slash at the end because of that.

casey:
Thanks for the background, I like the new behavior, although it's definitely unfortunate that I introduced an incompatibility with make :/

camerondavison:
Makes sense. Is there a thought to do the same for the shebang polyglot syntax?

casey:
Shebang recipes should be safe! Shebang recipes are extracted without doing any processing aside from stripping indentation, so any escaping is done by the underlying interpreter, whatever that may be.

PR #636, opened 2020-06-09
Ignores '@' returned from interpolation evaluation

rjsberry:
Just now interprets this literally rather than propagating it to make the line run quietly.

### DIFF
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -206,6 +206,7 @@ impl<'src, D> Recipe<'src, D> {
         }
         let mut evaluated = String::new();
         let mut continued = false;
+        let quiet_command = lines.peek().map(|line| line.is_quiet()).unwrap_or(false);
         loop {
           if lines.peek().is_none() {
             break;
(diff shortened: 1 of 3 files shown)


src/recipe.rs
```diff
@@ -206,12 +206,16 @@ impl<'src, D> Recipe<'src, D> {
...
         let mut evaluated = String::new();
         let mut continued = false;
+        let mut quiet_command = false;
         loop {
           if lines.peek().is_none() {
             break;
           }
           let line = lines.next().unwrap();
           line_number += 1;
+          if evaluated.is_empty() {
```
casey: This will return true if we're on the second line, but the first line didn't evaluate to any text, for example:

```
foo:
  {{""}}\
  @echo foo 2>/dev/null
```

In practice, I don't think anyone is likely to hit this, but it would be nice to add a test for it and only mark lines as quiet if the leading `@` is on the first line.

I think this can be avoided by moving the check for a quiet command outside the loop, like so:

```
let quiet_command = lines.next().map(|line| line.is_quiet()).unwrap_or(false);
```

casey (changes requested):
Nice! Just one issue with an edge case.

src/recipe.rs
rjsberry: Nice catch.

casey:
Merged, thank you very much!

PR #639, opened 2020-06-10
Add pre-commit justfile example

### LINKED ISSUE #637: Usage example
Usage idea that nicely pairs with `pre-commit` and helps out with - semantically replenished - `just commit` & `just amend` skipping the hooks for the current raid.

Also I found it pretty convenient to separate `install` and `setup` for onboarding.

```justfile
# use with https://github.com/casey/just

# fixe auto-fixable lint issues in staged files
fix:
	pre-commit run go-returns  # fixes all Go lint issues
	pre-commit run prettier    # fixes all Markdown (& other) lint issues

# lint most common issues in - or due - to staged files
lint:
	pre-commit run go-vet-mod || true  # runs go vet
	pre-commit run go-lint    || true  # runs golint
	pre-commit run go-critic  || true  # runs gocritic

# lint all issues in - or due - to staged files
lint-all:
	pre-commit run golangci-lint-mod || true  # runs golangci-lint

# run tests in - or due - to staged files
test:
	pre-commit run go-test-mod || true  # runs go test

# commit skipping pre-commit hooks
commit m:
	git commit --no-verify -m "{{m}}"

# amend skipping pre-commit hooks
amend:
	git commit --amend --no-verify

# install/update code automation (prettier, pre-commit, goreturns, lintpack, gocritic, golangci-lint)
install:
	npm i -g prettier
	curl https://pre-commit.com/install-local.py | python3 -
	go get github.com/sqs/goreturns
	go get github.com/go-lintpack/lintpack/...
	go get github.com/go-critic/go-critic/...
	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.27.0

# setup/update pre-commit hooks (optional)
setup:
	pre-commit install --install-hooks # uninstall: `pre-commit uninstall`
```

blaggacao:
closes #637

### DIFF
--- a/examples/pre-commit.just
+++ b/examples/pre-commit.just
@@ -8,7 +8,7 @@
 # pre-commit brings about encapsulation of your
 # most common repo scripting tasks. It is perfectly
 # usable without actually setting up precommit hooks.
-# If you chose to, this justfiles include shorthands
+# If you chose to, this justfiles includes shorthands
 # for git commit and amend to keep pre-commit out of
 # the way when in flow on a feature branch.
 
--- a/examples/pre-commit/00.justfile
+++ b/examples/pre-commit/00.justfile
@@ -1,0 +1,59 @@
+# use with https://github.com/casey/just
+
+# Example combining just + pre-commit
+# pre-commit: https://pre-commit.com/
+# > A framework for managing and maintaining
+# > multi-language pre-commit hooks.
+
+# pre-commit brings about encapsulation of your
+# most common repo scripting tasks. It is perfectly
+# usable without actually setting up precommit hooks.
+# If you chose to, this justfiles include shorthands
+# for git commit and amend to keep pre-commit out of
+# the way when in flow on a feature branch.
+
+# uses: https://github.com/tekwizely/pre-commit-golang 
+# uses: https://github.com/prettier/prettier (pre-commit hook)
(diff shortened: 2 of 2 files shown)


casey:
Looks good! Is the `00.justfile` name intentional? Can it just be `Justfile`?

blaggacao:
Jep it's intentional, if somebody comes along with another example, one day...
Maybe there might also be a more common topic than `pre-commit` where it could get denser, so I figured we should provide a non-semantic mechanism (ordinal) to account for multiple examples.

Do you think that's over-engineered?

casey:
I think if there's another pre-commit example, it could go in `pre-commit-foo/Justfile`.

I'm actually really not sure if they should be in subdirectories at all. The reason I thought subdirectories would be good is so the actual file could be called `Justfile`, but it might be better to just have them be in `examples`, with a `.just` suffix, so `examples/pre-commit.just`, `examples/automation.just`, etc.

I think it's probably good to make the hierarchy more flat, and then add depth later as needed.

blaggacao:
Agree!

casey:
Awesome, looks great! Thanks for this, I think example justfiles will be a great addition to the docs. Soon, I'll add a link in the readme, and also add an integration test that makes sure that example justfiles always parse correctly.

PR #640, opened 2020-06-10
 Add keybase example 

blaggacao:
Just a nice idea since keybase does authenticates the user by means of the device he runs it.


casey:
Merged! Thank you!

PR #641, opened 2020-06-10
Link example justfiles in readme

casey:


blaggacao:
LGTM, nice choice to put it prominently!

casey:
Definitely! I think that when learning a new language, just looking at some examples is very helpful.

PR #646, opened 2020-06-13
Clarify variadic parameters with default values in README

rjsberry:
Both `+` and `*` variadics accept defaults! My mistake, I should have caught this in #645 cba69db.


casey:
Nice catch, thanks!

Issue #573, opened 2020-01-15
Add recipes to generated completion scripts


casey:
In #572, I'll be landing a `--completions` flag to make `just` output completion scripts for various shells. However, these completion scripts should also allow completing recipe names (like `build` or `test`). I'll need to add this functionality separately.

See #223 for completion scripts that others have written.

Add recipe completions to:
- [x] Bash
- [x] Zsh
- [x] Fish
- [x] PowerShell
- [ ] Elvish

gotcha:
@casey If the `just.bash` file (see https://github.com/casey/just/pull/572/files#diff-1dcc75cc227d1a56d11d839f20f6a50c) is generated with `clap`, how do you plan to maintain it 
if I fiddle to merge the recipe completion I have in https://github.com/gotcha/just-bash-completion ?

casey:
@gotcha Good question!

I was thinking that I would first write the completion script to a string while generating it in `config.rs`, and then modifying it before printing it. (Either by appending more stuff to the end, or doing a search and replace if it's more involved than that.)

If you can show me what should be added to the completion script (either in a gist, or a PR), then I can make the changes in rust to add it when it's generated too.

gergelyk:
<ARGUMENTS>In case of bash, I do this dirty trick to replace `<ARGUMENTS>...` with (fine tuned) output of `just --list`.

```sh
source <(OPTS=$(just --list | tail -n +2 | sed 's/    / /g' | tr -d '\n'); just --completions bash | sed "s/  <ARGUMENTS>... /$OPTS/g")
```

Do you think it would make sense to include similar solution in default script generated by `just --completions bash`? Probably some dedicated variant of `just --list` would be applicable here.

casey:
That's an awesome trick! I'd be worried about adding it into the default script, just because it adds dependencies on `tail`, `sed`, and `tr`. But if you can modify the bash completion script to include recipe names without using those additional commands, I can modify the scripts when they're generated, here: https://github.com/Insomniak47/just/blob/master/src/subcommand.rs

gergelyk:
I agree. `tail`, `sed`, `tr` is probably what we wand to avoid. As an alternative I would suggest extending functionality of `--list`. E.g. `--list-inline` would give the same output as `--list` but in a single line, without header line and space-separated. This way completion script could be modified so that instead of `<ARGUMENTS>...` we have `$(just --list-inline)`.

casey:
[...]

PR #651, opened 2020-06-23
powershell recipe autocomplete implementation

Insomniak47:
Adds powershell recipe autocomplete as mentioned in: #573

Because it's executing in a scriptblock I have to include the function doing it as a local function. I've also set it up so that if you specify a justfile with --justfile it will use that for the completions. 

```
ex. 
justfile
	recipes:
		- A
		- B
		- C

justfileB:
	recipes:
		- X
		- Y
		- Z
```

`just <tab>` would get you the results of just --list in order.
`just --justfile ".\justfileB" <tab>`
will complete X, Y, Z. You can test it out from outside of the just repo. I've exercised this a bunch locally but not with any exceptional cases. As well the default vscode powershell extension formatter took over and formatted the file. I'm definitely fine with reverting it but you'll likely run into it again.

### DIFF
--- a/completions/just.powershell
+++ b/completions/just.powershell
@@ -73,9 +73,8 @@ Register-ArgumentCompleter -Native -CommandName 'just' -ScriptBlock {
             $justArgs += @("--justfile", $justFileLocation)
         }
 
-        $recipesRaw = $(just @justArgs) | Select-Object -Skip 1
-        $recipes = $recipesRaw | ForEach-Object { (($_.Trim()) -split ' ')[0] }
-        return  $recipes | ForEach-Object { [CompletionResult]::new($_) }
+        $recipes = $(just @justArgs) -split ' '
+        return $recipes | ForEach-Object { [CompletionResult]::new($_) }
     }
 
     $elementValues = $commandElements | Select-Object -ExpandProperty Value
--- a/completions/just.powershell
+++ b/completions/just.powershell
@@ -67,7 +67,7 @@ Register-ArgumentCompleter -Native -CommandName 'just' -ScriptBlock {
             $justFileLocation = $commandElements[$justFileIndex + 1]
         }
 
-        $justArgs = @("--list")
+        $justArgs = @("--summary")
 
         if (Test-Path $justFileLocation) {
             $justArgs += @("--justfile", $justFileLocation)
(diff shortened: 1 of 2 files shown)


casey:
Thanks for this! Completions are generated in a somewhat weird way. First, completions are generated by Clap, the command-line argument parser that Just uses. Then, they're modified in code to add recipe name completions.

There are tests which check that the completions on disk match the completions that are output by the code, so that they don't get out of sync.

In order to make the tests pass, the modifications to the PowerShell completions should be done [here](https://github.com/casey/just/blob/863fb53885b602d1346071d51dd929bef7fc76ff/src/subcommand.rs#L113). If you're not comfortable with Rust, I can move the changes for you.

Either way, the whitespace/formatting changes should be reverted, to match what is output by Clap and keep the diff minimal.

Insomniak47:
I'll take a look at getting that into the generator

casey:
Definitely let me know if you need any help!

casey:
@Insomniak47 I just pushed a diff that moves the changes into `src/subcommand.rs`, which should make the tests pass. However, I noticed that the code used `--list`, instead of `--summary`. `--summary` prints a space-separated list of recipe names, and doesn't print comments or arguments, so it's more suitable to be processed by scripts.

Since I changed it to `--summary`, the recipe splitting logic is probably wrong now. Would you mind taking a look? The modification should be made to the replacement strings in `POWERSHELL_COMPLETION_REPLACEMENTS` in `src/subcommand.rs`.

Also, I don't have powershell installed locally, so if you could test that it works, that would be much appreciated.

Insomniak47:
Hey, I had completely forgotten about this. I've taken a look and made a fix to work with your changes. It looks like it simplifies it a little bit. I've tested it locally. 

```pwsh
cargo run -- --completions powershell > .\test.ps1
. .\test.ps1
just b<tab> --> just build
just s<tab> --> just sloc<tab> --> just spam
cd .. 
just <tab> --> just --choose<tab> just --chooser --> etc etc
just b<tab> --> just b (no change)
just --justfile .\just\justfile b<tab> --> just [..] build
# Same as if we're pointing at that justfile!
just --justfile .\just\.editorconfig b<tab> --> (no change, as expected)
```

It seems like it's working relatively well. There might be a few edge cases but this seems to be reasonably well. I've only tested it on windows in pwsh/Powershell and not on pwsh on linux

casey:
Excellent! Thank you for the PR, this is a great quality of life improvement.

PR #663, opened 2020-07-31
Strip Linux binary

xrelkd:
Strip Linux binary before packaging, this will remove `debug information` and make the binary smaller.

### DIFF
--- a/bin/package
+++ b/bin/package
@@ -16,7 +16,12 @@ test -f Cargo.lock || cargo generate-lockfile
 echo "Building $bin..."
 
 case $os in
-  ubuntu-latest | macos-latest)
+  ubuntu-latest)
+    cargo rustc --bin $bin --target $target --release
+    executable=target/$target/release/$bin
+    strip $executable
+    ;;
+  macos-latest)
     cargo rustc --bin $bin --target $target --release
     executable=target/$target/release/$bin
     ;;


casey:
Thanks for the PR! I'm not sure why the Windows tests are failing, but I think they're safe to ignore.

How much does this reduce the binary size? Also, are there any drawbacks to doing this? For example, is it possible to get backtraces using `RUST_BACKTRACE=1` after stripping?

xrelkd:
The size of binary will reduce from `4.1 MiB` to `2.3 MiB`, after stripping. (test with `just-v0.7.1-x86_64-unknown-linux-musl`)
The stack backtrace will be useless after stripping the binary, because the debug information is removed.

For a stripped binary, its stack backtrace will be:
```
thread 'main' panicked at 'test 123', src/main.rs:4:5
stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: __libc_start_main
  12: <unknown>
```

Should I close this PR?

casey:
Thanks for the info!

I could really be persuaded either way.

4mib to 2mib is a great savings, although disk and bandwidth are very cheap these days.

I'm not sure stack traces have ever actually been used to debug release binaries. Just doesn't crash a whole lot, and when it does it tends to be reproducible. So users can provide a justfile and command that trigger the crash, which can then be debugged on a debug build.

I think I'm inclined to leave this open for now in the hope of getting some feedback. I'd also be curious what the best practices of package managers are.

xrelkd:
Thanks for the review!!
I will keep this PR open.

casey:
I think I'm going to close this. I haven't gotten any feedback on it either way, so I'm inclined to not strip the binary, since in the event of a bug that causes a crash, I'd like to keep backtraces, in case they're helpful to end users. I do appreciate the PR though!

PR #685, opened 2020-10-01
Use recipes for bash shell completions.

vikesh-raj:
Use just --list command to get the available recipes and use that to complete the command.


casey:
Thank you for this! `--list` includes doc comments and arguments, in addition to the recipe's name. Should `--summary` be used instead?

vikesh-raj:
I don't have rust environment or zsh for running tests.
I like this program a lot. I sometimes forget the recipe name. This tab complete feature helps in listing the recipes and completing them.

casey:
Glad you like it, and no worries!

The reason the tests are failing is a bit convoluted. The completion scripts are generated automatically using clap, an command line argument parser. But, the generated scripts don't include the ability to complete recipes, since clap only knows about flags and options. So, I wrote a rust program that makes modifications to the completion scripts to include recipe completion.

The test that's failing checks that the generated completion script is the same as what's checked in, which is why it's failing.

I'm happy to merge the changes in to the script, I should be able to do it tomorrow.

casey:
I modified the script, so the tests should pass now. (You can ignore the error, it's spurious.) However, there's a problem, with the modifications, the bash completion script won't complete flags and options, so for example, the following don't work:

```
$ just <TAB>
$ just -<TAB>
```

vikesh-raj:
@casey , I have now modified the the subcommand.rs file to replace a bigger portion.
just -TAB , now completes the options
just TAB, completes the recipes.
This is the same behaviour for make command.

casey:
Awesome, merged! Thank you very much for the pull request!

PR #686, opened 2020-10-02
Move separate quiet config value to verbosity

### LINKED ISSUE #520: Make `--quiet` a level of verbosity
Currently, a `Config` has a `quiet` field *and* a `verbosity` field. This is somewhat silly, they should be a single `verbosity` field, with quiet being represented as `Verbosity::Quiet`.

Celeo:
Fixes #520

Moves the separate `quiet` variable in the `Config` struct into the existing `verbosity` field; the `Verbosity` enum now has a `Quiet` value. When running, the presence of the quiet flag will set the verbosity to quiet, overriding any number of verbosity flags also in the CLI args.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -20,7 +20,6 @@ pub(crate) struct Config {
   pub(crate) highlight:            bool,
   pub(crate) invocation_directory: PathBuf,
   pub(crate) load_dotenv:          bool,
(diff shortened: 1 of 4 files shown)


casey:
Awesome, looks good. Thank you!

Issue #660, opened 2020-07-18
How can I ignore an error in a command?


mortoray:
There are some commands for which I want to ignore error returns and continue processing. Is there a way to mark a line as ignoring the return value?

This came up with killall which I used to end a process that may or may not be there. The current alternative I have is `killall some_proce || true`, though that seems like a bash thing, not a justfile thing.

casey:
I think that currently, `foo || true` is the best way to do this. If you have a lot of commands you'd like to ignore errors for, you can switch to using a shebang recipe, which behaves like a normal shell script with respect to error codes, i.e. ignores them:

```
foo:
  #!/usr/bin/env sh
  cat bar
```

Make actually has a feature for this. If you put `-` in front of a command, it will ignore errors:

```
$ cat Makefile
foo:
  exit 1
$ make
exit 1
make: *** [foo] Error 1
```

```
$ cat Makefile
foo:
  -exit 1
$ make
exit 1
make: [foo] Error 1 (ignored)
```

It's a bit magical, but I wouldn't be opposed to adding this feature to Just. It would also make a great first issue for anyone interested.

iwillspeak:
I've taken a stab at implemnting this feature in #687 .

casey:
This has been done! Thank you very much, @iwillspeak. I added some documentation in the readme, and cut v0.8.0 that includes the feature. Technically, this is a breaking change, which is why the minor version was bumped to 8, but I don't think it will be a problem in practice, since someone would have to have a recipe line starting with `-`.

PR #687, opened 2020-10-03
Allow Suppressing of Failing Commands with `-` Prefix

iwillspeak:
Similar to `make` if a command is prefixed with `-` then the exit
status is ignored.

Fix for  #660


iwillspeak:
I’m not sure what’s going on here with the snafu-derive error. It doesn’t happen locally on macOS. Could it be a spurious build failure ?

casey:
Nice, merged! The build failure is sporadic failure, but I found a workaround, so hopefully it wont' be a problem going forward.

PR #692, opened 2020-10-04
Fix doc Ignoring Errors with `-` in README

kenden:
Fix doc Ignoring Errors with `-` in README.
The README example forgets to use `-`


casey:
Good catch, thank you very much!

PR #697, opened 2020-10-09
fix(zsh): Fix variable completion

### LINKED ISSUE #695: zsh autocomplete issue tabbing after passing --set VAR value
There is an issue with the 
zsh autocompletion when pressing tab after 
`just --set VAR value `
It displays a block of error.

### Load autocompletion in your terminal
`$ source <(just --completions=zsh)`

### Create a justfile 

```make
GREET_WORD := 'Hello'
greet name:
  @echo {{GREET_WORD}} {{name}}!
```
### Use it

```make
$ just greet Robert
Hello Robert!
```
### Use it with autocompletion

If we want to set the variable GREET_WORD from the command line using `--set`, we would enter:  
`$ just --set GREET_WORD ` then press Tab  
---> It autocompletes to:  
`$ just --set GREET_WORD greet`  
here, it should not autocomplete with a target, or anything, just wait for the user to enter a value.

But, if we type a correct content for the variable:  
`just --set GREET_WORD Hola ` then press Tab
----> we get:
```bash
$ just --set GREET_WORD Hola …error: The argument '--set <VARIABLE> <VALUE>' requires a value but none was supplied

USAGE:
    just --color <COLOR> --set <VARIABLE> <VALUE> --shell <SHELL> --shell-arg <SHELL-ARG>... <--choose|--completions <SHELL>|--dump|--edit|--init|--evaluate|--list|--show <RECIPE>|--summary|--variables>

For more information try --help
error: The argument '--set <VARIABLE> <VALUE>' requires a value but none was supplied

USAGE:
    just --color <COLOR> --set <VARIABLE> <VALUE> --shell <SHELL> --shell-arg <SHELL-ARG>... <--choose|--completions <SHELL>|--dump|--edit|--init|--evaluate|--list|--show <RECIPE>|--summary|--variables>

For more information try --help
error: The argument '--set <VARIABLE> <VALUE>' requires a value but none was supplied

USAGE:
    just --color <COLOR> --set <VARIABLE> <VALUE> --shell <SHELL> --shell-arg <SHELL-ARG>... <--choose|--completions <SHELL>|--dump|--edit|--init|--evaluate|--list|--show <RECIPE>|--summary|--variables>

For more information try --help
```
-> Here, it should have autocompleted with a target.

You can see the steps above in this recording:
just-autocomplete-zsh-issue

## Note

It would be nice to be able to set parameters with --set using space OR `=`, e.g.  
`just --set GREET_WORD Hola greet Robert`  
or  
`just --set GREET_WORD=Hola greet Robert`  
or  
`just --set=GREET_WORD=Hola greet Robert`  
or (with quotes, even):
`just --set="GREET_WORD=Hola" greet Robert`
(just mentioning this because using a = sign would not trigger this issue)

## Versions
`just v0.8.0`
`zsh 5.7.1 (x86_64-apple-darwin19.0)`

heyrict:
This PR contains the following fixes:

- Allow completion of variables in the command context
- Allow completion of variables after `--set` flag

Closes #695

### DIFF
--- a/completions/just.zsh
+++ b/completions/just.zsh
@@ -15,6 +15,7 @@ _just() {
 
     local context curcontext="$curcontext" state line
     local common=(
+'--chooser=[Override binary invoked by `--choose`]' \
 '--color=[Print colorful output]: :(auto always never)' \
 '-f+[Use <JUSTFILE> as justfile.]' \
 '--justfile=[Use <JUSTFILE> as justfile.]' \
@@ -28,12 +29,16 @@ _just() {
 '--show=[Show information about <RECIPE>]: :_just_commands' \
 '(-q --quiet)--dry-run[Print what just would do without doing it]' \
 '--highlight[Highlight echoed recipe lines in bold]' \
+'--no-dotenv[Don'\''t load `.env` file]' \
 '--no-highlight[Don'\''t highlight echoed recipe lines in bold]' \
 '(--dry-run)-q[Suppress all output]' \
 '(--dry-run)--quiet[Suppress all output]' \
 '--clear-shell-args[Clear shell arguments]' \
+'-u[Return list and summary entries in source order]' \
+'--unsorted[Return list and summary entries in source order]' \
 '*-v[Use verbose output]' \
 '*--verbose[Use verbose output]' \
+'--choose[Select one or more recipes to run using a binary. If `--chooser` is not passed the chooser defaults to the value of $JUST_CHOOSER, falling back to `fzf`]' \
 '--dump[Print entire justfile]' \
 '-e[Edit justfile with editor given by $VISUAL or $EDITOR, falling back to `vim`]' \
 '--edit[Edit justfile with editor given by $VISUAL or $EDITOR, falling back to `vim`]' \
--- a/completions/just.zsh
+++ b/completions/just.zsh
@@ -63,14 +63,11 @@ _just() {
         args)
             curcontext="${curcontext%:*}-${words[2]}:"
 
-            local lastarg=${words[${#words}]}
-
             local cmds; cmds=(
                 ${(s: :)$(_call_program commands just --summary)}
             )
 
             # Find first recipe name
-            integer skip=0
             for ((i = 2; i < $#words; i++ )) do
                 if [[ ${cmds[(I)${words[i]}]} -gt 0 ]]; then
                     recipe=${words[i]}
@@ -78,7 +75,7 @@ _just() {
                 fi
             done
 
-            if [[ ${lastarg} = */* ]]; then
+            if [[ $lastarg = */* ]]; then
                 # Arguments contain slash would be recognised as a file
                 _arguments -s -S $common '*:: :_files'
             elif [[ $lastarg = *=* ]]; then
--- a/completions/just.zsh
+++ b/completions/just.zsh
@@ -63,6 +63,8 @@ _just() {
         args)
             curcontext="${curcontext%:*}-${words[2]}:"
 
+            local lastarg=${words[${#words}]}
+
             local cmds; cmds=(
                 ${(s: :)$(_call_program commands just --summary)}
             )
--- a/completions/just.zsh
+++ b/completions/just.zsh
@@ -15,7 +15,6 @@ _just() {
 
     local context curcontext="$curcontext" state line
     local common=(
-'--chooser=[Override binary invoked by `--choose`]' \
 '--color=[Print colorful output]: :(auto always never)' \
 '-f+[Use <JUSTFILE> as justfile.]' \
 '--justfile=[Use <JUSTFILE> as justfile.]' \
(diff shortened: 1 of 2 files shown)


casey:
Awesome! I found a couple issues:

This shows variables, but the "variables" header is at the top of the output, not above the variables:

```
just <TAB> 
```

Screen Shot 2020-10-09 at 16 30 11

This completes VARIABLE_NAME:

```
just --set V<TAB> 
```

But it adds a `=` after `VARIABLE_NAME`, but `=` is only used for the `just VARIABLE_NAME=value` shorthand, not the `--set` form, which uses `just --set VARIABLE_NAME value`.

This produces an error:

```
just --set VALUE foo <TAB>
```

Screen Shot 2020-10-09 at 16 33 58

It looks like perhaps 'foo' isn't being passed to the underlying invocation?

casey:
The tests were failing because a new version of clippy introduced a new lint, and because the generated completion scripts were behind the code in `subcommands.rs`. I pushed diffs that fix both errors.

heyrict:
> But it adds a = after VARIABLE_NAME, but = is only used for the just VARIABLE_NAME=value shorthand, not the --set form, which uses just --set VARIABLE_NAME value

By convention, one argument can be followed by at most one flag asaik. It's out of my knowledge to pass two argument to one flag. I can remove the completion part of `--set` option if you think it is not suitable here.

> This shows variables, but the "variables" header is at the top of the output, not above the variables

Can I have a look at your zshrc?

>  I pushed diffs that fix both errors.

Thanks :)

casey:
> By convention, one argument can be followed by at most one flag asaik. It's out of my knowledge to pass two argument to one flag. I can remove the completion part of `--set` option if you think it is not suitable here.

Ah, okay. I think in that case it's probably better not to try to complete `--set` in that case.

> Can I have a look at your zshrc?

Yup, it's here: https://github.com/casey/local/blob/master/etc/zshrc

It refers to other files, which are also in that repo, and I think they're all in this folder: https://github.com/casey/local/tree/master/etc/zsh

heyrict:
Thanks :)

> This shows variables, but the "variables" header is at the top of the output, not above the variables

This issue doesn't exist on my side. I'll have a look at your code once I have time.

casey:
> This issue doesn't exist on my side. I'll have a look at your code once I have time.

Hmm, interesting. I don't want to hold up this diff because of a problem that might only be on my end.

I pushed a diff that makes the test pass, but when testing locally I ran into the following issue:

```
just --set V<TAB>
```

Completes to:

```
just --set VARIABLE_NAME=
```

The `=` can only be used with sets of the form `just VARIABLE_NAME=value`, and not `--set VARIABLE_NAME value`, which doesn't take a `=`. I tried having a look at the code to see if I could fix this myself, but I have absolutely no idea how zsh completion scripts work 😅

casey:
I noticed that this PR introduces a `numargs` variable, that maps flags to the number of arguments they take. Will this have to be kept in sync with new flags, and what will happen if a new flag is added that takes arguments, but this map isn't updated?

completions/just.zsh
```diff
@@ -82,20 +120,43 @@ _just() {
...
         ${${${(M)"${(f)$(_call_program commands just --list)}":#    *}/ ##/}/ ##/:Args: }
     )
 
-    _describe -t commands 'just commands' commands "$@"
+    if compset -P '*='; then
+        case "${${words[-1]%=*}#*=}" in
+            *) _message 'value' && ret=0 ;;
+        esac
+    else
+        _describe -t variables 'variables' variables -qS "=" && ret=0
```
heyrict, line 117:
```suggestion
        _describe -t variables 'variables' variables && ret=0
```

This should fix the equal sign. I'll push a commit once I'm at my PC.

heyrict:
> I noticed that this PR introduces a `numargs` variable, that maps flags to the number of arguments they take. Will this have to be kept in sync with new flags, and what will happen if a new flag is added that takes arguments, but this map isn't updated?

I'm afraid not. Probably checking the first argument that matches the name of an existing recipe should work if you'd like to avoid the args checking hack. I'll try that once I have time.

casey:
> I'm afraid not. Probably checking the first argument that matches the name of an existing recipe should work if you'd like to avoid the args checking hack. I'll try that once I have time.

That would be great. My expectation is that I'll forget to update this script when I add new command line options. If the completion script has fewer features, like not completing certain things that it could, I think that worth it if it doesn't get out of date.

kenden:
Hi @casey, about: 
> The `=` can only be used with sets of the form `just VARIABLE_NAME=value`, and not `--set VARIABLE_NAME value`, which doesn't take a `=`.

Do you think `just` could be changed to allow a space _and_ a `=` when using `--set`?
i.e. to have both '`--set VARIABLE_NAME value`' and '`--set VARIABLE_NAME=value`' be valid syntax.  
It seems pretty common for other commands to allow both. 
And doing this should allow completing the variable name too?

casey:
This is definitely a good idea, but I don't think it can be implemented without undesirable side-effects with the argument parser, clap, that Just uses.

When I set the "value delimiter" for `--set` to `=`, this does work as you'd expect `--set foo=bar`. However, it will error if you do `--set foo=bar=baz`, i.e. if you use `=` in the value that you're trying to set. I think this would be undesirable, since it would mean that setting variables would fail if the value happened to contain a `=`.

(The reason it fails is because it interprets `foo=bar=baz` as three arguments to `--set`, which is like if you did `--set foo bar --set baz`, the second set only has the key and no value.)

casey:
I think this has an issue, if any flags or variables are given, recipe name tab completion doesn't work, and it instead shows the usage string for the first recipe:

```
$ just JUST_LOG=bar <TAB>
# add git log messages to changelog
changes:
    git log --pretty=format: >> CHANGELOG.md
```

heyrict:
It seems that I've accidentally removed one line from the source code. It should be fine now.

casey:
It looks like there's some state that's persisting between calls to the completion script.

```
$ just JUST_LOG=foo <TAB> # shows recipes and variables
$ just JUST_LOG=foo chang<TAB> # completes to `changes`
$ just JUST_LOG=foo <TAB> # now shows the text of the `changes recipe`
```

heyrict:
> It looks like there's some state that's persisting between calls to the completion script.

Thanks for the effort on debugging. Fixed now.

casey:
Awesome, looks good!

PR #729, opened 2020-11-26
Add link to GitHub Action

rossmacarthur:
Just wanted to share the GitHub Action I wrote to install `just`. I use `just` regularly as a command runner in my projects and often want to run it in CI. See [extractions/setup-just](https://github.com/extractions/setup-just).

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1181,6 +1181,18 @@ Tools that pair nicely with `just` include:
 
 - https://github.com/mattgreen/watchexec[`watchexec`] — a simple tool that watches a path and runs a command whenever it detects modifications.
 
+=== GitHub Actions
+
+link:https://github.com/extractions/setup-just[extractions/setup-just] can be used to install `just` in a GitHub Actions workflow.
+
+Example usage:
+
+```yaml
+- uses: extractions/setup-just@v1
+  with:
+    just-version: 0.8  # optional semver specification, otherwise latest
+```
+
 === Shell Alias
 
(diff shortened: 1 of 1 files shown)


casey:
This is awesome!

For the readme, what do you think about:

````
=== GitHub Actions

link:https://github.com/extractions/setup-just[extractions/setup-just] can be used to install `just` in a GitHub Actions workflow.

Example usage:

```yaml
- uses: extractions/setup-just@v1
  with:
    just-version: 0.8  # optional semver specification, otherwise latest
```

````

casey:
Is "semver specification" the word for a semver requirement specifier, i.e. "^0.5.0"?

I was trying to figure out what people call it  most frequently, but nothing stood out. It seems like, in practice, people don't distinguish between a semantic version, e.g. `0.5.0`, and a npm-style semver specifier, i.e `^0.5.0`.

In any case, `semver specification` is fine.

rossmacarthur:
Happy with your requested changes to the text, I have updated the branch.

Yes, it is a semver requirement specifier. I'm using this package in the GitHub Action: https://www.npmjs.com/package/semver. This package resolves plain versions slightly differently to Cargo though where `0.5.0` is equivalent to `=0.5.0` not `^0.5.0`.

casey:
Awesome, merged.

You might consider documenting in the action's readme that it's an npm-style semver spec. Just users are pretty varied, but I think a good chunk of them are Rust users, and they might be confused by the difference.

Thanks so much for writing and documenting this, I think it's a really nice addition.

PR #736, opened 2020-12-13
Readme: Add Alpine Linux package

jirutka:
Note: Alpine 3.13 has not been released yet, it should be till the end of this year.


casey:
Awesome, thank you very much for creating the package!

Should this be merged after 3.13 releases, so that Alpine users aren't confused that the package isn't available yet?

jirutka:
> Awesome, thank you very much for creating the package!

You’re welcome!

> Should this be merged after 3.13 releases, so that Alpine users aren't confused that the package isn't available yet?

Yes, it will be better.

casey:
Sounds good, I'll check around Jan 1st, which is when I think it's due out. But if I forget, please feel free to ping the PR.

jirutka:
Alpine 3.13.0 is released! :)

casey:
Merged, thank you so much!

casey:
(I removed the version, just because we don't have it for other distros, and I think it might look nicer as its own column if we did add it. Also, I changed the apk link to the apk docs, since that's probably what's most useful for someone looking to install just on Alpine.)

PR #742, opened 2021-01-07
github uses actions/cache@v2

zyctree:
https://github.com/actions/cache/blob/main/examples.md#rust---cargo

but the cache in windows still cannot work,  seems https://github.com/actions/toolkit/pull/670 could fix it, need to wait github to bump a new version to include this fix

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -45,7 +45,7 @@ jobs:
         echo /usr/local/opt/gnu-tar/libexec/gnubin > $GITHUB_PATH
 
     - name: Cache Cargo
-      uses: actions/cache@v1
+      uses: actions/cache@v2
       with:
         path: |
           ~/.cargo/registry


casey:
Thanks for this PR! Does the actions/cache@v1 work in Windows, or is it newly broken in v2?

zyctree:
actions/cache@v1  is broken in Windows

from the output of the last action job, https://github.com/casey/just/runs/1546397193?check_suite_focus=true, 

```
Post job cleanup.
"C:\Program Files\Git\usr\bin\tar.exe" -cz -f D:\a\_temp\d43ac80a-b321-468c-9777-c01c94d9a40b\cache.tgz -C C:\Users\runneradmin\.cargo\registry
~\.cargo\git
target .
/usr/bin/tar: ~.cargogit: Cannot stat: No such file or directory
/usr/bin/tar: target: Cannot stat: No such file or directory
tar (child): Cannot connect to D: resolve failed
/usr/bin/tar: D\:\\a\\_temp\\d43ac80a-b321-468c-9777-c01c94d9a40b\\cache.tgz: Cannot write: Broken pipe
/usr/bin/tar: Child returned status 128
/usr/bin/tar: Error is not recoverable: exiting now
```

it uses actions/cache@v1 , and resolve ~\.cargo\git to ~.cargogit, looks like it cannot handle `\` in windows. i guess it is using linux (wsl or mingw) version of tar

casey:
Gotcha, thanks for the explanation. I wasn't even aware that caching was broken on Windows!

PR #759, opened 2021-03-12
Add note about Emacs package just-mode

leon-barrett:
I added an Emacs package for editing justfiles, `just-mode`, and I thought it might be worth including in the README file.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1285,9 +1285,13 @@ augroup filetypedetect
 augroup END
 ```
 
-==== Vim and Emacs
+==== Emacs
 
-Include the following in a `justfile` to enable syntax highlighting in vim and emacs:
+There is a MELPA package, https://melpa.org/#/just-mode[just-mode], for automatic Emacs syntax highlighting and automatic indentation in justfiles.
+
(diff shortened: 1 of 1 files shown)


casey:
That's so cool! Thanks so much for writing it!

[This](https://github.com/leon-barrett/just-mode.el/blob/45c248fe72d4a15c5a9f26bc0b27adb874265f53/just-mode.el#L111) is very cool! So if you edit a recipe in a justfile that uses differently configured tabs, it will insert the pre-existing indent?

leon-barrett:
Thanks. It's not quite as cool as you make it sound--it's more like, while you're in a rule, it'll keep matching the indent of the previous line. When you start a new rule, it'll just use the default 4 spaces.

I believe that's helpful for writing things like Python scripts where you want varying levels of indentation within a rule. It's not super graceful, but it does something at least. (That's why there's a TODO to use nested modes for better scripting, but it's still a TODO because honestly just getting the syntax highlighting nice was hard enough for now.)

casey:
Still very nice. "It does something at least" is often the best that one can hope for!

Issue #743, opened 2021-01-07
Strange quoting behavior when using cmd.exe


casey:
This was noticed by @SuperCuber in #741.

Quoting commands, which works when using `cmd.exe` interactively, doesn't work in a Justfile.

This works:

```
set shell := ["cmd.exe", "/c"]

run:
    cmd.exe
```

But this doesn't:

```
set shell := ["cmd.exe", "/c"]

run:
    "cmd.exe"
```

The `"`s are passed to `cmd.exe`, and interpreted as part of the command name. Aside from just being weird, this makes it impossible to run commands with spaces.

casey:
I have a feeling that this might be an inherent limitation of cmd.exe.

Let's use this Justfile as an example:

```make
foo:
  echo "bar"
```

When `just` executes a recipe line, it passes the whole line as a single argument to the shell binary. When running the `foo` recipe with `sh`, the default shell, this is like running the following command:

```
sh -c 'echo "bar"'
```

`sh` receives two arguments `-c` and `echo "bar"`. It interprets `echo "bar"` as a command, and performs all the usual variable interpolation and quote interpretation. This strips the double quotes around `"bar"`, so the argument to echo is just `bar`.

However, when `cmd.exe` processes a command after `/c`, it looks like it doesn't process quotes at all. If you have this Justfile:

```
set shell := ["cmd.exe", "/c"]

foo:
    echo "bar"
```

Which translates into `cmd.exe` being passed the argument `echo "bar"`, it prints out `"bar"`, so it's not stripping the quotes at all.

So I'm not sure how to make this work. I tried different combinations of quotes, spaces, carets (which is an escape character in `cmd.exe`) but couldn't get it to work.

Is using PowerShell an option for you? I think it has much saner quoting behavior than `cmd.exe`.

Also, if someone knows how to get this to work, definitely chime in.

I'm thinking that I should perhaps put the powershell instructions first in the readme, so that people are more likely to try it than `cmd.exe`, and maybe add a warning next to the `cmd.exe` instructions that it's weird and probably not what you want if you can use powershell.

SuperCuber:
from `cmd /?`:
```
If /C or /K is specified, then the remainder of the command line after
the switch is processed as a command line, where the following logic is
used to process quote (") characters:

    1.  If all of the following conditions are met, then quote characters
        on the command line are preserved:

        - no /S switch
        - exactly two quote characters
        - no special characters between the two quote characters,
          where special is one of: &<>()@^|
[...]

PR #768, opened 2021-03-25
De-emphasize cmd.exe in readme

casey:
cmd.exe is very fiddly, and Powershell is widely available, so recommend
Powershell over cmd.exe in the readme

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -88,7 +88,18 @@ On Windows, `just` works with the `sh` provided by https://git-scm.com[Git for W
 
 If you'd rather not install `sh`, you can use the `shell` setting to use the shell of your choice.
 
-Like `cmd.exe`:
+Like Powershell:
+
+```make
+
+# use Powershell instead of sh:
+set shell := ["powershell.exe", "-c"]
+
+hello:
+  Write-Host "Hello, world!"
+```
+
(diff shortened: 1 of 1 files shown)


runeimp:
@casey why do you consider CMD fiddly? I prefer it over PowerShell in the vast majority of cases when I need to use Windows. Then again my computer experience started with DOS 3.2 so I'm quite used to it. I consider PS needlessly verbose in almost all cases. It's POSIX'ish using `ls` instead of `dir` but PS's version of `ls` is just horrible compared to `ls` on any version of BSD, Linux, or System V inspired system I've used. Not looking for you to defend your decision. I'm really curious on your thoughts. I'm sure most people that started with Windows that had PS available prefer it over CMD because it's so fancy. Also "Get off my lawn!"  ;-)

casey:
It's definitely a fair question!

I don't remember the details, but I remember that at one point I had extreme difficulty with quoting when using cmd.exe. This also came up in issue #743. I believe that in that issue, there wasn't really any way to get the desired behavior, due to differences in how cmd.exe quoting works when invoking cmd.exe from the terminal vs calling it directly from another process.

But, I don't use windows a ton, so I don't actually have that much first-hand experience.

PR #775, opened 2021-03-27
Add just_executable() function

### LINKED ISSUE #774: Function to get just' executable path (for when just is not in $PATH)
Hello, I have a simple justfile like:
```make
default:
  @just --list

test:
  echo "it works!"
```

I'm using the [Nix package manager](https://nixos.org/) which allows me to run software without really installing them, and thus they are not in PATH.
I'm trying to run `just` like this as a bootstrapping step: `nix run nixpkgs/nixpkgs-unstable#just` to run the first (default) recipe, but it fails rightfully with `sh: just: command not found` when trying to run `@just --list` in the `default` recipe.

I would like a function to get the full path to the `just` binary that is currently running, to be able to re-run it from a recipe.
The usage would be something like this:
```make
default:
  @{{just_executable()}} --list
```
And this would expand at runtime to something like (here using Nix):
```make
default:
  @/nix/store/ajq3r70f4pz5sqh5p83bbhzrkdj1cdrz-just-0.8.4/bin/just --list
```

bew:
Closes #774

~I did not add tests because I couldn't find tests for the other functions, but if you need them I can add those.~
_EDIT: nevermind, I've found them, will try to make a test_

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -585,6 +585,20 @@ $ just system-info
 This is an x86_64 machine
 ```
 
+- `just_executable()` - Path to the running just executable.
+
+For example:
+
+```make
+itself:
+    @echo The executable is at: {{just_executable()}}
+```
+
+```
+$ just itself
+The executable is at: /bin/just
+```
+
 ==== Environment Variables
 
 - `env_var(key)` – Retrieves the environment variable with name `key`, aborting if it is not present.
--- a/README.adoc
+++ b/README.adoc
@@ -585,20 +585,6 @@ $ just system-info
 This is an x86_64 machine
 ```
 
-- `just_executable()` - Path to the running just executable.
-
-For example:
-
-```make
-itself:
-    @echo The executable is at: {{just_executable()}}
-```
-
-```
-$ just itself
-The executable is at: /bin/just
-```
-
 ==== Environment Variables
 
 - `env_var(key)` – Retrieves the environment variable with name `key`, aborting if it is not present.
@@ -634,6 +620,22 @@ For example, to run a command relative to the location of the current justfile:
 ```
 script:
   ./{{justfile_directory()}}/scripts/some_script
+```
+
+==== Just Executable
+
+- `just_executable()` - Absolute path to the just executable.
+
+For example:
+
+```make
+executable:
+    @echo The executable is at: {{just_executable()}}
+```
+
+```
+$ just
+The executable is at: /bin/just
 ```
 
 ==== Dotenv Integration
--- a/src/function.rs
+++ b/src/function.rs
@@ -126,16 +126,13 @@ fn env_var_or_default(
 }
 
 fn just_executable(_context: &FunctionContext) -> Result<String, String> {
-  let exe_path = std::env::current_exe()
-    .map_err(|e| format!("Error getting just executable: {}", e))?;
+  let exe_path =
+    std::env::current_exe().map_err(|e| format!("Error getting just executable: {}", e))?;
 
-  exe_path
-    .to_str()
-    .map(str::to_owned)
-    .ok_or_else(|| {
-      format!(
-        "Just executable path is not valid unicode: {}",
-        exe_path.to_string_lossy()
-      )
-    })
+  exe_path.to_str().map(str::to_owned).ok_or_else(|| {
+    format!(
+      "Just executable path is not valid unicode: {}",
+      exe_path.to_string_lossy()
+    )
+  })
 }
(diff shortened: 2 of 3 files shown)


bew:
For the windows build, I'm not sure how to fix this:
```
---- misc::test_just_executable_function stdout ----
thread 'misc::test_just_executable_function' panicked at 'assertion failed: `(left == right)`: bad output

Diff < left / right > :
 Output {
<    stdout: "Executable path is: D:ajustjusttargetdebugjust.exe\n",
>    stdout: "Executable path is: D:\\a\\just\\just\\target\\debug\\just.exe\n",
     stderr: "",
     status: 0,
 }

', tests\test.rs:123:5
```
Do I need to change the value returned by the function to make `\` visible? Like using `path.display()` before returning or sth like that?

_still new to rust_

tests/misc.rs
```diff
@@ -1203,6 +1203,15 @@ test! {
   status:   EXIT_FAILURE,
 }
 
+test! {
+  name:     test_just_executable_function,
+  justfile: "a:\n  @echo Executable path is: {{just_executable()}}",
```
casey: You can use multi-line strings here. Also, I think the windows test failure is due to backslashes in the path being interpreted. Try this:

```rust
test! {
  name:     test_just_executable_function,
  justfile: "
    a:
      @echo Executable path is: '{{just_executable()}}'
    ",
```

(Multi-line strings aren't a feature of rust, they're from the test macro.)

src/function.rs
```diff
@@ -123,3 +124,15 @@ fn env_var_or_default(
     Ok(value) => Ok(value),
   }
 }
+
+fn just_executable(_context: &FunctionContext) -> Result<String, String> {
+  let exe_path =
+    std::env::current_exe().map_err(|e| format!("Error getting just executable: {}", e))?;
```
casey: How about `Error getting current executable: {}`. Function error messages are printed, it will refer to the function name, which will mention `just_executable`.

src/function.rs
```diff
@@ -123,3 +124,15 @@ fn env_var_or_default(
...
+
+  exe_path.to_str().map(str::to_owned).ok_or_else(|| {
+    format!(
+      "Just executable path is not valid unicode: {}",
```
casey: How about `Executable path is not valid unicode: {}`

README.adoc
```diff
@@ -585,6 +585,20 @@ $ just system-info
 This is an x86_64 machine
 ```
 
+- `just_executable()` - Path to the running just executable.
```
casey: `just_executable` isn't super similar to existing functions, so I think it should get it's own section after the `Justfile and Justfile Directory` section. Something like:

```
==== Just Executable

- `just_executable()` - Absolute to the just executable.
```

README.adoc
```diff
@@ -585,6 +585,20 @@ $ just system-info
...
+
+For example:
+
+```make
+itself:
```
casey: `itself` is a kinda weird name, how about just "executable".

README.adoc
```diff
@@ -585,6 +585,20 @@ $ just system-info
...
+    @echo The executable is at: {{just_executable()}}
+```
+
+```
+$ just itself
```
casey: Since the first recipe is run by default, this can just be:

```
$ just
The executable is at: /bin/just
```

casey (changes requested):
Nice, thank you for the PR! I made comments regarding style and formatting, but otherwise this looks great.

bew:
I changed the PR following your comments!
I don't know how you prefer to do reviews, if you want me mark your comments as resolved or not, so I didn't do it to be safe :)
(I usually prefer this anyway, so you can review the changes and mark-resolved your comments as you go)

Edit: windows test is not better :confused:

casey:
> I don't know how you prefer to do reviews, if you want me mark your comments as resolved or not, so I didn't do it to be safe :)

Either way is fine with me!

> Edit: windows test is not better 😕

Hmmm, try this:

```
printf '%s\n' '{{just_executable()}}'
```

The behavior of echo with respect to backslashes can be inconsistent. Printf is consistent across platforms, and does not interpret escape sequences in arguments after the first.

bew:
I got some weird messages with that printf line, and I had a hard time finding the reason ^^ With errors ranging from `Unexpected closing delimiter '}'` to `Unterminated string` or a crash of the test function..

The solution was to not use a `\n` but a `\\n` to ensure the newline is not included in the justfile, but only in the recipe' output.

_Let's see how Windows work now_
:tada: :tada: CI is finally green :tada: :tada:

casey:
Nice, merged! Thank you for the idea, and the implementation.

I just released [v0.8.6](https://github.com/casey/just/releases/tag/v0.8.6) that contains this change.

PR #785, opened 2021-03-31
Release v0.9.1

casey:
- Bump version: 0.9.0 → 0.9.1
- Update dependencies
- Update changelog
- Update man page
- Update config test


casey:
Actually, I'm not going to release this yet. This would have added a warning if a `.env` file is found but `dotenv-load` is unset. However, the `dotenv-load` setting was only released a few days ago, so this would have created a situation where if you were using multiple versions of just with the same justfile, you might not have been able to silence the warning by adding `set dotenv-load` because that would have caused an error when using the justfile with the older version of just without the setting.

PR #796, opened 2021-04-06
Remove interpolation stack

casey:


casey:
Good to know that this is possible, but leaving it for when I actually have a feature that depends on it.

PR #814, opened 2021-05-03
just is now in Arch official repos

svenstaro:
I pushed `just` to official Arch repos so the README should reflect that.


casey:
Awesome, thank you!

PR #826, opened 2021-05-11
Fix powershell shebang, closes #825

### LINKED ISSUE #825: #!powershell.exe does not works
```
set shell := ["powershell.exe", "-c"]

default:
    #!powershell.exe
    Write-Host Hello-World
```
We cannot create powershell shebang because "just" will create tmp script file without any ext

Powershell cannot run script without `.ps1` extension.

Maybe we should support custom extension to created script file.

sigoden:


casey:
Thanks for the PR!

Can you add a test that this works on windows? Check out the `tests` directory. You can put it in a new module, maybe `tests/shebang.rs`.

Also, what if the users writes:

```
foo:
  #!powershell
```

…without `.exe`. If that works, we should check for that too. Also, do we need to add a suffix for `pwsh.exe`?

Even though this is windows specific, let's do it on every platform. The less platform-specific code there is the better, and it won't hurt to add a `.ps1` suffix on macOS and Linux.

sigoden:
`.exe` extensions is required on windows. So we woudl better use platform-specific code

src/platform_interface.rs
```diff
@@ -11,7 +11,13 @@ pub(crate) trait PlatformInterface {
   ) -> Result<Command, OutputError>;
   
   /// Get generate file ext
-  fn get_recipe_file_ext(command: &str) -> String;
+  fn get_script_file_ext(command: &str) -> &str {
+      if command.ends_with("powershell") {
```
casey: Should this also check for `powershell.exe`?

```suggestion
      if command.ends_with("powershell") || command.ends_with("powershell.exe") {
```

src/platform_interface.rs
```diff
@@ -11,7 +11,13 @@ pub(crate) trait PlatformInterface {
   ) -> Result<Command, OutputError>;
   
   /// Get generate file ext
-  fn get_recipe_file_ext(command: &str) -> String;
+  fn get_script_file_ext(command: &str) -> &str {
```
casey: Let's call this `get_script_file_suffix`, to match `std::env::consts::EXE_SUFFIX`.
```suggestion
  fn get_script_file_suffix(command: &str) -> &str {
```

src/recipe.rs
```diff
@@ -129,8 +129,8 @@ impl<'src, D> Recipe<'src, D> {
           io_error: error,
         })?;
       let mut path = tmp.path().to_path_buf();
-      let ext = Platform::get_recipe_file_ext(interpreter);
-      path.push(format!("{}{}", self.name(), ext));
+      let ext_name = Platform::get_script_file_ext(interpreter);
```
```suggestion
      let ext_name = Platform::get_script_file_suffix(interpreter);
```

casey:
```suggestion
      let suffix = Platform::get_script_file_suffix(interpreter);
```

tests/shebang.rs
```diff
@@ -0,0 +1,31 @@
...
+
+  let output = Command::new(executable_path("just"))
+    .current_dir(tmp.path())
+    .output()
+    .unwrap();
+
+  let stdout = "Hello-World\n";
+
+  assert_stdout(&output, stdout);
+}
```
casey: You can use the `test!` macro. I haven't tested it, but it should be something like this:

```suggestion
#![cfg(windows)]
test! {
  justfile: r#"
  default:
    #!powershell
     Write-Host Hello-World
"#,
  stdout: "Hello-World\n",
}

#![cfg(windows)]
test! {
  justfile: r#"
  default:
    #!powershell.exe
     Write-Host Hello-World
"#,
  stdout: "Hello-World\n",
}
```

src/recipe.rs
```diff
@@ -107,6 +107,20 @@ impl<'src, D> Recipe<'src, D> {
...
+        })?;
+
+      let Shebang {
+        interpreter,
+        argument,
+      } = Shebang::new(shebang_line).ok_or_else(|| RuntimeError::Internal {
+        message: format!("bad shebang line: {}", shebang_line),
+      })?;
+
+
```
```suggestion

```

tests/shebang.rs
```diff
@@ -0,0 +1,21 @@
...
+test! {
+  name:     powershell,
+  justfile: r#"
+default:
+  #!powershell
+  Write-Host Hello-World
+"#,
+  stdout: "Hello-World\n",
+}
+
```
casey:
```suggestion

```

PR #829, opened 2021-05-15
'else' is a keyword

SOF3:
It seems that `else` was incorrect typed as a rule instead of a keyword here.


casey:
Nice, thank you for the PR!

PR #831, opened 2021-05-16
NEWLINE should not match backslash-escaped feed

SOF3:
The definition of `TEXT` here is a bit confusing. Perhaps we shouldn't need this special case if `TEXT` defines how it deals with newlines?


casey:
Thanks for the PR!

Is `(?<!\\)` a negative lookbehind? I agree that `(?<!\\)\n|\n\r` is definitely more correct, although since the grammar is intended primarily as documentation, I'm inclined to keep it simple as opposed to making it perfect.

Maybe adding:

 ```
ESCAPED_NEWLINE = \\\n|\r\n
```

Would make it clearer. What do you think?

SOF3:
Sounds reasonable.

In fact, the documentation is quite unclear, in particular with newlines and indents. For example, there is no rigorous definition of `INDENT` (I understand that it *cannot* be rigorous because of `--list-heading` and `--list-prefix`, but let's assume the normal case; I don't know whether there actually exist people who want to make their lives harder with these options). In fact, the notion of `INDENT` and `DEDENT` turned out to complicate the definition, since there is only a maximum of one level of indentation in justfile.

Another confusing point is the definition of `TEXT`, which turned out to be matching just anything except a comment or newline, with the exception of backslash-escaped newline. I think a more clear definition of `TEXT` might suffice as well.

casey:
> In fact, the documentation is quite unclear, in particular with newlines and indents. For example, there is no rigorous definition of `INDENT`

Yeah, I agree it's not ideal. I think it's hard because the rigorous definition would be quite complicated, whereas the grammar should let users get an intuitive grasp on what's accepted that, as long as it doesn't confuse them, doesn't have to be totally correct.

> (I understand that it _cannot_ be rigorous because of `--list-heading` and `--list-prefix`, but let's assume the normal case; I don't know whether there actually exist people who want to make their lives harder with these options).

I might be misunderstanding, but `--list-heading` and `--list-prefix` both only affect the output of `--list`, and don't change the grammar, parsing, or lexing.

> In fact, the notion of `INDENT` and `DEDENT` turned out to complicate the definition, since there is only a maximum of one level of indentation in justfile.

I think there can be multiple levels of indentation produced by the lexer, but there are no constructs the parser currently accepts that include nested indentation, so it's always a parse error.

> Another confusing point is the definition of `TEXT`, which turned out to be matching just anything except a comment or newline, with the exception of backslash-escaped newline. I think a more clear definition of `TEXT` might suffice as well.

Another thing is that `TEXT` doesn't match `{{`, which starts an interpolation.

casey:
Maybe one way to simplify the presentation of the grammar is to allow multiple tokens to match the same text, but present them in priority order, so more specific tokens are produced instead of more general tokens (an interpolation token with `{{` instead of a text token with `{{`), but general tokens stay simple, since they don't have to say "any text except for a `{{`".

casey:
Closing this due to inactivity.

PR #835, opened 2021-05-20
Add `vim-just` installation instructions to readme

casey:


casey:
@NoahTheDuke Just added `vim-just` to the readme. `vim-just` is now *way* better than `ft=make`, thanks again for your efforts!

Issue #817, opened 2021-05-06
Autoformatter/prettifier for justfiles


kwshi:
Same idea as `pretty` for JS/CSS/HTML and `black` for Python, but for justfiles.  It looks like `just`'s syntax tree representation is pretty comprehensive already and simple enough that it shouldn't be too hard to write a `Module -> String` pretty-printer that can be used to autoformat justfiles.  Also I think it would be a nice to have.

casey:
This would be great!

I would suggest making it a subcommand of Just, invoked with `just --fmt`, that rewrites the justfile in place with a formatted version.

`Justfile` already implements `Display`, and produces a valid justfile. However, it strips comments, and doesn't preserve the order of recipes in a justfile.

So giving `Module` a `Display` implementation, or a `pretty_print` method, would probably be the way to go.

casey:
This was done in #837.

PR #837, opened 2021-05-23
Add `--fmt` subcommand

vglfr:
Implements #817.

Things I'd like to get feedback so I could fix them:

- copypasted `impl<'src> Display for UnresolvedRecipe<'src>` (cannot wrestle borrow checker)
- ugly `tree: (justfile ())` instead of `tree: (justfile)` to pass tests (not sure that matters)
- justfile in repository uses tabs but `--fmt` formats them to spaces, probably because it reuses space-based impl originally used for `--dump` (I don't have any preference on the matter)


casey:
Thanks for taking a crack at this:

> * copypasted `impl<'src> Display for UnresolvedRecipe<'src>` (cannot wrestle borrow checker)

What was the issue that you ran into? I think you could bound `D` in Recipe by `Display`:

```rust
pub(crate) struct Recipe<'src, D: Display = Dependency<'src>> {
```

You'll also need to add it to the impl blocks:

```rust
impl<'src, D: Display> Recipe<'src, D> {
```

> * ugly `tree: (justfile ())` instead of `tree: (justfile)` to pass tests (not sure that matters)

If there are no comments, shouldn't that avoid an empty `()`?

> * justfile in repository uses tabs but `--fmt` formats them to spaces, probably because it reuses space-based impl originally used for `--dump` (I don't have any preference on the matter)

I'm not sure what `--fmt` should use by default. I use tabs, because for a long time I used make-based syntax highlighting in vim, which highlights space-indentation as an error, since makefiles only use tabs.

I used to dislike tabs, but have actually come around to them, since tabs let everyone use a different tab-width. I'm inclined to keep them, and change `--dump` to use tabs.

casey:
Actually, for now, let's use spaces, just to keep the PR-size reasonable. We can adjust `--fmt` defaults and configuration going forward.

vglfr:
Fixed it, thanks! It turns out `impl<'src> Display for Recipe<'src>` needs to be replaced by `impl<'src, D: Display> Display for Recipe<'src, D>`.

As for `(justfile ())` it's probably false alarm. In that particular case `justfile.tree()` is called on `# foo`. Can't see any harm done, it's just untidy.

casey:
Let me know when this is ready to review. The main thing I'd like to see are a bunch of tests, probably for every construct, that the output of `--fmt` is reasonable, and also a valid justfile. I think the output of `--dump` should be the same as `--fmt`, so the tests can call `--dump` and check the output. See the tests in `tests/misc.rs` for some examples. The new tests can go in `tests/fmt.rs`.

vglfr:
I think it's ready. Few random notes before review:
- `--fmt` and `--dump` are the same for recipes, assignments, and aliases; `--fmt` additionally supports sets and comments
- `--fmt` removes extra newlines between two aliases, sets, assignments, or comments, thus implementing basic grouping
- conditional inside recipe changed from `{{if ... { } else { } }}` to `{{ if ... { } else { } }}` while dropping trailing whitespace
- indented backticks keep their triple quotes (previously were formatted to single quote)

casey:
Awesome! I just ran CI, and I think the completion scripts are out of date.

> * `--fmt` and `--dump` are the same for recipes, assignments, and aliases; `--fmt` additionally supports sets and comments

I think that `--dump` and `--fmt` should produce the same output, so I would leave comments in `--dump`. I looked at the code, but it wasn't obvious to me why sets would be different between the two. Are they handled differently?

> * `--fmt` removes extra newlines between two aliases, sets, assignments, or comments, thus implementing basic grouping

Sounds good!

> * conditional inside recipe changed from `{{if ... { } else { } }}` to `{{ if ... { } else { } }}` while dropping trailing whitespace

Also good, it definitely looks better this way.

> * indented backticks keep their triple quotes (previously were formatted to single quote)

Nice. We might eventually want to make `--fmt` convert indented backticks to some standard formatting, but this is great for now.

vglfr:
Updated completions.

>I think that --dump and --fmt should produce the same output, so I would leave comments in --dump. I looked at the code, but it wasn't obvious to me why sets would be different between the two. Are they handled differently?

Yes, `--dump` isn't printing sets at all. I think that's original behavior. Adding sets to `--dump` should be straightforward. However it might be impossible to determine from `Justfile` struct whether setting was set explicitly (when it is default). So either all settings or non-default settings could be dumped. `--fmt` doesn't have this dilemma, it just formats all sets as they appear in justfile.

Recipe docstring comments are displayed in `--dump` all right, as well as comments in recipe body. However non-recipe-docstring top-level comments are skipped (including essential ones like first line shebang). Adding them to `--dump` would require modifying `Justfile` struct.

Please let me know what's the right way to handle those.

casey:
Ah, right. That makes sense. I think we should change it so that `--dump` prints out the `Module`, like `--fmt`. I don't think anyone is relying on comments being stripped from `--dump`, and making them do the same thing would allow users to, for example, redirect the output of `--dump` to a new justfile and have it be valid.

vglfr:
Updated. Redirecting `--dump` sounds cool.

casey:
Awesome! Just pushed a commit fixing a small formatting diff. I'll check this out today or tomorrow.

src/assignment.rs
```diff
@@ -2,3 +2,15 @@ use crate::common::*;
...
+
+impl<'src> Display for Assignment<'src> {
+  fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
+    write!(
+      f,
+      "{}{} := {}",
+      if self.export { "export " } else { "" },
+      self.name,
+      self.value
+    )
```
casey: How about this:

```suggestion
    if self.export {
      write!(f, "export ")?;
    }
    write!(
      f,
      "{} := {}",
      self.name,
      self.value
    )
```

It's two calls to `write!(…)`, but a little bit clearer.

src/config.rs
```diff
@@ -266,6 +269,11 @@ impl Config {
         "Evaluate and print all variables. If a variable name is given as an argument, only print \
          that variable's value.",
       ))
+      .arg(
+        Arg::with_name(cmd::FORMAT)
+          .long("fmt")
+          .help("Autoformat justfile and override it in-place"),
```
```suggestion
          .help("Format justfile and overwrite it in-place"),
```

src/config.rs
```diff
@@ -676,8 +687,12 @@ impl Config {
     self.run(justfile, search, overrides, &recipes)
   }
 
-  fn dump(justfile: Justfile) {
-    println!("{}", justfile);
+  fn dump(&self, src: &str) -> Result<(), i32> {
```
casey: Let's de-duplicate the `dump` and `format` functions. Something like:

```rust
fn format_src<'src>(&self, src: &'src str) -> CompilationResult<'src, String> {
    let tokens = Lexer::lex(src)?;
    let ast = Parser::parse(&tokens)?;
    Ok(ast.to_string())
}
```

Which can then be called from both `dump` and `format`.

src/module.rs
```diff
@@ -1,5 +1,7 @@
 use crate::common::*;
 
+use std::mem;
```
casey: This can be added to `common`.

src/parser.rs
```diff
@@ -1097,7 +1111,7 @@ mod tests {
   test! {
     name: comment,
     text: "# foo",
-    tree: (justfile),
+    tree: (justfile ()),
```
casey: Let's actually include the comment in the tree, since we want to preserve them. So this would be something like:

```rust
(justfile (comment "# foo"))
```

src/parser.rs
```diff
@@ -299,17 +299,31 @@ impl<'tokens, 'src> Parser<'tokens, 'src> {
     let mut items = Vec::new();
 
     let mut doc = None;
+    let mut comments: Vec<&'src str> = vec![];
```
casey: Is the lifetime annotation necessary? Could this just be:

```
let mut comments: Vec<&str> = vec![];
```

Also, when creating empty vecs, I usually use `Vec::new()`:

```suggestion
    let mut comments: Vec<&str> = Vec::new();
```

src/item.rs
```diff
@@ -5,6 +5,19 @@ use crate::common::*;
 pub(crate) enum Item<'src> {
   Alias(Alias<'src, Name<'src>>),
   Assignment(Assignment<'src>),
+  Comment(Vec<&'src str>),
```
casey: Is there a reason that comments are stored together, i.e. as `Comment(Vec<&'src str>)` instead of `Comment(&'src str)`?

src/parser.rs
```diff
@@ -1176,7 +1190,7 @@ mod tests {
 
       bar:
     ",
-    tree: (justfile (recipe bar)),
+    tree: (justfile () (recipe bar)),
```
```suggestion
    tree: (justfile (comment "# foo") (recipe bar)),
```

src/recipe.rs
```diff
@@ -344,7 +344,10 @@ impl<'src> Display for Recipe<'src> {
         }
         match fragment {
           Fragment::Text { token } => write!(f, "{}", token.lexeme())?,
-          Fragment::Interpolation { expression, .. } => write!(f, "{{{{{}}}}}", expression)?,
+          Fragment::Interpolation { expression, .. } => match expression {
```
casey: I'd actually be fine with adding space around all expressions. It would remove this special case, and might be clearer. What do you think?

src/config.rs
```diff
@@ -555,8 +565,9 @@ impl Config {
       Choose { overrides, chooser } =>
         self.choose(justfile, &search, overrides, chooser.as_deref())?,
       Command { overrides, .. } => self.run(justfile, &search, overrides, &[])?,
-      Dump => Self::dump(justfile),
+      Dump => self.dump(&src)?,
```
casey: We're currently compiling twice, which would be nice to avoid.

Instead of calling `Compiler::compile`, let's do something like:

```rust
    let tokens = Lexer::lex(src).eprint(self.color)?;
    let ast = Parser::parse(&tokens).eprint(self.color)?;
    let justfile = Analyzer::analyze(ast.clone()).eprint(self.color)?;
```

Then, the AST can be passed into `dump` and `format`, so we don't have to recompile.

casey (changes requested):
This looks great! Some minor comments.

src/recipe.rs
vglfr: Totally agree. It certainly looks cleaner to me.

src/item.rs
vglfr: Withouth `Vec` it is definitely cleaner.

vglfr:
Fixed. Thanks for a review!

casey (changes requested):
Looks good! A few minor things.

src/testing.rs
```diff
@@ -3,6 +3,7 @@ use crate::common::*;
 use pretty_assertions::assert_eq;
 
 pub(crate) fn compile(text: &str) -> Justfile {
+  use super::compiler::Compiler;
```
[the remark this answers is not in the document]
vglfr: It would give "unused import" compiler warning. We don't use `Compiler` anywhere except this test.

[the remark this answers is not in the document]
vglfr: We might drop `Compiler`, just didn't feel brave enough to do it on my own)

src/parser.rs
```diff
@@ -605,16 +611,20 @@ impl<'tokens, 'src> Parser<'tokens, 'src> {
 
     let body = self.parse_body()?;
 
-    Ok(Recipe {
+    let recipe = Recipe {
       private: name.lexeme().starts_with('_'),
       shebang: body.first().map(Line::is_shebang).unwrap_or(false),
       parameters: positional.into_iter().chain(variadic).collect(),
-      doc,
+      doc: doc.map(|x| x[1..].trim_start()),
```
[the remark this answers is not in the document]
vglfr: This is problematic. The logic of adding comment looks like:
- if comment is found and next token is not `Identifier` then add comment
- otherwise next `Identifier` could be:
  - if `Recipe` then add comment to a docstring
  - otherwise insert comment to a tree before that non-recipe identifier

The point of `&mut` is to signal `Recipe` to `parse_justfile` by setting `*doc = None`. Alternatives to this are:
- find a way to tell whether next identifier is recipe (which seems difficult because lines 317-355 are all about it)
- set `doc = None` in five possible places where `parse_recipe` is called

Current solution seemed a lesser evil. If there's an elegant one-line call to determine whether next identifier is recipe than it would be way better than this.

src/config.rs
```diff
@@ -713,6 +727,11 @@ impl Config {
     }
   }
 
+  fn format(ast: Module, search: &Search) -> Result<(), i32> {
+    fs::write(&search.justfile, ast.to_string()).unwrap();
```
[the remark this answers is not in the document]
vglfr: Trailing newline is added in `Display` for all items. Testing is difficult because in integration tests `\nset dotenv-load := true\n` is manually appended and then stripped, and when `set`s are tested extra newline should be stripped then because of the grouping so I end up with just `trim_end()` for both have and want.

Still this has to be tested. I'll come up with something.

src/testing.rs
```diff
@@ -3,6 +3,7 @@ use crate::common::*;
 use pretty_assertions::assert_eq;
```
casey, line 4:
```suggestion
use crate::compiler::Compiler;
use pretty_assertions::assert_eq;
```

src/testing.rs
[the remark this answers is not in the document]
casey: Ah, I see. In that case it can just be moved into the above move block.

src/parser.rs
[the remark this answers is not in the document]
casey: Thanks for the explanation. I agree, the current solution sounds like the best option. The logic of determining if the next identifier is a recipe is quite complicated.

src/config.rs
```diff
@@ -713,6 +727,24 @@ impl Config {
...
+        eprintln!(
+          "Failed to write justfile to `{}`: {}",
+          search.justfile.display(),
+          error
+        );
+      }
+      Err(EXIT_FAILURE)
+    } else {
+      if self.verbosity.loud() {
+        eprintln!("Wrote justfile to `{}`", search.justfile.display());
```
casey, line 742: This is nice!

casey:
I think that adding `dotenv-load := true` to tests was actually just a temporary hack to make refactoring easier. I removed it in #853, so I think all the special case stuff around it can be removed. Sorry for the churn! I should have removed it long ago.

vglfr:
Fixed, thanks!

src/config.rs
```diff
@@ -713,6 +727,24 @@ impl Config {
     }
   }
 
+  fn format(&self, ast: Module, search: &Search) -> Result<(), i32> {
+    if let Err(error) = fs::write(&search.justfile, ast.to_string()) {
```
```suggestion
    if let Err(error) = File::open(&search.justfile).and_then(|mut file| write!(file, "{}", ast)) {
```

casey:
Hooookay!

Sorry for the delay in reviewing this. I was in Miami for the Bitcoin conference. (Which was like a Bitcoin-themed circus, a bit superficial, but lots of fun.)

I changed the help message text a little bit, refactored the format subcommand to make `Display` write directly to disk instead of creating a temporary, and refactored the doc comment handling a little bit.

My concern with the doc comment handling was that doc comment might get "lost", i.e. never actually turned into an item. I changed it so that comments are always inserted into the `items` vector, and popped off and passed into each call of `parse_recipe`. (Only if `eol_since_last_comment` is false, which is set to true whenever the parser encounters a newline.)

I was going to say that this is ready to merge, but I thought of something looking over the code one last time. Comments that are on the same line as items aren't preserved. For example, this comment won't be preserved as an `Item`.

```
alias x := y # foo
```

Since this PR is already pretty large, I was thinking about merging this, and then adding an `--unstable` flag. Unless `--unstable` was passed, `--fmt` would return an error. My thinking is that it's okay if the output of `--dump` is missing things, but since `--fmt` overwrites the justfile, there's a higher bar for preserving everything in the justfile.

What do you think?

casey:
Just thinking out loud:

The method that swallows comments is `expect_eol`. I'm thinking that we should move the `items` vec into the `Parser` struct, and then have individual parsing functions push items into the vec, instead of returning items. This would allow methods like `expect_eol` to insert comments. I'm not sure the best way preserve the position of comments (whether they appear on lines by themselves or after items). We might have to create an `Item::Eol` that's created for each newline.

casey:
> Hi, thanks for fixes, they're really instructive (which is what I'm looking for - besides the fact that I'm using Just in all my projects and would love to pay back).

You bet!

> Same line comments is something I've overlooked. Totally fine with `--unstable`.

Awesome, I'll merge this now, and add an `--unstable` flag before the next release.

PR #838, opened 2021-05-25
Add dollar sign ($) to parameter grammar

NoahTheDuke:
Described [here](https://github.com/casey/just#environment-variables-1).


casey:
Nice catch!

PR #864, opened 2021-06-10
Add Sublime Text syntax file

casey:


Frederick888:
I'm thinking about adding `file_extensions: [Justfile]`, but I don't use Sublime so not sure if it's ok.

Although http://www.sublimetext.com/docs/syntax.html is unclear about files without extensions, it seems the same rule for files without basenames, i.e. using full name, also applies in practice. For example, https://github.com/sublimehq/Packages/blob/master/Makefile/Makefile.sublime-syntax.

I'm asking this cos I'd like to use this file in [bat](https://github.com/sharkdp/bat) and it ignores syntaxes without this attribute: https://github.com/sharkdp/bat/blob/12ecb325c911d2f059962ca71f5d90b87f92dc85/src/bin/bat/main.rs#L87

casey:
Interesting! I don't have any experience with Sublime. It should probably be `[Justfile, just]`, since `.just` is also a possibility.

I'm guessing that anything that doesn't understand this attribute will just ignore it, so I'd be happy to merge a PR adding the attribute.

PR #872, opened 2021-06-14
Adding some path functions

### LINKED ISSUE #871: [Enhancement proposal] Replacement Strings
I think that a feature like [GNU parallel's replacement strings](https://www.gnu.org/software/parallel/parallel_tutorial.html#Replacement-strings) would be handy.

I was wondering something like:
```make
foo := "A/B/C.D"

echo {{foo.}}   // A/B/C   (Removes extension)
echo {{foo/}}   // C.D     (Removes the path)
echo {{foo//}}  // A/B     (Keeps the path only)
echo {{foo/.}}  // C       (Removes path and extension)
```

What do you think @casey?

I'm terribly sorry since I don't know Rust enough to contribute by myself and that's why this is an issue and not a PR. If I'll spot the parser I'll begin to do some test.

TonioGela:
I've tried to add the functions I talked about in #871, mirroring the standard library functions from `Path`.
I've added a bunch of tests, but I haven't updated the documentation.
I'm looking forward to getting this pull request reviewed because it will be my first rust contribution if it gets merged.
For the very same reason, it could be non-optimal or straightforward.
Closes #871.


casey:
Awesome, thanks for tackling this! I probably won't be able to take a look at this today, but I should have time tomorrow. In the meantime, it looks like there are some errors from clippy on CI.

TonioGela:
> Awesome, thanks for tackling this! I probably won't be able to take a look at this today, but I should have time tomorrow. In the meantime, it looks like there are some errors from clippy on CI.

Managed it, thanks!

casey:
Just took a quick look, and overall I think it looks good. You're running into a lot of the things that make Rust painful, for example, the fact that paths might not be valid Unicode.

I think one thing that might make this a whole lot nicer is if you use [camino](https://docs.rs/camino/1.0.4/camino/)'s `Utf8Path` type. That way, you can do `Utf8Path::new(argument)`, which will always succeed, and thereafter you won't have to deal with the error case of non-unicode paths.

The other thing that would be nice would be to de-deuplicate these format calls:

```
    .ok_or_else(|| {
      format!(
        "Cannot get file stem from {}",
        if file_path.is_empty() {
          "an empty string"
        } else {
          file_path
        }
      )
```

I'm not entirely sure what the best way to do this would be, but you could do something like:

```
fn extraction_error_message(part: &string, path: &Utf8Path) -> String {
  format!("Cannot extract {} from `{}`, part, path)
}
```

If the path is set in backticks, then I think:

```
Cannot extract X from ``
```

Is a good error message.

casey:
Actually, you could skip the function and just put `format!("Cannot extract file stem from `{}`, path)`. Without the `if … { … } else { … }` I don't think it's complicated enough to deserve a function.

casey:
Also, thanks for adding tests! That's something that a lot of people forget to do.

TonioGela:
> Also, thanks for adding tests! That's something that a lot of people forget to do.

My pleasure. I reimplemented the functions using camino and simplifying the error handling. I'm totally not sure about dependency management. Am I supposed to commit the Cargo.lock file too?

TonioGela:
And btw compliments for the functions "interface", adding functionality (even coming from a totally different language like Scala) was pretty trivial since it is really well designed ;)

src/function.rs
```diff
@@ -136,3 +141,42 @@ fn just_executable(_context: &FunctionContext) -> Result<String, String> {
     )
   })
 }
+
+fn file_name(_context: &FunctionContext, file_path: &str) -> Result<String, String> {
+  Utf8Path::new(file_path)
+    .file_name()
+    .map(std::borrow::ToOwned::to_owned)
```
casey: You can do `str::to_owned` here, which I think is a little is a little more readable.

Cargo.toml
```diff
@@ -32,6 +32,7 @@ strum_macros  = "0.21.1"
 target        = "1.0.0"
 tempfile      = "3.0.0"
 unicode-width = "0.1.0"
+camino        = "1.0.4"
```
casey: This list should be sorted.

src/function.rs
```diff
@@ -1,7 +1,7 @@
 use crate::common::*;
 
+use camino::Utf8Path;
```
casey: You can add this import to `common.rs`

src/function.rs
```diff
@@ -136,3 +141,42 @@ fn just_executable(_context: &FunctionContext) -> Result<String, String> {
     )
   })
 }
+
+fn file_name(_context: &FunctionContext, file_path: &str) -> Result<String, String> {
```
casey: For the argument name for this and the other functions,, how about `path` instead?

src/function.rs
```diff
@@ -136,3 +141,42 @@ fn just_executable(_context: &FunctionContext) -> Result<String, String> {
...
+  Utf8Path::new(file_path)
+    .file_name()
+    .map(std::borrow::ToOwned::to_owned)
+    .ok_or_else(|| format!("Cannot get file name from `{}`", file_path))
+}
+
+fn parent_directory(_context: &FunctionContext, file_path: &str) -> Result<String, String> {
+  Utf8Path::new(file_path)
+    .parent()
+    .map(Utf8Path::as_str)
```
casey: I think these two maps can be replaced with `Utf8Path::to_string`

src/function.rs
```diff
@@ -136,3 +141,42 @@ fn just_executable(_context: &FunctionContext) -> Result<String, String> {
...
+    .map(std::borrow::ToOwned::to_owned)
+    .ok_or_else(|| format!("Cannot get parent directory from `{}`", file_path))
+}
+
+fn file_stem(_context: &FunctionContext, file_path: &str) -> Result<String, String> {
+  Utf8Path::new(file_path)
+    .file_stem()
+    .map(std::borrow::ToOwned::to_owned)
```
casey: `str::to_owned`

src/function.rs
```diff
@@ -136,3 +141,42 @@ fn just_executable(_context: &FunctionContext) -> Result<String, String> {
...
+    .ok_or_else(|| format!("Cannot get file stem from `{}`", file_path))
+}
+
+fn without_extension(_context: &FunctionContext, file_path: &str) -> Result<String, String> {
+  let path = Utf8Path::new(file_path);
```
casey: I think it would be better to do this in two steps, first get the parent and report an error if you can't get it, then get the file_stem and report an error, and then return the error.

Also, you can do `Utf8Path::join` instead of `format!("{}/{}", …)`.

src/function.rs
```diff
@@ -136,3 +141,42 @@ fn just_executable(_context: &FunctionContext) -> Result<String, String> {
...
+    .map(Utf8Path::as_str)
+    .zip(path.file_stem())
+    .map(|(p, n)| format!("{}/{}", p, n))
+    .ok_or_else(|| format!("Cannot remove extension from `{}`", file_path))
+}
+
+fn extension(_context: &FunctionContext, file_path: &str) -> Result<String, String> {
+  Utf8Path::new(file_path)
+    .extension()
+    .map(std::borrow::ToOwned::to_owned)
```
casey: `str::to_owned`

casey (commented):
Looking better! A few more comments.

As for the lock file, yes, that should be committed. The general rule is that for binary crates, such as `just`, the lockfile should be committed so that builds are reproducible. For library crates, the lockfile shouldn't be committed, since it's up to the binary crate that the library is eventually included in to determine dependency versions to use.

src/function.rs
casey: For example:

```
let parent = Utf8Path::new(file_path).parent().ok_or_else(|| format!("Cannot extract parent from `{}`, path)?;
```

TonioGela:
> Looking better! A few more comments.
> 
> As for the lock file, yes, that should be committed. The general rule is that for binary crates, such as `just`, the lockfile should be committed so that builds are reproducible. For library crates, the lockfile shouldn't be committed, since it's up to the binary crate that the library is eventually included in to determine dependency versions to use.

Makes totally sense

casey:
Not sure if you saw, but I added some comments in-line, [for example](https://github.com/casey/just/pull/872#discussion_r652150228).

src/function.rs
TonioGela: I'm not sure if I added it in the correct way

TonioGela:
> Not sure if you saw, but I added some comments in-line, [for example](https://github.com/casey/just/pull/872#discussion_r652150228).

Yes, I read the previous comment from the phone, and it hid the in-line comments. I think I addressed them all now.

casey:
Looks great! How about using consistent language in all the error message, so:

```
Cannot extract X from `{}`
```

Also, instead of:

```
Cannot extract file_stem from `{}`
```

How about:

```
Cannot extract file stem from `{}`
```

TonioGela:
Typo, my bad. Is [this](https://github.com/casey/just/pull/872#discussion_r652194963) okay?

casey:
Ahhh, damn, sorry, I forgot to hit the "submit review" button, so you couldn't see any of them.

TonioGela:
Just applied everything and verified locally that both tests and clippy are good

casey:
Great, looks good, although I think there's a formatting issue.

casey:
Merged!

PR #878, opened 2021-06-24
Add file_extensions to Sublime syntax

Frederick888:
Although [1] is unclear about files without extensions, it seems the
same rule for files without basenames, i.e. using full name, also
applies in practice. For example, [2].

This also allows bat [3] to import this file [4].

In addition to `Justfile`, `just` is also added as suggested [5].

[1] http://www.sublimetext.com/docs/syntax.html
[2] https://github.com/sublimehq/Packages/blob/master/Makefile/Makefile.sublime-syntax
[3] https://github.com/sharkdp/bat
[4] https://github.com/sharkdp/bat/blob/12ecb325c911d2f059962ca71f5d90b87f92dc85/src/bin/bat/main.rs#L87
[5] https://github.com/casey/just/pull/864#issuecomment-867398424

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -218,6 +218,7 @@ dependencies = [
  "test-utilities",
  "unicode-width",
  "which",
+ "yaml-rust",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


casey:
Looks good to me! I also added a [test](https://github.com/casey/just/pull/878/commits/436e559a871bf5f2e7c81faed731b6a6e0c9b762) that will make sure that the file continues to be valid YAML in the future.

PR #902, opened 2021-07-10
Expose library interface

casey:
- [x] Make `run` take arguments
- [ ] Use `get_matches_from_safe` and print help and version text ourselves
- [ ] Eliminate all uses of `std::process::exit`, to avoid terminating calling process
- [ ] Figure out how to deal with signal handler

---

Some things which are out of scope for the first MVP, but which might be nice for later:

- Don't print to stdout/stderr, or make it configurable where messages are printed to
- Pass in environment variables
- Other things which should be passed in?

---

@thomcc I decided to not go with returning an `Error` type for now, and just return an `i32`. Since this is really the first release, I think planning for a few breaking changes is reasonable.

Currently, there are many places where `just` might exit the calling process, which is clearly undesirable. For example, if the arguments are bad, or if a command fails, or if the `ctrl-c` interrupt handler gets invoked.

If this is useful to you, I could document these shortcomings, merge it as-is, and then worry about fixing those things later.

What do you think?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,5 @@
 fn main() {
-  if let Err(code) = just::run() {
+  if let Err(code) = just::run(std::env::args_os()) {
     std::process::exit(code);
   }
 }
--- a/src/run.rs
+++ b/src/run.rs
@@ -1,6 +1,6 @@
 use crate::common::*;
 
-pub fn run() -> Result<(), i32> {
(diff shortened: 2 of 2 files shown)


casey:
Closing this due to inactivity.

thomcc:
Ah, sorry, I stopped using just (I was also on vacation when this was uploaded, and forgot to get back to it afterwards). This would have helped at the time, though.

casey:
No worries, feel free to ping me in the future if you wind up wanting it.

PR #904, opened 2021-07-13
Fix typos and capitalization of words

mtsknn:


README.adoc
```diff
@@ -258,7 +258,7 @@ $ just another-recipe
 This is another recipe.
 ```
 
-`just` prints each command to standard error before running it, which is why `echo 'This is a recipe!'` was printed. This is suppressed for lines starting with `@`, which is why `echo 'Another recipe.'` was not printed.
+Just prints each command to standard error before running it, which is why `echo 'This is a recipe!'` was printed. This is suppressed for lines starting with `@`, which is why `echo 'Another recipe.'` was not printed.
```
mtsknn: Btw is "standard error" correct or should it be "standard output"?

casey: Yup, standard error is correct.

README.adoc
```diff
@@ -384,7 +384,7 @@ Available recipes:
 
 The argument to `--list-heading` replaces both the heading and the newline
 following it, so it should contain a newline if non-empty. It works this way so
-you can suppress the heading line entirely by passing the empty string:
+you can suppress the heading line entirely by passing an empty string:
```
casey: "the empty string" sounds weird, but it's correct, so let's keep it. ("an empty string" is also correct, of course)

README.adoc
```diff
@@ -67,11 +67,11 @@ Just has a ton of useful features, and many improvements over Make:
   link:https://github.com/casey/just#writing-recipes-in-other-languages[arbitrary
   languages], like Python or NodeJS.
 
-- `just` can be invoked from any subdirectory, not just the directory that contains the `Justfile`.
+- `just` can be invoked from any subdirectory, not just the directory that contains the `justfile`.
 
 - And link:https://github.com/casey/just#manual[much more]!
 
-If you need help with `just` please feel free to open an issue or ping me on link:https://discord.gg/ezYScXR[discord]. Feature requests and bug reports are always welcome!
+If you need help with Just please feel free to open an issue or ping me on link:https://discord.gg/ezYScXR[Discord]. Feature requests and bug reports are always welcome!
```
casey: Here and later on, I think I actually prefer "`just`" instead of "Just". "Just" is such a common english word that I think it's good to use backticks to set it off.

README.adoc
```diff
@@ -142,7 +142,7 @@ curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -
 
 == Editor Support
 
-`justfile` syntax is close enough to `make` that you may want to tell your editor to use make syntax highlighting for just.
+`justfile` syntax is close enough to `make` that you may want to tell your editor to use Make syntax highlighting for Just.
```
casey: Ditto here, I think `make` is clearer than Make.

README.adoc
```diff
@@ -1167,10 +1167,10 @@ It isn't strictly necessary, but `set -euxo pipefail` turns on a few useful
...
 recipe:
 
-- `set -e` makes bash exit if a command fails.
-- `set -u` makes bash exit if a variable is undefined.
-- `set -x` makes bash print each script line before it's run.
-- `set -o pipefail` makes bash exit if a command in a pipeline fails.
+- `set -e` makes Bash exit if a command fails.
+- `set -u` makes Bash exit if a variable is undefined.
+- `set -x` makes Bash print each script line before it's run.
+- `set -o pipefail` makes Bash exit if a command in a pipeline fails.
```
```suggestion
- `set -e` makes `bash` exit if a command fails.
- `set -u` makes `bash` exit if a variable is undefined.
- `set -x` makes `bash` print each script line before it's run.
- `set -o pipefail` makes `bash` exit if a command in a pipeline fails.
```

casey (commented):
Thank you for this PR! I think that I like having binary names be lowercase and in backticks, as opposed to using the capitalized name. Would you mind changing instances of `just`, `make`, `bash`, and `cargo` to use backticks?

README.adoc
mtsknn: Done

README.adoc
mtsknn: Reverted

mtsknn:
Yep, "just" is a common word so I originally capitalized it everywhere to make it clearer that we are talking about `just`.

I now changed all instances to "`just`", except in headings it's still "Just". Similarly for `make`, `bash` and `cargo`. (Though line 1545 still has "Bash" instead of "`bash`" because I think in that case the capitalized name is clearer and consistent with the other names on that line. Similarly for "Cargo" on line 109.)

README.adoc
mtsknn: Done

README.adoc
mtsknn: 👍 Done

casey:
Awesome, looks great. This is a big improvement to the consistency of the readme. Thank you very much!

PR #907, opened 2021-07-19
[WIP] Allow duplicates

MadBomber:
This is a draft.  I do not know the Rust language.  Once past the adding of the new flag to the config object, I then got completely bogged down in the analyze_recipe function since I _assumed_ that config was a global object.

So why is allowing duplicate recipes a good thing?  In the context of the ability to import/include/require/with other files that contain justfile content it is likely that lower-level justfiles can include upper-level files (based on directory hierarchy or any other kind of organization) and over-ride previously defined recipes.

Consider this kind of directory organization:
```
projects
├── justfile
├── one
│   └── justfile
├── three
│   └── justfile
└── two
    └── justfile
```
At the `projects` directory level we have generic recipes that we want to be available in each of the projects one, two and three.  So the `justfile` in each of those specific projects "includes" the upper-level projects/justfile... but because of the specific projects unique needs, one or more of the "included" recipes from `projects/justfile` needs to be over-ridden.

Currently the error checking for duplicate recipes will not allow a recipe to be redefined.... well, and there is that no import/include functionality exists within just - yet.   That is why I wrote `justprep` as a preprocessor.  The hope is that when the include/import capability gets added to just then my justprep stop-gap will not be necessary.

But first we need to allow duplicates.

### DIFF
--- a/src/analyzer.rs
+++ b/src/analyzer.rs
@@ -91,13 +91,20 @@ impl<'src> Analyzer<'src> {
     })
   }
 
-  fn analyze_recipe(&self, recipe: &UnresolvedRecipe<'src>) -> CompilationResult<'src, ()> {
-    // if let Some(original) = self.recipes.get(recipe.name.lexeme()) {
-    //   return Err(recipe.name.token().error(DuplicateRecipe {
-    //     recipe: original.name(),
-    //     first:  original.line_number(),
-    //   }));
-    // }
+  fn analyze_recipe(
+                    &self,
+                    recipe: &UnresolvedRecipe<'src>
+                  ) -> CompilationResult<'src, ()> {
+
+    if let Some(original) = self.recipes.get(recipe.name.lexeme()) {
+      // TODO if !config.allow_duplicates {
+      //      The problem is that "config" is not defined - I thought it was global
+        return Err(recipe.name.token().error(DuplicateRecipe {
+            recipe: original.name(),
+            first:  original.line_number(),
+          }));
+      // TODO }
+    }
 
     let mut parameters = BTreeSet::new();
     let mut passed_default = false;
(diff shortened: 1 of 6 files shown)


tests/misc.rs
```diff
@@ -1305,18 +1305,18 @@ test! {
...
+//   justfile: "b:\nb:",
+//   args:     ("b"),
+//   stdout:   "",
+//   stderr:   "error: Recipe `b` first defined on line 1 is redefined on line 2
+//   |
+// 2 | b:
+//   | ^
+// ",
+//   status:   EXIT_FAILURE,
+// }
```
MadBomber, line 1319: This change of the test conditions needs to be reverted and made conditional based on the value of the config.allow_duplicates flag.

src/compilation_error.rs
```diff
@@ -103,7 +103,7 @@ impl Display for CompilationError<'_> {
       DuplicateRecipe { recipe, first } => {
         writeln!(
           f,
-          "Recipe `{}` first defined on line {} is redefined on line {}",
+          "TDV: Recipe `{}` first defined on line {} is redefined on line {}",
```
MadBomber, line 106: TDV that is my initials.  Its a place holder to remind me to come back here and do something.  Its like a TODO but at this point I don't know what should be done.

src/analyzer.rs
```diff
@@ -91,12 +91,19 @@ impl<'src> Analyzer<'src> {
...
+                    recipe: &UnresolvedRecipe<'src>
+                  ) -> CompilationResult<'src, ()> {
+
     if let Some(original) = self.recipes.get(recipe.name.lexeme()) {
-      return Err(recipe.name.token().error(DuplicateRecipe {
-        recipe: original.name(),
-        first:  original.line_number(),
-      }));
+      // TODO if !config.allow_duplicates {
+      //      The problem is that "config" is not defined - I thought it was global
```
MadBomber, line 101: There do not appear to be any existing CLI Flags that are used during the analysis phase.  I was using the --dry-run process as a guide but it pattern does not fit here.  I do not know the Rust language, so all of my attempts at bringing in the config context to the function have failed.

casey: I've been doing some big refactors, so things have moved around a lot, but `Subcommand::run` has the config and creates the `Analyzer`, so it would be the natural place to pass it in:

https://github.com/casey/just/blob/7efb82f4cb6cefa98a543da4e46f458fd6aeb7ed/src/subcommand.rs#L61

Rust is pretty brutal, so this might be a hard first PR!

casey:
Closing this. I think this is an interesting approach, but I think explicit module support will be cleaner and have fewer edge-cases.

MadBomber:
Okay. Sorry I could move this along.  I am not comfortable with Rust. 

Dewayne
   o-*

PR #924, opened 2021-07-28
Allow conditionally exporting variables

### LINKED ISSUE #922: Allow conditional export of variables
Something like:

```
export FOO := "hello" if "a" == "b"
```

The just variable `FOO` would always be defined, but would only be exported if the condition evaluated to true.

This syntax is quite weird, but it's the best I could come up with.

casey:
This is done except for the final syntax. See #922 for discussion.


casey:
This is a bit insane, and I can't figure out how to make this less insane, so I'm going to close it until such time as less insane syntax and semantics are discovered.

PR #933, opened 2021-08-02
Generate changelog from commit history

casey:


simonsan:
Not sure if you are already aware of https://github.com/saschagrunert/git-journal

Just want to let it here, as it might spare some work. 
Cheers

casey:
> Not sure if you are already aware of https://github.com/saschagrunert/git-journal
> 
> Just want to let it here, as it might spare some work.
> Cheers

Thanks for the pointer! I didn't know about `git-journal`. It looks promising, but one tricky aspect is that it looks like you have add metadata to commit messages in order for it to format them for the commit log. That's not ideal for `just`, since older commits don't have that metadata. I'd also have to remember to be consistent in adding that metadata in the future.

simonsan:
> 
> 
> > Not sure if you are already aware of https://github.com/saschagrunert/git-journal
> > Just want to let it here, as it might spare some work.
> > Cheers
> 
> Thanks for the pointer! I didn't know about `git-journal`. It looks promising, but one tricky aspect is that it looks like you have add metadata to commit messages in order for it to format them for the commit log. That's not ideal for `just`, since older commits don't have that metadata. I'd also have to remember to be consistent in adding that metadata in the future.

Just found https://github.com/orhun/git-cliff today, which might solve that with regex based custom parsers? Anyway, loving just and keep up the nice work!

casey:
> Just found https://github.com/orhun/git-cliff today, which might solve that with regex based custom parsers? Anyway, loving just and keep up the nice work!

Nice, I'll check it out. Glad you're enjoying just! Getting nice feedback makes it all worth it :)

casey:
Closing this, since I'm working on a general-purpose tool to do thi.

Issue #802, opened 2021-04-15
Allow choosing the dotenv filename with a command line flag


casey:
Requested by a user in Discord, allow setting the dotenv filename with a flag, for example:

```
$ just --dotenv-name .env-test
```

I'm not sure if this should be called `--dotenv-name` or `--dotenv-filename`.

bddap:
`--dotenv-file`?

casey:
> `--dotenv-file`?

I'm imagining that there might be one flag that takes a filename, and one that takes a full path, which is why I'm leaning towards `--dotenv-filename` and `--dotenv-path`. If it's `--dotenv-file` and `--dotenv-path`, it would be a big ambiguous which is which.

bddap:
From the readme:

> just will load environment variables from a file named `.env`. This file can be located in the same directory as your justfile or in a parent directory.

Ah. I see. So `--dotenv-filename` would not accept a path as `--dotenv-path` might.

`just --dotenv-filename .env-test` and `just --dotenv-path .env-test` would both be valid but the first invocation would also check parent directories for files named `.env-test`.

casey:
Yup, that's right.

casey:
And maybe if people want it, `--dotenv-dir` will set the directory to start looking for a dotenv file.

Celeo:
If the user specifies `--dotenv-path` and that path doesn't resolve to a file, should an error be returned, or should the search for the environment file "fall back" to looking for an `.env` file?

casey:
Fixed in #941.

PR #941, opened 2021-08-05
Add support for specifying name/path of .env

Celeo:
#802

This allows users to overwrite just looking for a file called .env in the working directory or its ancestors.

## Open questions

1. I have these two new args, `--dotenv-filename` and `--dotenv-path` mutually exclusive, with the latter taking precedence in the code over the former. Is that correct?
2. Currently, if the path from `--dotenv-path` does not resolve to a file, then the existing functionality of searching for a file called ".env" in the working directory or its ancestors is taken. If the type of `Config::dotenv_path` was an `Option<PathBuf>`, I could differentiate between "user set" and "default value", but I didn't see any of the other args as options so I'm hesitant to introduce that without checking with you first. What's the desired behavior here?
3. Do you want tests of the overridden dotenv file lookup?


src/config.rs
```diff
@@ -317,6 +321,19 @@ impl Config {
           .long("variables")
           .help("List names of variables"),
       )
+      .arg(
+        Arg::with_name(arg::DOTENV_FILENAME)
+          .long("dotenv-filename")
+          .takes_value(true)
+          .help("Load from environment file <DOTENV-FILENAME> instead of .env")
```
```suggestion
          .help("Search for environment file named <DOTENV-FILENAME> instead of `.env`")
```

src/config.rs
```diff
@@ -317,6 +321,19 @@ impl Config {
...
+          .conflicts_with(arg::DOTENV_PATH),
+      )
+      .arg(
+        Arg::with_name(arg::DOTENV_PATH)
+          .long("dotenv-path")
+          .help("Load from environment file at <DOTENV-PATH> instead of ./.env")
```
```suggestion
          .help("Load environment file at <DOTENV-PATH> instead of searching for `.env` file")
```

src/load_dotenv.rs
```diff
@@ -1,43 +1,57 @@
...
   config: &Config,
   settings: &Settings,
-  working_directory: &Path,
+  path: &Path,
 ) -> RunResult<'static, BTreeMap<String, String>> {
   // `dotenv::from_path_iter` should eventually be un-deprecated, see:
   // https://github.com/dotenv-rs/dotenv/issues/13
   #![allow(deprecated)]
 
+  if settings.dotenv_load.is_none()
```
casey: This warning concerns implicit loading of `.env` files, as some future version of `just` will switch from loading `.env` files automatically to only on request.

If the user passes either `--dotenv-path` or `--dotenv-filename`, then this warning shouldn't be printed, since they explicitly requested that an environment file be loaded.

src/config.rs
```diff
@@ -29,6 +29,8 @@ pub(crate) struct Config {
   pub(crate) subcommand:           Subcommand,
   pub(crate) unsorted:             bool,
   pub(crate) unstable:             bool,
+  pub(crate) dotenv_filename:      String,
+  pub(crate) dotenv_path:          PathBuf,
```
casey: I think both of these will need to be options, and should be `None` if the flags aren't passed:

```suggestion
  pub(crate) dotenv_filename:      Option<OsString>,
  pub(crate) dotenv_path:          Option<PathBuf>,
```

We'll need to check if the flags were passed explicitly to determine if we should emit the deprecation warning.

src/load_dotenv.rs
```diff
@@ -1,43 +1,57 @@
...
     return Ok(BTreeMap::new());
   }
 
-  for directory in working_directory.ancestors() {
-    let path = directory.join(".env");
+  // look directly for the environment file if specified by path
+  let path = working_directory.join(&config.dotenv_path);
+  if path.is_file() {
+    return load_from_file(config, settings, &path);
+  }
```
casey: This should return an error if the user passed `--dotenv-path` and the file cannot be loaded, which we can do if we just unconditionally load the file when the flag is passed:

```suggestion
  if let Some(dotenv_path) = &config.dotenv_path {
    return load_from_file(config, settings, &dotenv_path);
  }
```

I don't think we need to do `working_directory.join(&config.dotenv_path)`, since file reads are always done relative to the current working directory.

src/load_dotenv.rs
```diff
@@ -1,43 +1,57 @@
 use crate::common::*;
 
-pub(crate) fn load_dotenv(
+fn load_from_file(
```
casey, line 37: Let's move this function after `load_dotenv` in the file, since it's always called afterwards.

casey (commented):
Thanks for this!

> 1. I have these two new args, `--dotenv-filename` and `--dotenv-path` mutually exclusive, with the latter taking precedence in the code over the former. Is that correct?

What do you mean by the latter taking precedence? If they're mutually exclusive, isn't there no precedence either way, since they can't be passed together?

> 2. Currently, if the path from `--dotenv-path` does not resolve to a file, then the existing functionality of searching for a file called ".env" in the working directory or its ancestors is taken. If the type of `Config::dotenv_path` was an `Option<PathBuf>`, I could differentiate between "user set" and "default value", but I didn't see any of the other args as options so I'm hesitant to introduce that without checking with you first. What's the desired behavior here?

Definitely make them `Option`s. Although it's nice to avoid options where possible, for example by populating the config with default values, in this case we need to know what was passed, so it's fine. We should return an error if `--dotenv-path` is passed but we couldn't load an environment file from that path, so we'll need to know if it was passed. Also, both `--dotenv-path` and `--dotenv-filename` should silence the deprecation warning related to implicitly loading `.env` files, so they'll both need to be `Option`s.

> 3. Do you want tests of the overridden dotenv file lookup?

Yes please! 

There should be tests that:

- The options conflict
- if --dotenv-path is passed and file doesn't exist, an error is printed
- Passing --dotenv-path works and loads the right environment file
- Passing --dotenv-filename works and loads the right environment file

Most of the integration tests use the `test!` macro. Recently, I improved the `Test` object, and started writing tests in an imperative style, which is a little less magical and more flexible, for example because you can check what's in the temporary directory afterwards, pass in a temporary directory, and do arbitrary setup. See `dot_justfile_conflicts_with_justfile` for an example of the imperative style. Both ultimately use the `Test` object, so they both test the same things, so feel free to use whichever you prefer.

Celeo:
My most recent battle with the CI job seems to be a success ;)

src/config.rs
```diff
@@ -317,6 +321,19 @@ impl Config {
...
+        Arg::with_name(arg::DOTENV_FILENAME)
+          .long("dotenv-filename")
+          .takes_value(true)
+          .help("Search for environment file named <DOTENV-FILENAME> instead of `.env`")
+          .conflicts_with(arg::DOTENV_PATH),
+      )
+      .arg(
+        Arg::with_name(arg::DOTENV_PATH)
+          .long("dotenv-path")
+          .help("Load environment file at <DOTENV-PATH> instead of searching for `.env` file")
```
```suggestion
          .help("Load environment file at <DOTENV-PATH> instead of searching for one")
```

src/load_dotenv.rs
```diff
@@ -1,45 +1,65 @@
...
-  for directory in working_directory.ancestors() {
-    let path = directory.join(".env");
+  if let Some(path) = &config.dotenv_path {
+    return load_from_file(config, settings, &path);
+  }
+
+  let filename = match &config.dotenv_filename {
+    Some(name) => name,
+    None => DEFAULT_DOTENV_FILENAME,
+  };
```
casey: Can this be something like:

```suggestion
  let filename = config.dotenv_filename.unwrap_or(DEFAULT_DOTENV_FILENAME);
```

Might need to juggle the types a little bit to get it to work.

src/load_dotenv.rs
```diff
@@ -1,45 +1,65 @@
...
+) -> RunResult<'static, BTreeMap<String, String>> {
+  // `dotenv::from_path_iter` should eventually be un-deprecated, see:
+  // https://github.com/dotenv-rs/dotenv/issues/13
+  #![allow(deprecated)]
+
+  if settings.dotenv_load.is_none()
+    && config.verbosity.loud()
+    && config.dotenv_filename.is_none()
+    && config.dotenv_path.is_none()
+    && !std::env::var_os("JUST_SUPPRESS_DOTENV_LOAD_WARNING")
```
casey:
```suggestion
  if config.verbosity.loud()
    && settings.dotenv_load.is_none()
    && config.dotenv_filename.is_none()
    && config.dotenv_path.is_none()
    && !std::env::var_os("JUST_SUPPRESS_DOTENV_LOAD_WARNING")
```

tests/dotenv.rs
```diff
@@ -101,3 +101,46 @@ echo $DOTENV_KEY
     .suppress_dotenv_load_warning(false)
     .run();
 }
+
+#[test]
+fn path_not_found() {
+  Test::new()
+    .justfile("foo:\n\techo $NAME")
```
casey: The argument to `Test::justfile` is passed to `unindent`, so you can write it like this:

```suggestion
    .justfile("
      foo:
        echo $NAME
    ")
```

tests/dotenv.rs
```diff
@@ -101,3 +101,46 @@ echo $DOTENV_KEY
...
+      "error: Failed to load environment file: No such file or directory (os error 2)\n"
+    })
+    .status(EXIT_FAILURE)
+    .run();
+}
+
+#[test]
+fn path_resolves() {
+  Test::new()
+    .justfile("foo:\n\t#!/bin/bash\n\techo $NAME")
```
casey: Can be split onto multiple lines, see above comment.

tests/dotenv.rs
```diff
@@ -101,3 +101,46 @@ echo $DOTENV_KEY
...
+    .args(&["--dotenv-path", "subdir/.env"])
+    .stdout("bar\n")
+    .status(EXIT_SUCCESS)
+    .run();
+}
+
+#[test]
+fn filename_resolves() {
+  Test::new()
+    .justfile("foo:\n\t#!/bin/bash\n\techo $NAME")
```
casey: Can be split onto multiple lines, see above comment.

casey (commented):
Awesome, tests look great. Just a few minor comments.

casey:
Let me know when this is ready for another look, I saw some pushes, but wasn't sure if it's ready for review.

Celeo:
Yup, ready for another review; I think I got all your feedback incorporated and working.

src/load_dotenv.rs
```diff
@@ -1,45 +1,66 @@
...
 pub(crate) fn load_dotenv(
   config: &Config,
   settings: &Settings,
   working_directory: &Path,
 ) -> RunResult<'static, BTreeMap<String, String>> {
-  // `dotenv::from_path_iter` should eventually be un-deprecated, see:
-  // https://github.com/dotenv-rs/dotenv/issues/13
-  #![allow(deprecated)]
-
   if !settings.dotenv_load.unwrap_or(true) {
```
casey: Okay, last thing I swear. I just realized that if the user is passing `--dotenv-path` or `--dotenv-filename`, that should probably override `set dotenv-load := false`.

So this should be something like:

```suggestion
  if !settings.dotenv_load.unwrap_or(true) && config.dotenv_filename.is_none() && config.dotenv_path.is_none() {
```

This should also have tests for both cases, if you have a justfile with `set dotenv-load := false`, passing either flag causes an environment file to be loaded.

casey (commented):
One last thing!

casey:
And def ping the PR with a comment when you're ready for me to take a look at it, I turned off push notifications because they get spammy for some of the repos I watch.

Celeo:
Fixed, and tests added.

> Okay, last thing I swear

No worries!

> push notifications because they get spammy 

Yup, makes total sense!

casey:
I just noticed the use of `/bin/bash`, can the recipes just be normal recipes, i.e. no shebang? I think on at least some systems, `bash` isn't installed as `/bin/bash`, so if it can't be removed, they should be `#!/usr/bin/env bash`.

Celeo:
Ah right, thanks for catching that. I was confused by the presence of output on stderr, but that's the printing of the recipe content. Swapped with the `@` prefix for quieting that, and works. :+1:

casey:
Nice, merged! Thank you very much for the PR!

xorander00:
Love the project, this and go-task are my go-to task runners currently :) I haven't had the chance yet to the read this entire PR yet, but I figured I'd throw my own $0.02 into it here.  
  
The CLI args are nice and I'll likely use them, but personally I think since the potential name and number of .env files can vary between projects, the justfile itself should have a setting to specify the file(s) to load and the order in which to load them.  go-task implements this approach by defining an array in Taskfile.yml that specifies the dotenv files to look for and load from left-to-right (later loaded dotenv files overriding values from previously loaded ones).  That means the justfile would then kind of document the .env files as well, instead of having to figure it out via command line params.

casey:
@xorander00 Definitely agree, this just happened to be implemented first. This has been discussed before, but I added a dedicated issue to track it in #945.

PR #950, opened 2021-08-15
Mention release RSS feed in readme

casey:


README.adoc
```diff
@@ -143,6 +143,10 @@ Example usage:
     just-version: 0.8  # optional semver specification, otherwise latest
 ```
 
+=== Release RSS Feed
+
+An https://en.wikipedia.org/wiki/RSS[RSS feed] of `just` releases is available https://github.com/casey/just/releases.atom[here].
```
chris-morgan, line 148: Wah! This isn’t RSS! This is Atom!

(I want RSS *dead*. It’s technically significantly inferior to Atom, and universally supported outside podcasts. People using the term RSS for the whole class is one of my pet peeves.)

casey: I agree it's annoying, but since so many people call it an "RSS Feed", I'm not sure we could do much better. If we called it an "Release Feed", people might not know what it is,, and if we called it a "Release Atom Feed" definitely nobody would know what it is ;)

PR #969, opened 2021-09-16
Default `positional-arguments` to true

casey:


casey:
Sadly, this breaks Powershell, which complains if it gets extra positional parameters that aren't passed. I'm not sure this is worth it in that case.

PR #992, opened 2021-10-09
Add `--dump-format json`

### LINKED ISSUE #638: Allow introspecting justfile contents
Continuing the discussion in #298. `just --summary` prints a space-separated list of recipe names, for use in shell completion scripts and other tools.

However, this only provides recipe names, and doesn't give arguments, doc comments, variables, or dependency information.

It might be useful to provide a command that returns, in a structured, easy-to-parse format, all available information about a Justfile.

JSON is probably appropriate, since it doesn't need to be human readable or editable, and JSON is probably the most widely supported format.

The output could contain a dictionary with all information about a parsed justfile, including:

- [ ] Recipe names
- [ ] Parameter names
- [ ] Parameter types (variadic or not)
- [ ] Parameter defaults
- [ ] Recipe bodies, as a list of lines
- [ ] Entire recipes, parameters and dependencies, for display to the user
- [ ] Assignment names
- [ ] Assignment expressions
- [ ] Entire assignments, name and expression, for display to the user
- [ ] Alias names
- [ ] Alias targets

I'm inclined to wait to add this until someone needs it, so I can work with them to make sure the format and contents meets their needs. However, this isn't hard to add at all, so don't be shy!

I'd probably start with recipe names, and gradually add more information in a backwards-compatible way, as it's needed.

### LINKED ISSUE #989: Machine readable metadata from justfile?
I'm exploring showing visual depictions of dependencies as a means of automated documentation.

Idea:

just -f justfile --metadata | `<custom script>` | mermaid live image/link

Where the link goes to e.g. https://mermaid.ink/svg/eyJjb2RlIjoiZ3JhcGggVERcbkFbQ2hyaXN0bWFzXSAtLT58R2V0IG1vbmV5fCBCKEdvIHNob3BwaW5nKVxuQiAtLT4gQ3tMZXQgbWUgdGhpbmt9XG5DIC0tPnxPbmV8IERbTGFwdG9wXVxuQyAtLT58VHdvfCBFW2lQaG9uZV1cbkMgLS0-fFRocmVlfCBGW2ZhOmZhLWNhciBDYXJdXG4gICIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9

(the graph content is baked into the URL and is a simple DSL)

What I would need is a machine readable (JSON preferably) description of a justfile, something like:

```json
{
    "commands": [
        {
            "command": "do_stuff",
            "docstring": "this is the docstring"
        },
        {
            "command": "do_stuff_but_more",
            "docstring": "this is the docstring",
            "dependencies": [
                {
                    "command": "do_stuff",
                    "args": []
                }
            ]
        }
    ]
}
```

That would open the door to other downstream/upstream tools being able to automatically understand `justfiles`.

Happy to elaborate more if this doesn't make sense. Basically, allow outputting the data structure right after parsing.

casey:
Closes #989.

To Do
- [x] Document
- [x] Use `--format=json` instead of `--json` subcommand
- [x] Mark as `--unstable`
- [ ] Consider using a less weird format for expressions and recipe bodies


casey:
@dionjwa Have a look at this PR, in particular the tests in `tests/json.rs`, which show the JSON output.

casey:
The feature is unstable for now, which means that you have to use the `--unstable` flag. I did this because I'm a bit unsure of the exact details of the JSON format, and might want to change it based on feedback.

In particular, expressions are serialized in a lispy way which is easy to read, as opposed to the standard rust way. Also, recipes bodies are lists of lists containing either strings, or *another* list, containing an expression.

so:

```
foo:
  bar
  {{ baz }}
```

has:
```
...
"body":
  [
    // first line, contains recipe text
    ["bar"],
    // second line, contains an interpolation, represented as a list
    // the contents of the interpolation is an expression, in this case `["variable", "baz"]`,
    // which is the same representation of expressions used elsewhere
    [[["variable", "baz"]]],
  ]
...
```

casey:
@dionjwa Ping! Let me know if this format is what you were looking for, if you get the chance.

dionjwa:
Feedback:

Maybe `--dump` automatically defaults to `--dump=json`, allowing future formats (unlikely, JSON is too ubiquitous, but?) if/when needed (`--dump=yaml` or `--dump yaml`).

As long as the recipe data contains the dependencies, I'm good. The fact that you're actually putting the recipe content there as expressions is just extra awesomeness. I don't need that part right now, but I can easily imagine wanting that in the future since it allows introspection.

Thank you so much!

casey:
> Maybe `--dump` automatically defaults to `--dump=json`, allowing future formats (unlikely, JSON is too ubiquitous, but?) if/when needed (`--dump=yaml` or `--dump yaml`).

`--dump` is already used to dump the text contents of a `justfile`, so making it default to `--dump` wouldn't be backwards compatible.

> As long as the recipe data contains the dependencies, I'm good. The fact that you're actually putting the recipe content there as expressions is just extra awesomeness. I don't need that part right now, but I can easily imagine wanting that in the future since it allows introspection.

Sweet, good to know! What language and library will you use to deserialize the JSON?

> Thank you so much!

You bet! This has been a long time coming, so I'm glad to have a user driving it forward.

casey:
@dionjwa Merged! Let me know how it works for you. Also, if you need me to cut a release so you can experiment with it, let me know.

dionjwa:
That would be great, but also no rush 👍🏻

casey:
@dionjwa I released 0.10.4, which includes `--dump-format json`. Hope it works for you!

dionjwa:
You're wonderful, thank you! I'll let you know when I finish this crazy esoteric use case I have in mind

casey:
You bet! Looking forward to it :)

Issue #851, opened 2021-06-02
Test that project justfile is always formatted


casey:
Run `just --fmt` and complain if it makes a change. Also add `--check` to `--fmt` so that it returns a nonzero error code if there would be a diff.

starthal:
Closed by #1001?

casey:
Nice, thanks for noticing!

PR #1001, opened 2021-10-14
fmt: check formatting only in dry-run

hdhoang:
Let's have a try for the 2nd part of #851. Instead of adding a `--check` flag, I reused `--dry-run` here as a PoC. However, `dry-run` conflicts with `-q` flag, thus the "well-formed" output always appear.

```sh
$ cargo -q run -- --unstable --dry-run --fmt
error: Justfile `B:\code\just\justfile` is not well-formatted.
$ echo $?
1

$ cargo -q run -- --unstable --fmt
Wrote justfile to `B:\code\just\justfile`
 $ cargo -q run -- --unstable --dry-run --fmt
Justfile `B:\code\just\justfile` is well-formatted.
```

For the next step, I don't have a good direction to do pretty diffing of the AST & file content yet.

Hope this helps. Please label "hacktoberfest"+"hacktoberfest-accepted" if you don't mind, thanks!


src/subcommand.rs
```diff
@@ -258,7 +258,31 @@ impl Subcommand {
...
+      let mut file = File::open(&search.justfile).map_err(|e| Error::Load {
+        path: search.justfile.clone(),
+        io_error: e,
+      })?;
+      file
+        .read_to_string(&mut current_content)
+        .map_err(|e| Error::Load {
+          path: search.justfile.clone(),
+          io_error: e,
+        })?;
```
casey: The `src` can be passed in from outside, so we can avoid loading the file twice.

src/subcommand.rs
```diff
@@ -258,7 +258,31 @@ impl Subcommand {
...
+
+      if current_content != format!("{}", ast) {
```
casey:
```suggestion
      if src != ast.to_string() {
```

src/subcommand.rs
```diff
@@ -258,7 +258,31 @@ impl Subcommand {
...
+        Err(Error::NotWellFormatted {
+          justfile: search.justfile.clone(),
+        })
+      } else {
+        if config.verbosity.loud() {
+          eprintln!("Justfile `{}` is well-formatted.", search.justfile.display());
```
casey:
```suggestion
          eprintln!("No diffs found.", search.justfile.display());
```

src/subcommand.rs
```diff
@@ -258,7 +258,31 @@ impl Subcommand {
...
+      })?;
+      file
+        .read_to_string(&mut current_content)
+        .map_err(|e| Error::Load {
+          path: search.justfile.clone(),
+          io_error: e,
+        })?;
+
+      if current_content != format!("{}", ast) {
+        Err(Error::NotWellFormatted {
```
casey: I think we probably shouldn't call justfiles that produce diffs "not well formatted". The just have some different formatting which is not necessarily bad, just nonstandard.

```suggestion
        Err(Error::FormatCheckFoundDiff {
```

src/error.rs
```diff
@@ -95,6 +95,9 @@ pub(crate) enum Error<'src> {
   },
   NoChoosableRecipes,
   NoRecipes,
+  NotWellFormatted {
+    justfile: PathBuf,
+  },
```
```suggestion
  FormatCheckFoundDiff,
```

casey: I don't think we need to mention the path.

src/error.rs
```diff
@@ -510,6 +513,13 @@ impl<'src> ColorDisplay for Error<'src> {
       NoRecipes => {
         write!(f, "Justfile contains no recipes.")?;
       }
+      NotWellFormatted { justfile } => {
+        write!(
+          f,
+          "Justfile `{}` is not well-formatted.",
+          justfile.display()
+        )?;
+      }
```
```suggestion
      FormatCheckFoundDiff => {
        write!(
          f,
          "Diff found in justfile.",
        )?;
      }
```

casey (commented):
It should be okay just to merge this, it will be marked as accepted because it's labeled `hacktoberfest`.

Some comments. For diffing, the crate https://crates.io/crates/prettydiff looks promising.

src/subcommand.rs
hdhoang: neat! applied, thanks

src/subcommand.rs
hdhoang: Good wording, applied

src/error.rs
hdhoang: Okay, the maintainers should be aware of their intended Justfile already

hdhoang:
`prettydiff` looks okay enough! I hope you're okay with scoped `use`, and the by-word mode.

Ah sorry to bother again, but please add label "hacktoberfest-accepted" on PRs. The rules for this year distinguish "hacktoberfest" topic on projects, vs "hacktoberfest-accepted" I think.
image

We appreciate your patience about this.

src/error.rs
```diff
@@ -458,6 +459,9 @@ impl<'src> ColorDisplay for Error<'src> {
           write!(f, "\n{}", suggestion)?;
         }
       }
+      FormatCheckFoundDiff => {
+        write!(f, "Format check found some difference.")?;
```
casey: How about:

```suggestion
        write!(f, "Formatted justfile differs from original.")?;
```

src/subcommand.rs
```diff
@@ -255,10 +255,22 @@ impl Subcommand {
     Ok(())
   }
 
-  fn format(config: &Config, ast: Ast, search: &Search) -> Result<(), Error<'static>> {
+  fn format(config: &Config, ast: Ast, src: &str, search: &Search) -> Result<(), Error<'static>> {
     config.require_unstable("The `--fmt` command is currently unstable.")?;
 
-    if let Err(io_error) =
+    if config.dry_run {
+      let new_src = ast.to_string();
```
casey:
```suggestion
      let src_formatted = ast.to_string();
```

src/subcommand.rs
```diff
@@ -255,10 +255,22 @@ impl Subcommand {
...
+      if src != new_src {
```
casey:
```suggestion
      if src != src_formatted {
```

src/subcommand.rs
```diff
@@ -255,10 +255,22 @@ impl Subcommand {
...
+        use prettydiff::diff_words;
+        println!("Formatting diff:\n\n{}", diff_words(&src, &new_src));
```
casey: Since it's only used once, it's fine to not have a use statement at all:

```suggestion
        println!("Formatting diff:\n\n{}", prettydiff::diff_words(&src, &new_src));
```

src/subcommand.rs
```diff
@@ -255,10 +255,22 @@ impl Subcommand {
...
+        Err(Error::FormatCheckFoundDiff)
+      } else {
+        if config.verbosity.loud() {
+          eprintln!("No format difference found.");
```
casey: I think it's actually okay to not print anything in case of success. It's very common to use `rust fmt -- --check` and 
only check for the exit code.

```suggestion
```

casey (commented):
Some random comments. This definitely needs tests in `test/fmt.rs`:

- Check that if there's no diff, the command returns success
- Check that if there is a diff, the command returns failure, and that the diff looks correct

casey:
> Ah sorry to bother again, but please add label "hacktoberfest-accepted" on PRs. The rules for this year distinguish "hacktoberfest" topic on projects, vs "hacktoberfest-accepted" I think. image

Done! Sorry, I must have been looking at outdated documentation.

> We appreciate your patience about this.

No problem at all!

src/subcommand.rs
hdhoang: Done (but the new `similar` crate uses more symbols)

src/subcommand.rs
hdhoang: I pulled out `src_formatted` to outer scope too

src/error.rs
hdhoang: Done

hdhoang:
I have incorporated your comments.

I also replaced `prettydiff` with `similar`, because it has 0-dep, and battle-tested in pijul. Plain output is also easier to copy into test macro. Though I still don't understand the double `\n\n` in expected stdout

tests/fmt.rs
```diff
@@ -11,6 +11,28 @@ test! {
   status: EXIT_FAILURE,
 }
 
+test! {
+  name: dry_run_ok,
+  justfile: "",
+  args: ("--unstable", "--fmt", "--dry-run"),
```
casey: I think this should be a more complicated justfile that has no diff, so at least one assignment and one recipe, with some text in the body.

casey (commented):
> Though I still don't understand the double \n\n in expected stdout

I think this is because you're using `println`, which will also print a newline, and the diff itself also has a newline. Try `print`, which doesn't add an extra newline.

tests/fmt.rs
hdhoang: True, done

hdhoang:
hi, i hope this finds you well. I have updated the test's justfile

casey:
Sounds good. I think the next step is probably to replace `--dry-run` with `--check`.

hdhoang:
I see the value in having `--check` for fmt, similar to other tools.

I made 2 attempts, the tip is as you described with `--fmt --check`, but `--fmt --dry-run` runs formatting. A previous state added `--check` as an alias for `config.dry_run` field. Do you think the second method is more friendly?

I suppose an `-s, --simulate` like apt & youtube-dl would be another similar flag to `dry-run`, if you think aliases are useful.

casey:
I think that `just --fmt` should format, and `just --check` should check the formatting. I think that `just --fmt --dry-run` is probably unnecessary, since it's more typing that `just --fmt`.

hdhoang:
I have made it into a stable subcommand. If that's alright, I will update completions files and docs too.

casey:
> I have made it into a stable subcommand. If that's alright, I will update completions files and docs too.

Is there an advantage to making it into its own subcommand? I was thinking that for consistency with `cargo fmt`, which is `cargo fmt` to format and `cargo fmt --check` to check, it should be a flag. Not sure what other tools do, but I assume it's similar.

hdhoang:
As a subcommand, it can be shortest `just --check`. I agree that for consistency it should be `just --fmt --check`. I will make it that way, and also allow `--check` without `unstable` because checking is safe.

casey:
> As a subcommand, it can be shortest `just --check`.

I think `just --check` is confusing, since `--check` doesn't make it clear that's checking formatting. I think only making it available as `just --fmt --check` is the way to go.

>  also allow `--check` without `unstable` because checking is safe.

`--check` should still be unstable, because we haven't figured out stability guarantees for formatting. Someone might use `just --fmt --check` on CI, and if the formatting changes, their CI will break, so we shouldn't make `just --fmt --check` stable before we're pretty sure that the formatting doesn't change.

hdhoang:
Hi, I have updated the behaviour & squashed the commits. Thanks for the patience

src/config.rs
```diff
@@ -81,6 +82,7 @@ mod cmd {
 
 mod arg {
   pub(crate) const ARGUMENTS: &str = "ARGUMENTS";
+  pub(crate) const CHECK_FMT: &str = "CHECK";
```
casey: The const name should match the value, so:
```suggestion
  pub(crate) const CHECK: &str = "CHECK";
```

src/config.rs
```diff
@@ -116,6 +118,12 @@ impl Config {
       .version_message("Print version information")
       .setting(AppSettings::ColoredHelp)
       .setting(AppSettings::TrailingVarArg)
+      .arg(
+        Arg::with_name(arg::CHECK_FMT)
```
```suggestion
        Arg::with_name(arg::CHECK)
```

src/config.rs
```diff
@@ -116,6 +118,12 @@ impl Config {
...
+          .long("check")
+          .requires(cmd::FORMAT)
+          .help("Make `--fmt` check formatting only. Exits with 1 and prints diff if formatting is required."),
```
casey: I stole this from `rustfmt`'s help:

```suggestion
          .help("Run `--fmt` in 'check' mode. Exits with 0 if justfile is formatted correctly. Exits with 1 and prints a diff if formatting is required."),
```

src/config.rs
```diff
@@ -532,6 +540,7 @@ impl Config {
       || matches.occurrences_of(arg::SHELL_ARG) > 0;
 
     Ok(Self {
+      check_fmt: matches.is_present(arg::CHECK_FMT),
```
```suggestion
      check: matches.is_present(arg::CHECK),
```

src/subcommand.rs
```diff
@@ -255,11 +255,22 @@ impl Subcommand {
...
   }
 
-  fn format(config: &Config, ast: Ast, search: &Search) -> Result<(), Error<'static>> {
+  fn format(config: &Config, ast: Ast, src: &str, search: &Search) -> Result<(), Error<'static>> {
     config.require_unstable("The `--fmt` command is currently unstable.")?;
 
-    if let Err(io_error) =
-      File::create(&search.justfile).and_then(|mut file| write!(file, "{}", ast))
+    let src_formatted = ast.to_string();
+    if config.check_fmt {
```
casey:
```suggestion
    if config.check {
```

casey (commented):
Nice, some minor comments, but overall looks good. You'll need to run `just generate-completions` to make the tests pass. They're complaining that the checked-in completions aren't up-to-date.

tests/fmt.rs
```diff
@@ -11,6 +11,51 @@ test! {
   status: EXIT_FAILURE,
 }
 
+test! {
+  name: check_fmt_without_fmt,
```
hdhoang:
```suggestion
  name: check_without_fmt,
```

hdhoang:
I have made the shell completions. However I don't have `help2man` working, please regen it after merg, thanks

casey:
Nice! The man page isn't checked in, so no need to generate it. Thanks for sticking with this!

Issue #412, opened 2019-04-17
Improve shell not found error message


casey:
The current error message is sadly lacking:

```
: just --shell foo
error: Backtick could not be run because just could not find `sh`:
No such file or directory (os error 2)   |
39 | version = `sed -En 's/version[[:space:]]*=[[:space:]]*"([^"]+)"/v\1/p' Cargo.toml | head -1`
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The same error is also printed when a backtick fails with a not-found error, like ```bar := `not-a-command````. This should be improved as well.

Swordelf2:
The errors seem to be fine at the moment, don't they?

casey:
The problem with the error is that it says "just could not find `sh`", when really the problem is that just couldn't find the custom shell the user set, in this case `foo`, so it's a bit misleading.

hdhoang:
Hi, do you think a simple substitution to platform-neutral "subcommand" works? Like so https://github.com/hdhoang/just/commit/2fd4c85ee60a6bdcd496b67d04866dc032387d91

My environment is native windows + nushell, but I also have scoop+busybox on PATH. This complicated the stderr regex much. I hope the tests pass for other configurations, or we can remove them.

To get the literal invoked command, I think Error variants needed more fields. The most relevant filed I can see now is `token.src`. But it carries much more than command word.

casey:
I think that "subcommand" is probably confusing, since that usually refers to subcommands of multi-command programs, like `cargo run` and `cargo build`. "the shell" is probably better, i.e. "Recipe `{}` could not be run because just could not run the shell:".

saona-raimundo:
I am sorry but I do not understand this error.
I am running 
```
just
```
on Windows and getting back
```
error: Recipe `XXX` could not be run because just could not find the shell: program not found
```

I believe I do not have a shell installed in Windows? (for some reason)

[The documentation on shell](https://just.systems/man/en/chapter_58.html#configuring-the-shell) does not give an example on how to set the shell to `cmd` for example. I tried `--shell cmd` but this does not work.

Thanks in advance!

starthal:
Is further work needed here?

casey:
@starthal Nice catch! I guess someone fixed this a while ago and the issue was never closed. I added tests for the error message in #2145.

PR #1002, opened 2021-10-16
error: reword `sh` to platform-neutral "shell command"

hdhoang:
A simple workaround for #412, to avoid confusion on Windows without `sh`. It's a hint to look up `shell` variable, instead of trying to find a UNIX `sh` implementation.

There's no existing test with this text, and I tried making some with `not-a-command` in `shell`, `x:=\`\``, and recipe. But they are platform-specific too, and changes depending on whether I have busybox utils in PATH or not. So I have left them out for this PR.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -253,17 +253,17 @@ impl<'src> ColorDisplay for Error<'src> {
           match io_error.kind() {
             io::ErrorKind::NotFound => write!(
               f,
-              "Backtick could not be run because just could not find `sh`:\n{}",
+              "Backtick could not be run because just could not find the shell command:\n{}",
               io_error
             ),
             io::ErrorKind::PermissionDenied => write!(
               f,
-              "Backtick could not be run because just could not run `sh`:\n{}",
+              "Backtick could not be run because just could not run the shell command:\n{}",
               io_error
             ),
             _ => write!(
               f,
-              "Backtick could not be run because of an IO error while launching `sh`:\n{}",
+              "Backtick could not be run because of an IO error while launching the shell command:\n{}",
               io_error
             ),
           }?;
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -253,17 +253,17 @@ impl<'src> ColorDisplay for Error<'src> {
           match io_error.kind() {
             io::ErrorKind::NotFound => write!(
               f,
-              "Backtick could not be run because just could not find `sh`:\n{}",
+              "Backtick could not be run because just could not find the shell command:\n{}",
```
casey: I think maybe just "the shell" is better.

```suggestion
              "Backtick could not be run because just could not find the shell:\n{}",
```

src/error.rs
```diff
@@ -253,17 +253,17 @@ impl<'src> ColorDisplay for Error<'src> {
...
               io_error
             ),
             io::ErrorKind::PermissionDenied => write!(
               f,
-              "Backtick could not be run because just could not run `sh`:\n{}",
+              "Backtick could not be run because just could not run the shell command:\n{}",
```
```suggestion
              "Backtick could not be run because just could not run the shell:\n{}",
```

src/error.rs
```diff
@@ -253,17 +253,17 @@ impl<'src> ColorDisplay for Error<'src> {
...
               io_error
             ),
             _ => write!(
               f,
-              "Backtick could not be run because of an IO error while launching `sh`:\n{}",
+              "Backtick could not be run because of an IO error while launching the shell command:\n{}",
```
```suggestion
              "Backtick could not be run because of an IO error while launching the shell:\n{}",
```

src/error.rs
```diff
@@ -481,17 +481,17 @@ impl<'src> ColorDisplay for Error<'src> {
         match io_error.kind() {
           io::ErrorKind::NotFound => write!(
             f,
-            "Recipe `{}` could not be run because just could not find `sh`: {}",
+            "Recipe `{}` could not be run because just could not find the shell command: {}",
```
```suggestion
            "Recipe `{}` could not be run because just could not find the shell: {}",
```

src/error.rs
```diff
@@ -481,17 +481,17 @@ impl<'src> ColorDisplay for Error<'src> {
...
             recipe, io_error
           ),
           io::ErrorKind::PermissionDenied => write!(
             f,
-            "Recipe `{}` could not be run because just could not run `sh`: {}",
+            "Recipe `{}` could not be run because just could not run the shell command: {}",
```
```suggestion
            "Recipe `{}` could not be run because just could not run the shell: {}",
```

src/error.rs
```diff
@@ -481,17 +481,17 @@ impl<'src> ColorDisplay for Error<'src> {
...
             recipe, io_error
           ),
           _ => write!(
             f,
-            "Recipe `{}` could not be run because of an IO error while launching `sh`: {}",
+            "Recipe `{}` could not be run because of an IO error while launching the shell command: {}",
```
```suggestion
            "Recipe `{}` could not be run because of an IO error while launching the shell: {}",
```

casey (commented):
I think this is okay without tests. I feel like usually, tests that just check the contents of string literals aren't super useful, since they will just reproduce the string, and aren't testing complex behavior.

hdhoang:
Done, Now you have done all the work :)

casey:
Merging! This is still a huge improvement, and wouldn't have happened without your initiative.

PR #1007, opened 2021-10-27
Fix readme `make`'s not correctly displayed

peter50216:
Code blocks displayed on GitHub .adoc doesn't work well when immediately followed by a `'`, and the code block is incorrectly spanned to the next backtick instead.

Fix this by using [unconstrained quote](https://github.com/asciidoctor/asciidoctor.org/issues/64) ` `` ` instead.


casey:
Nice, thank you very much!

PR #1027, opened 2021-11-15
Use ps1 extension for pwsh shebangs

dmringo:
As of PowerShell/PowerShell#15859, which is in PowerShell Core 7.2, `pwsh.exe` requires a `ps1` extension for `-File` arguments on Windows. This "works on my (Win 10) machine" but I know very little about PowerShell and even less about Rust. Not sure if it's worth trying to make Just's behavior here Windows-specific.

### DIFF
--- a/src/shebang.rs
+++ b/src/shebang.rs
@@ -41,7 +41,7 @@ impl<'line> Shebang<'line> {
   pub(crate) fn script_filename(&self, recipe: &str) -> String {
     match self.interpreter_filename() {
       "cmd" | "cmd.exe" => format!("{}.bat", recipe),
-      "powershell" | "powershell.exe" => format!("{}.ps1", recipe),
+      "powershell" | "powershell.exe" | "pwsh" | "pwsh.exe" => format!("{}.ps1", recipe),
       _ => recipe.to_owned(),
     }
   }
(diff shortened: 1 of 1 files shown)


casey:
Looks good to me! Could you add a test? No worries if not, I can add it. It should be really easy, just copy the `powershell_script_filename` test in the same file twice, give them new names, and test the cases you added.

dmringo:
> 
> 
> Looks good to me! Could you add a test? No worries if not, I can add it. It should be really easy, just copy the `powershell_script_filename` test in the same file twice, give them new names, and test the cases you added.

Done!

I was tempted to join the powershell filename tests together (like the various `check` calls in `split_shebang`) but maybe it's better to keep them separate to have well-named tests. Is the `#test` annotation something you can abstract over in Rust?

Thanks for all your work on this project. `just` is near the top of my most used (and loved) CLI tools.

casey:
> Done!

Awesome!

> I was tempted to join the powershell filename tests together (like the various `check` calls in `split_shebang`) but maybe it's better to keep them separate to have well-named tests. Is the `#test` annotation something you can abstract over in Rust?

I think separate well-named tests is good. You can abstract over the test annotation using macros, but it winds up being a little gross. I'm actually moving away from having macros that generate tests, the tests in the `tests` directory are mixed, some use macros, some don't.

> Thanks for all your work on this project. `just` is near the top of my most used (and loved) CLI tools.

You bet! That's great to hear.

PR #1030, opened 2021-11-20
Update README.adoc

soenkehahn:
Before this PR the monospace formatting would extend until `.PHONY`. I'm not sure exactly why, [this](https://docs.asciidoctor.org/asciidoc/latest/text/monospace/) documentation sounds like the version on master should work. In any case, switching to [unconstrained](https://docs.asciidoctor.org/asciidoc/latest/text/monospace/#unconstrained) monospace seems to fix it on github.


casey:
Nice, thank you!

PR #1035, opened 2021-11-25
Make completions work with an alias

kurtbuilds:
Right now if you use the README recommended alias, `alias j=just`, in bash, it stops shell completion from working.

This is a small PR to allow the bash completions script to work with an alias, and adds instructions in the README for how to use it.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1750,6 +1750,12 @@ Tools that pair nicely with `just` include:
 
 For lightning-fast command running, put `alias j=just` in your shell's configuration file.
 
+The aliased command may not keep the shell completion functionality described in the next section. In Bash, you can use the same completion function as `just` for your aliased command.
+
+```sh
+complete -F _just -o bashdefault -o default j
+```
+
 === Shell Completion Scripts
 
 Shell completion scripts for Bash, Zsh, Fish, PowerShell, and Elvish are available in the link:completions[] directory. Please refer to your shell's documentation for how to install them.
--- a/README.adoc
+++ b/README.adoc
@@ -1750,7 +1750,7 @@ Tools that pair nicely with `just` include:
 
 For lightning-fast command running, put `alias j=just` in your shell's configuration file.
 
-The aliased command may not keep the shell completion functionality described in the next section. In Bash, you can use the same completion function as `just` for your aliased command.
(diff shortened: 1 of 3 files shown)


casey:
Thanks for the PR! The completions are actually generated by `clap`, the command-line parsing library that `just` uses. I think this should actually be submitted as a PR to the [upstream project](https://github.com/clap-rs/clap).

For now though, the changes need to be added to the `BASH_COMPLETION_REPLACEMENTS` [here](https://github.com/casey/just/blob/a4bd1cb67cea252f9d77ad9c8ad9f0ddfb6e0926/src/completions.rs#L159). It takes the form of a slice of tuples, with the first element being the part to look for, and the second being the part to replace.

kurtbuilds:
Thank you for the comments! I fixed this PR and added an upstream PR to clap to fix it for other projects.

Note that `just` still relies on `clap` version 2.33.3, and I submitted the PR to the master branch, which has version 3.0.0-beta5, so after that PR merges, this PR still should remain until `just` updates `clap` to ^3.

I confirmed that the checked-in completions.bash file does not change when re-running completions with the updated completion function included in this PR.

README.adoc
```diff
@@ -1750,6 +1750,12 @@ Tools that pair nicely with `just` include:
 
 For lightning-fast command running, put `alias j=just` in your shell's configuration file.
 
+The aliased command may not keep the shell completion functionality described in the next section. In Bash, you can use the same completion function as `just` for your aliased command.
```
casey: This works in `zsh`, so I'm thinking this might be a bash specific issue, so it would be useful to call out Bash early on.

```suggestion
In `bash`, the aliased command may not keep the shell completion functionality described in the next section. Add the following line to your `.bashrc` to use the same completion function as `just` for your aliased command:
```

casey:
Nice, one comment, but otherwise looks great.

kurtbuilds:
@casey Wanted to make sure this doesn't get lost in the ether - anything blocking merging this in?

casey:
Thanks for pinging this issue! I didn't get a notification when you made the change to this branch, but it looks good to go. I hit the auto-merge button so it should merge when the tests pass.

casey:
The tests failed, it looks like because there's a formatting diff.

kurtbuilds:
Oops, thanks. I ran tests but not formatting. Fixed. It looks like me pushing that caused the auto-merge to reset.

Thank you!

PR #1043, opened 2021-12-05
Release 0.10.5

casey:
- Bump version: 0.10.4 → 0.10.5
- Update changelog
- Update config test
- Update dependencies
- Update man page


page-down:
I noticed in `man/just.1` the date has changed, but the version number is still 0.10.4 .
Maybe you want to update this version number in the next update.

casey:
Ahh, good catch, I must have regenerated the man page before I changed the version number. It should be fixed in the next update.

Issue #1050, opened 2021-12-18
Cross-platform justfiles & default shell


michidk:
I know that there is already #531 as well as #3 but I am more looking for a workaround to have this right now.

What I want to achieve is something like this:
```make
[conditional(os_family = "windows")]
set shell := ["powershell"]
```

Because 
```make
set shell := ["powershell"]
```
makes the file unusable for linux users that work on the same project.

I have hoped that one could use conditionals for that:
```make
shell := if os() == "windows" { "powershell" } else { "bash" }
```
But then I get the error: `...  because just could not find the shell: The system cannot find the file specified. (os error 2)`
Putting the `set` keyword at the beginning gives `Expected '[', but found identifier`

But even if that would work, I don't want to default to `bash` for Linux, so it would be nice if something like this was possible:
```make
shell := if os() == "windows" { "powershell" } else { shell }
```

I also tried the `--shell` flag with various arguments ("powershell", "powershell.exe", "["powershell"]") but couldn't get it to work.
This solution would be okay, since I then could set an alias on my system to always include that flag.

It would be even better if the default shell on Windows would just be "powershell".

casey:
> It would be even better if the default shell on Windows would just be "powershell".

I think this makes a lot of sense.

Very early on, the way I got `just` working on Windows was to install cygwin, a unix userland (`sh` and utilities, like `cp`, `rm`, etc) for Windows. This worked well, and a lot of justfiles were sort of cross-platform, but required installing cygwin. Cygwin is pretty weird, and pretty esoteric. Most Windows users these days use WSL if they want a unix command line, and use powershell when they're in native Windows.

I think it makes sense to default to Powershell if it's the native Windows binary:

- Powershell *is* the system shell on Windows. For a windows user, it should be the default
- Powershell is installed on all modern Windows machines
- cmd.exe is unpopular and weird, so that's not a good option
- Powershell doesn't require any dependencies

Changing the default, however, would break a lot of existing justfiles. So, what I propose as a first step is to add a setting that changes the shell to powershell, but only if the os is windows.

Something like:

```
# use powershell on Windows instead of `sh`
set windows-powershell := true
```

This would allow writing cross-platform-ish Justfiles, as long as you stick to the shared subset of `powershell` and `sh`, and would also allow Windows users to get a good experience that doesn't require installing any dependencies.

If this was popular, we could then figure out a way to phase it in as a default, or at least make it more convenient, perhaps with something like Rust editions.

casey:
[...]

PR #1055, opened 2021-12-24
Add environment variables for shell and shell-arg

michidk:
shell and shell arguments can now be defined system-wide using env vars. Discussion in #1050

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -218,6 +218,7 @@ impl Config {
           .long("shell")
           .takes_value(true)
           .default_value(DEFAULT_SHELL)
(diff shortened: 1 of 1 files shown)


michidk:
Im really confused on why this unit test still fails

casey:
Can you add a test for this?

casey:
I've been thinking about this, and I don't actually think it's a good idea. Pulling the shell from environment variables will break recursive invocations of just which expect the default shell.

casey:
Sorry for not thinking of this sooner!

PR #1057, opened 2021-12-25
Implement windows-powershell setting

### LINKED ISSUE #1050: Cross-platform justfiles & default shell

michidk:
Implement the windows-powershell setting to fix #1050.

To be able to run the same justfile on Linux and Windows `set shell` is not enough because it would change the shell on every platform. To set PowerShell as the default shell when running `just` on Windows set `windows-powershell` to true.
```make
set windows-powershell
hello-windows:
  Write-Host "Hello, world!"
hello-linux:
  echo 'Hello, world!'
```


michidk:
Hey @casey I finished the implementation of `windows-powershell`. The core logic is here: https://github.com/michidk/just/blob/windows-powershell/src/settings.rs#L41
But quite a few integration tests like ` byte_order_mark::ignore_leading_byte_order_mark` started failing after I implemented the shell selection logic (see GitHub actions).
I am not entirely sure what those tests do and what my code has do to with them failing, but I noticed in the test results that `bash` is used sometimes. Not sure if that's intended or some fallback in Linux?
If you have some time over the course of the next weeks, maybe you can have a look at it.

Happy new year!

casey:
It's not a very satisfying answer, but I think if a whole bunch of tests start failing, the best thing to do is to uncomment/revert code until you they all pass/you get the previous behavior, and then start modifying it incrementally until you get failing tests.

casey:
#1061 reminded me of something: If there are arguments that make powershell less error prone, we should pass them by default. So, for example, if there's something like `-u`, which makes evaluating undefined variables an error, we should pass that.

michidk:
> #1061 reminded me of something: If there are arguments that make powershell less error prone, we should pass them by default. So, for example, if there's something like `-u`, which makes evaluating undefined variables an error, we should pass that.

I don't think there is such an argument. I think I will add the `-NoLogo` argument though. The argument list can be found here: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe?view=powershell-5.1

michidk:
@casey I changed the `version` just-variable into an environment variable in the main justfile of this project itself because it would try to execute `sed` which is not available on Windows. Now I am able to execute most commands like `just fmt`.
Not sure if you agree with that change though, so please have a look at it and let me know what you think: https://github.com/michidk/just/commit/1dc2502d980e93abcae84f2bdc42ebe295080c22

casey:
I don't think the main justfile should be changed in this PR. In general, I try to keep PRs as small as possible, so they can be reviewed easily and merged quickly. Changing the main `justfile` to use Powershell would be a large change, since all the recipes would need to be rewritten to be compatible with Powershell, so it should be left out of this PR.

michidk:
> I don't think the main justfile should be changed in this PR. In general, I try to keep PRs as small as possible, so they can be reviewed easily and merged quickly. Changing the main `justfile` to use Powershell would be a large change, since all the recipes would need to be rewritten to be compatible with Powershell, so it should be left out of this PR.

I actually used this justfile for quite some time now. It works great - I never needed the parts that didn't work with PowerShell. But then I will keep it locally because I don't have the time to translate the whole justfile. Just as a sidenote: If variables would evaluate lazy (as soon as they are needed, and not before that) then the file in the current master would also work for most commands like `just fmt`.

michidk:
I reverted the previous commit, so that the justfile now matches with the one from the current master again (meaning it's unchanged in this PR).

I would consider the PR as ready for review now. Thanks to @Shemnei who worked with me on this feature.

casey:
Nice! I just wanted to comment, since I haven't been able to get to this as quickly as I hoped, but I should be able to take a look soon.

README.adoc
```diff
@@ -656,6 +657,21 @@ foo:
 
 `just` passes the command to be executed as an argument. Many shells will need an additional flag, often `-c`, to make them evaluate the first argument.
 
+==== Windows PowerShell
+
+To be able to run the same justfile on Linux and Windows `set shell` is not enough because it would change the shell on every platform. To set PowerShell as the default shell when running `just` on Windows set `windows-powershell` to true.
```
casey: I think I would phrase this a bit differently, maybe:

```suggestion
`just` uses `sh` on Windows by default. To use PowerShell instead, set `windows-powershell` to true.
```

README.adoc
```diff
@@ -656,6 +657,21 @@ foo:
...
+
+```make
+set windows-powershell
+
+hello-windows:
+  Write-Host "Hello, world!"
+
+hello-linux:
+  echo 'Hello, world!'
+```
```
casey: I think this example is confusing, because it shows a justfile with a `hello-windows` recipe that won't work on Linux. I would remove it.

```suggestion
```

src/config.rs
```diff
@@ -558,26 +554,27 @@ impl Config {
       }
     };
 
-    let shell_args = if matches.is_present(arg::CLEAR_SHELL_ARGS) {
-      Vec::new()
-    } else {
+    let shell_args = if matches.is_present(arg::SHELL_ARG) {
       matches
         .values_of(arg::SHELL_ARG)
         .unwrap()
```
casey: Instead of testing for existence and then unwrapping, this can be replaced with an `if let`:

```suggestion
    let shell_args = if let Some(shell_args) = matches.values_of(arg::SHELL_ARG) {
        shell_args
```

src/config.rs
```diff
@@ -558,26 +554,27 @@ impl Config {
...
+    let shell_args_present = matches.occurrences_of(arg::CLEAR_SHELL_ARGS) > 0
       || matches.occurrences_of(arg::SHELL_ARG) > 0;
 
     Ok(Self {
       check: matches.is_present(arg::CHECK),
       dry_run: matches.is_present(arg::DRY_RUN),
       dump_format: Self::dump_format_from_matches(matches)?,
       highlight: !matches.is_present(arg::NO_HIGHLIGHT),
-      shell: matches.value_of(arg::SHELL).unwrap().to_owned(),
+      shell: matches.value_of(arg::SHELL).unwrap_or("").to_owned(),
```
casey: `""` shouldn't be used to indicated that the argument isn't present, since the user could actually pass the empty string as the argument. We should use an `Option` for the `shell` field.

src/evaluator.rs
```diff
@@ -303,6 +303,7 @@ impl<'src, 'run> Evaluator<'src, 'run> {
 }
 
 #[cfg(test)]
+#[allow(dead_code)]
```
casey: Did something change which makes this necessary?

src/justfile.rs
```diff
@@ -474,6 +474,7 @@ mod tests {
   run_error! {
```
casey: This test is unrelated to this PR, so should be reverted. This PR should add additional tests, not change existing tests.

src/justfile.rs
```diff
@@ -486,10 +487,11 @@ mod tests {
     check: {
       assert_eq!(recipe, "fail");
       assert_eq!(code, 100);
-      assert_eq!(line_number, Some(2));
+      assert_eq!(line_number, Some(3));
     }
   }
 
+  #[cfg(target_os = "linux")]
```
casey: This change doesn't seem related to this PR.

src/justfile.rs
```diff
@@ -509,6 +511,27 @@ mod tests {
     }
   }
 
+  #[cfg(target_os = "windows")]
```
casey: This change doesn't seem related to this PR.

I think this PR should get a new test which just tests the feature being added.

This could be as simple as:

```rust
#[test]
fn windows_poweshell_setting_uses_powershell() {
  Test::new()
    .justfile(
     "
      set windows-powershell

      foo:
        Write-Output bar
    "#,
    )
    .stdout("bar")
    .stderr("Write-Output bar")
    .run();
}
```

This uses the new, struct-based tests instead of macros, which are preferred at this point, although both kinds of tests still exist.

This test can go in a new module `tests/windows_powershell.rs`

src/config.rs
casey: This might also let us get rid of `shell_present`, which would be nice.

src/config.rs
```diff
@@ -30,6 +27,7 @@ pub(crate) struct Config {
   pub(crate) shell_args: Vec<String>,
   pub(crate) shell_command: bool,
   pub(crate) shell_present: bool,
+  pub(crate) shell_args_present: bool,
```
casey: Instead of using `shell_args_present`, let's turn `shell_args` into an `Option<Vec<String>>`.

src/settings.rs
```diff
@@ -1,11 +1,17 @@
 use crate::common::*;
 
+pub(crate) const DEFAULT_SHELL: &str = "sh";
+pub(crate) const DEFAULT_SHELL_ARG: &[&str] = &["-cu"];
```
casey:
```suggestion
pub(crate) const DEFAULT_SHELL_ARGS: &[&str] = &["-cu"];
```

src/settings.rs
```diff
@@ -1,11 +1,17 @@
...
+pub(crate) const WINDOWS_DEFAULT_SHELL: &str = "powershell.exe";
+pub(crate) const WINDOWS_DEFAULT_SHELL_ARG: &[&str] = &["-NoLogo", "-Command"];
```
casey:
```suggestion
pub(crate) const WINDOWS_DEFAULT_SHELL_ARGS: &[&str] = &["-NoLogo", "-Command"];
```

src/settings.rs
```diff
@@ -27,22 +34,176 @@ impl<'src> Settings<'src> {
...
+    // 3) Default/PowerShell
+    let shell_or_args_present = config.shell_present | config.shell_args_present;
+
+    if let (Some(shell), false) = (&self.shell, shell_or_args_present) {
       shell.command.cooked.as_ref()
-    } else {
+    } else if config.shell_present {
       &config.shell
+    } else {
+      // Default value for shell
```
casey:
```suggestion
```

src/settings.rs
```diff
@@ -27,22 +34,176 @@ impl<'src> Settings<'src> {
   }
 
   pub(crate) fn shell_binary<'a>(&'a self, config: &'a Config) -> &'a str {
-    if let (Some(shell), false) = (&self.shell, config.shell_present) {
+    // Order of precedence:
+    // 1) CLI argument (config)
+    // 2) Settings in justfile (settings)
+    // 3) Default/PowerShell
```
casey: I think this is mostly evident from reading the code, so I'd remove this comment.

```suggestion
```

src/settings.rs
```diff
@@ -27,22 +34,176 @@ impl<'src> Settings<'src> {
...
+        WINDOWS_DEFAULT_SHELL
+      } else {
+        DEFAULT_SHELL
+      }
     }
   }
 
   pub(crate) fn shell_arguments<'a>(&'a self, config: &'a Config) -> Vec<&'a str> {
-    if let (Some(shell), false) = (&self.shell, config.shell_present) {
+    // Order of precedence like in fn shell_binary
```
casey:
```suggestion
```

src/settings.rs
```diff
@@ -27,22 +34,176 @@ impl<'src> Settings<'src> {
...
       shell
         .arguments
         .iter()
         .map(|argument| argument.cooked.as_ref())
         .collect()
-    } else {
+    } else if config.shell_args_present {
       config.shell_args.iter().map(String::as_ref).collect()
+    } else {
+      // Default value for shell-args
```
casey:
```suggestion
```

casey (changes requested):
Nice, this mostly looks good. Thanks for writing all the unit tests!

As for integration tests, there should be a new integration test dedicated to this feature, which I suggested in a comment, and none of the existing integration tests should change.

The reason being that when reviewing a PR like this, you really want to be sure that the feature works, and that nothing else changed. The easiest way to do this is to leave the existing tests alone, and write a new test that just covers the new feature.

src/evaluator.rs
michidk: Probably a leftover of some testing. Will remove

src/config.rs
michidk: Actually clippy want's us to use
```rust
    let shell_args = matches
      .values_of(arg::SHELL_ARG)
      .map_or(Vec::new(), |shell_args| {
        shell_args.map(str::to_owned).collect()
      });
```

src/config.rs
michidk: Done

michidk:
@casey Thanks for your review! I implemented all the changes.

README.adoc
```diff
@@ -656,6 +657,12 @@ foo:
 
 `just` passes the command to be executed as an argument. Many shells will need an additional flag, often `-c`, to make them evaluate the first argument.
 
+==== Windows PowerShell
+
+`just` uses `sh` on Windows by default. To use PowerShell instead, set `windows-powershell` to true.
+
+
+
```
casey, line 665:
```suggestion
```

src/config.rs
```diff
@@ -558,26 +552,32 @@ impl Config {
...
     };
 
-    let shell_present = matches.occurrences_of(arg::CLEAR_SHELL_ARGS) > 0
-      || matches.occurrences_of(arg::SHELL) > 0
-      || matches.occurrences_of(arg::SHELL_ARG) > 0;
+    let shell = if matches.occurrences_of(arg::SHELL) > 0 {
+      Some(matches.value_of(arg::SHELL).unwrap().to_owned())
+    } else {
+      None
+    };
```
casey: Can this be:

```suggestion
    let shell = matches.value_of(arg::SHELL).map(str::to_owned);
```

README.adoc
```suggestion
```

src/justfile.rs
```diff
@@ -471,25 +471,6 @@ mod tests {
     }
   }
 
-  run_error! {
```
casey: This test should be restored.

src/settings.rs
casey: I think this conversation was accidentally marked as resolved. Since it's an array of strings, I would call it `DEFAULT_SHELL_ARGS` even if there's only one.

src/settings.rs
```diff
@@ -1,11 +1,17 @@
 use crate::common::*;
 
+pub(crate) const DEFAULT_SHELL: &str = "sh";
+pub(crate) const DEFAULT_SHELL_ARG: &[&str] = &["-cu"];
+pub(crate) const WINDOWS_DEFAULT_SHELL: &str = "powershell.exe";
+pub(crate) const WINDOWS_DEFAULT_SHELL_ARGS: &[&str] = &["-NoLogo", "-Command"];
```
casey:
```suggestion
pub(crate) const WINDOWS_POWERSHELL_SHELL: &str = "powershell.exe";
pub(crate) const WINDOWS_POWERSHELL_ARGS: &[&str] = &["-NoLogo", "-Command"];
```

src/settings.rs
```diff
@@ -27,22 +34,156 @@ impl<'src> Settings<'src> {
   }
 
   pub(crate) fn shell_binary<'a>(&'a self, config: &'a Config) -> &'a str {
-    if let (Some(shell), false) = (&self.shell, config.shell_present) {
+    let shell_or_args_present = config.shell.is_some() | config.shell_args.is_some();
```
casey: It's better to use `||` instead of `|`. They both work, but for clarity, `|` should only be used when you want bitwise or.

```suggestion
    let shell_or_args_present = config.shell.is_some() || config.shell_args.is_some();
```

tests/lib.rs
```diff
@@ -40,4 +40,5 @@ mod sublime_syntax;
 mod subsequents;
 mod tempdir;
 mod undefined_variables;
+mod windows_powershell;
```
casey, line 44:
```suggestion
#[cfg(target_family = "windows")]
mod windows_powershell;
```

tests/windows_powershell.rs
```diff
@@ -0,0 +1,20 @@
+#[cfg(target_family = "windows")]
```
casey:
```suggestion
```

tests/windows_powershell.rs
```diff
@@ -0,0 +1,20 @@
...
+use crate::common::*;
+
+#[test]
+#[cfg(target_family = "windows")]
```
casey:
```suggestion
```

src/settings.rs
```diff
@@ -27,22 +34,156 @@ impl<'src> Settings<'src> {
...
+#[cfg(test)]
+mod tests {
+  use crate::setting::Shell;
+
+  use super::*;
+
+  #[test]
+  fn default_shell() {
+    let mut settings = Settings::new();
+    settings.windows_powershell = false;
```
casey: This defaults to false, so it can be omitted:

```suggestion
```

src/settings.rs
```diff
@@ -27,22 +34,156 @@ impl<'src> Settings<'src> {
...
+  }
+
+  #[test]
+  fn overwrite_shell() {
+    let mut settings = Settings::new();
+    settings.windows_powershell = false;
+
+    let config = Config {
+      shell_command: true,
+      shell: Some("lol".to_string()),
```
casey, line 116: lol shell is best shell

src/settings.rs
```diff
@@ -27,22 +34,156 @@ impl<'src> Settings<'src> {
...
+    };
+
+    assert_eq!(settings.shell_binary(&config), "lol");
+    assert_eq!(settings.shell_arguments(&config), vec!["-nice"]);
+  }
+
+  #[test]
+  fn shell_cooked() {
+    let mut settings = Settings::new();
+    settings.windows_powershell = false;
```
casey:
```suggestion
```

src/settings.rs
```diff
@@ -27,22 +34,156 @@ impl<'src> Settings<'src> {
...
+      shell_command: false,
+      ..testing::config(&[])
+    };
+
+    assert_eq!(settings.shell_binary(&config), "asdf.exe");
+    assert_eq!(settings.shell_arguments(&config), vec!["-nope"]);
+  }
+
+  #[test]
+  fn shell_args_present_but_not_shell() {
```
casey, line 181: For completeness, it would be good to see something like `shell_present_but_not_shell_args`

casey (changes requested):
Getting there!

michidk:
Done!

src/settings.rs
```diff
@@ -27,22 +34,167 @@ impl<'src> Settings<'src> {
...
+    } else {
+      assert_eq!(settings.shell_binary(&config), "sh");
+      assert_eq!(settings.shell_arguments(&config), vec!["-cu"]);
+    }
+  }
+
+  #[test]
+  fn overwrite_shell() {
+    let mut settings = Settings::new();
+    settings.windows_powershell = false;
```
casey:
```suggestion
```

src/settings.rs
```diff
@@ -27,22 +34,166 @@ impl<'src> Settings<'src> {
...
+      );
+    } else {
+      assert_eq!(settings.shell_binary(&config), "sh");
+      assert_eq!(settings.shell_arguments(&config), vec!["-cu"]);
+    }
+  }
+
+  #[test]
+  fn overwrite_shell() {
+    let mut settings = Settings::new();
```
casey:
```suggestion
    let settings = Settings::new();
```

casey:
Nice, merged! Thanks for sticking with it! This is a great feature, and long overdue.

michidk:
Thanks for reviewing!

Issue #1054, opened 2021-12-24
How to pass on command line arguments unchanged?


mpdude:
Please consider this `Justfile`:

```justfile
test *args='':
    bash -c 'while (( "$#" )); do echo - $1; shift; done' -- "{{args}}"
```

When running:

`$ just test first "second arg" third`

... I get ...

`- first second arg third`

What syntax should I use for `{{args}}` in the `Justfile` to pass on arguments unchanged, i. e. to get as result:

```
- first
- second arg
- third
```

🤷🏻‍♂️

casey:
The best thing to do is to turn on `positional-arguments`, which passes positional arguments to each recipe line, which avoids unwanted splitting:

```
set positional-arguments

test *args='':
    bash -c 'while (( "$#" )); do echo - $1; shift; done' -- "$@"
```

This works for me, but feel free to reopen if it doesn't work for your use case!

mpdude:
Will try, thanks!

Probably `"$@"` is what I should have been trying (and know from bash), but did not notice it in the README.

PR #1064, opened 2022-01-03
Mention "$@" in the README

mpdude:
This documents the feature explained in #1054. Much of the text is taken from `man bash`.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -619,7 +619,7 @@ foo
 hello
 ```
 
-When using a `sh`-compatible shell, such as `bash` or `zsh, `$@` expands to the positional arguments given to the recipe, starting from one. When used within double quotes as `"$@"`, arguments including whitespace will be passed on as if they were double-quoted. That is, "$@" is equivalent to "$1" "$2"... When there are no positional parameters, "$@" and $@ expand to nothing (i.e., they are removed).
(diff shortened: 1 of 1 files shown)


casey:
Nice, this is a great addition to the readme! Can you also mention that this is bourn shell specific? Maybe something like: "When using a `sh`-compatible shell, such as `bash` or `zsh,"

mpdude:
I did not know that – only _just_ learning ;-).

Text is updated, don't forget to squash-merge.

README.adoc
```diff
@@ -619,6 +619,25 @@ foo
 hello
 ```
 
+When using a `sh`-compatible shell, such as `bash` or `zsh, `$@` expands to the positional arguments given to the recipe, starting from one. When used within double quotes as `"$@"`, arguments including whitespace will be passed on as if they were double-quoted. That is, "$@" is equivalent to "$1" "$2"... When there are no positional parameters, "$@" and $@ expand to nothing (i.e., they are removed).
```
```suggestion
When using an `sh`-compatible shell, such as `bash` or `zsh`, `$@` expands to the positional arguments given to the recipe, starting from one. When used within double quotes as `"$@"`, arguments including whitespace will be passed on as if they were double-quoted. That is, "$@" is equivalent to "$1" "$2"... When there are no positional parameters, "$@" and $@ expand to nothing (i.e., they are removed).
```

PR #1066, opened 2022-01-04
Powershell exit codes unit tests

michidk:


casey:
I think this was covered in the merged PR.

michidk:
@casey I just tried those tests on the current master and all but the fist one still fail

casey:
Gotcha, sorry for closing prematurely! So the issue is that powershell doesn't correctly report exit codes? When you run those commands in a powershell terminal, can you get the correct exit codes?

michidk:
I am not too familiar with PSS myself, but it seems to work just fine in the console:
```
PS C:\Users\michi> powershell.exe -c exit 200
PS C:\Users\michi> echo $?
False
PS C:\Users\michi> powershell.exe -c exit 0
PS C:\Users\michi> echo $?
True
PS C:\Users\michi> echo $LASTEXITCODE
0
PS C:\Users\michi> powershell.exe -c exit 200
PS C:\Users\michi> echo $LASTEXITCODE
200
```

casey:
I think maybe we just need to treat this as a bug in powershell. The just-side code does exactly the same thing for every shell, and it seems very unlikely that there's a bug that's causing this error on our side, so we might not be able to fix it.

casey:
I think that powershell doesn't propagate codes when using `-Command` the way other shells do, see this stack overflow post:

https://stackoverflow.com/questions/50200325/returning-an-exit-code-from-a-powershell-script

So I think this is actually working as indented, powershell translates any non-zero exit code into `1`. Pretty odd behavior!

PR #1067, opened 2022-01-04
Upgrade SNAFU to 0.7

shepmaster:


casey:
Thanks for opening this PR! Is the `Snafu` suffix configurable? I found it pretty hard to understand why `CurrentDir` changed to `CurrentDirSnafu`, and would prefer something like `CurrentDirContext` which might be more self-explanatory.

shepmaster:
Thanks for using SNAFU!

>  Is the Snafu suffix configurable?

[It sure is](https://docs.rs/snafu/latest/snafu/derive.Snafu.html#changing-the-context-selector-suffix)!

casey:
🙌

PR #1069, opened 2022-01-05
Implement Custom Functions

### LINKED ISSUE #1059: User-defined functions
Awesome tool, I'm mostly very happy with it, and this is a relatively minor QOL issue issue with two workarounds I mention below.

My use-case is something like:
```
build_dir := "build"
file_name := "out"

default arg="": (html arg) (pdf arg)

html arg="":
    pandoc input.md -o "{{build_dir}}/{{arg}}/{{file_name}}.html"

pdf arg="":
    pandoc input.md -o "{{build_dir}}/{{arg}}/{{file_name}}.pdf"
```
I would like a way to deduplicate the specific string formatting `{{build_dir}}/{{arg}}/{{file_name}}` so that my uses of it don't become out of sync. My current solution is to use `arg` as a just variable instead of an argument, i.e. `arg := ""` and `file_path := build_dir + "/" + arg + "/" + file_name` at the top of the `Justfile`. This is imperfect because it requires passing the argument to the CLI as `just pdf arg=foo` instead of `just pdf foo`, and also it puts into global scope something that shouldn't necessarily be there.

Another workaround is something like this:
```
[...]

default arg="": (_default (build_dir + "/" + arg + "/" + file_name))
_default arg: (html arg) (pdf arg)

html arg="":
    pandoc input.md -o "{{arg}}.html"

pdf arg="":
    pandoc input.md -o "{{arg}}.pdf"
```
But this is inelegant and doesn't allow the `html` and `pdf` targets to be run directly.

Is there a better way to accomplish this currently? If not, is it possible to add some syntax for declaring functions on strings? I'm imagining something like (this is obviously just a first draft at a syntax):
```
buildpath(s) := build_dir + "/" + s + "/" + file_name

pdf arg="":
    pandoc input.md -o "{{buildpath(arg)}}.pdf"
```
I haven't looked super closely into the code base so I'm not sure how easy this would be. I'd be happy to look into writing a PR for this if it's desirable and feasible. Regardless, thanks for a great tool.

rileyshahar:
This will close #1059.

Todo:

- [X] Add integration tests.
- [X] Add closures to the formal grammar.
- [X] Create an AST item representing closures.
  - [X] Implement `Display` for `NamedClosure`.
  - [X] Implement `Node` for `NamedClosure`.
- [X] Parse closures.
- [ ] Analyze closures.
  - [X] Error on invalid function calls (e.x. insufficient arguments).
  - [ ] Do much of the stuff the assignment resolver has to do, I assume. (If possible consider extracting code.)
- [ ] Evaluate closures to functions.
  - [ ] Handle capturing of context properly.
  - [ ] Figure out how and whether to handle backticked closures.
- [ ] Properly shadow built-in functions.
- [ ] Evaluate closures as functions via a new thunk variant which defers knowledge of its function.
- [ ] Add documentation for closures.
- [ ] Implement proper error UI for closures.
  - [ ] Invalid calls should fail at compile time.


rileyshahar:
@casey (or others), thoughts on whether it should be possible to have a user function with the same name as a variable? There shouldn't be ambiguity because we don't have support for first class functions, so you can never have a closure object in an expression position. Also, right now you can have a variable with the same name as a built in function:
```
os := "nintendo 3ds"

foo:
  echo {{os}}
  echo {{os()}}
```
so maybe for consistency the same should be true of user functions.

On the other hand, if there's ever a desire for first class functions, I guess it would break backwards compatibility if we don't implement this restriction now. I sort of assume that's not a feature there's any good reason to support, but I want to check first, because it's not a technically very difficult restriction to implement.

casey:
Good question! I think we should disallow having variables and functions with the same name. It's easier to remove restrictions than it is to add them, and `just` tries very hard to be backwards compatible.

rileyshahar:
I guess my opinion is that it feels weird to have different restrictions here for built-in and user functions, and if you ever want first-class functions (or any other feature that requires disambiguation between variables and functions) there'd already need to be a breaking change for built-in functions. Just my two cents, I'll obviously go with what you decide.

rileyshahar:
Ah, also, here's a related issue: what is the proper behavior for redefining builtin functions? Either this is an error, in which case it becomes a breaking change to add any new builtin function, or the user function masks the builtin function, which probably is at least slightly error-prone. Presumably the second option is best, but we raise a warning?

casey:
> I guess my opinion is that it feels weird to have different restrictions here for built-in and user functions, and if you ever want first-class functions (or any other feature that requires disambiguation between variables and functions) there'd already need to be a breaking change for built-in functions. Just my two cents, I'll obviously go with what you decide.

Ahh, right, I missed the fact that you can currently shadow built-in functions. In that case, we should do what you suggest, and allow functions and variables to have the same name.

It might be nice in the future to fix that, to allow first-class functions, at which point I'll probably introduce an edition system like Rust's, where you can do something like `set edition := 2022` and opt into a set of changes that would otherwise be breaking.

casey:
> Ah, also, here's a related issue: what is the proper behavior for redefining builtin functions? Either this is an error, in which case it becomes a breaking change to add any new builtin function, or the user function masks the builtin function, which probably is at least slightly error-prone. Presumably the second option is best, but we raise a warning?

Hmmm, also a good question. Let's allow user-defined functions to shadow built-in functions. What do you think about no warning? It would be annoying if your old justfiles started to give warnings because a new built-in function was introduced.

rileyshahar:
Yeah, that makes sense. I don't think it's a huge deal I just think it might lead to confusing behavior in sufficiently large justfiles. No warning is fine though.

rileyshahar:
I think we'll end up having to dispatch functions at evaluation time.

Here's the issue: in order to parse a call to a `Thunk`, you want to know whether the call is of a built-in or user-defined function, but at parse time you don't yet know all of the user-defined functions, so in particular I can't see how to implement shadowing properly. There are a couple local solutions I can see:
1. when a closure is parsed, go back through all previously-parsed `Thunk`s and turn ones with matching names into `Thunk::User`s. This is pretty obviously not desirable IMO.
2. Change the shadowing rules such that e.x.
```
foo := os()
os() := "not linux"
bar := os()

default:
  echo {{foo}}
  echo {{bar}}
```
echoes
```
linux
not linux
```
This is probably a more standard solution, but seems undesirable for a language like `just`; I kind of think you want order of symbols in the file to be mostly irrelevant. It also requires that you pass some kind of parsing context to the function parser. I could see an argument for this though.

The other solution, which I guess I prefer right now, is to parse a call into an object like
```rust
struct Call<'src> {
  name: Name<'src>,
  args: Vec<Expression<'src>>,
}
```
Then we'd make the analyzer responsible for checking existence of the function and argument count, with the evaluator responsible for dispatching the call to either the rust code of the builtin function or the just code of the user function.

casey:
> This is probably a more standard solution, but seems undesirable for a language like `just`;

We definitely want to avoid this. Just resolves all symbols at compile time, and the order of all symbols (recipes, variables, settings, etc) does change execution or meaning.

> The other solution, which I guess I prefer right now, is to parse a call into an object like
> 
> ```rust
> struct Call<'src> {
>   name: Name<'src>,
>   args: Vec<Expression<'src>>,
> }
> ```
> 
> Then we'd make the analyzer responsible for checking existence of the function and argument count, with the evaluator responsible for dispatching the call to either the rust code of the builtin function or the just code of the user function.

I think this is the way to go, move function call resolution out of the parser and into the analyzer. The analyzer can then resolve function calls because it has full information.

I think it's just currently a quirk of the fact that there are no user-defined functions that the parser currently resolves them.

rileyshahar:
Just updating you, I haven't disappeared, just got busy. Should have time to work on this around Wednesday.

casey:
No worries, this is a big feature! No rush.

danielb2:
Is it Wednesday yet?

casey:
> Is it Wednesday yet?

We're all just like, unpaid nerds on the internet, working on this, so it may take many Wednesdays for any particular feature to be implemented.

danielb2:
> > Is it Wednesday yet?
> 
> We're all just like, unpaid nerds on the internet, working on this, so it may take many Wednesdays for any particular feature to be implemented.

and I'm just like, asking for an update. you know? like, no pressure 🤷‍♂️

casey:

PR #1083, opened 2022-02-02
Upgrade clap to 3.0

casey:


casey:
Closing, since there isn't a huge benefit, and it breaks the completion scripts.

PR #1089, opened 2022-02-05
Convert the README to Markdown

### LINKED ISSUE #1087: Convert readme to markdown
Now that GitHub auto-generates a table of contents, I think asciidoc has outlived its usefulness, and I should convert the readme to markdown so I don't have to remember the asciidoc syntax.

halostatue:
Resolves #1087

- Initially converted using `pandoc` via the `docbook` format. This introduced more than a few formatting oddities, so I went through and reviewed all of the cases, usually copying the fenced code blocks from the `asciidoc` file.

- Modified _most_ external references to use named references to improve the readability of the un-rendered version. If a link is specified as `[Foo]`, it means the same as `[Foo][]`, which means `[Foo][foo]`. This requires that there be a named reference `[foo]: http://foo.example.com` that I have placed at the bottom of the file. Named references are case-insensitive (at least on GitHub).

  - This makes a _huge_ difference in the readability of the installation table.

- Fixed some inconsistencies in fenced syntax for highlighting.

- Made `justfile` example indentation consistent (four spaces in all cases, although my preference would be two).

- Ensured that all instances of the word `justfile` (_except_ `Justfile and Justfile Directory`) are code-quoted, including the plural (`justfile`s).


halostatue:
For some reason, the rendered file isn’t visible in the diff. You can see the rendered file [here](https://github.com/halostatue/just/tree/convert-README-to-markdown).

I didn’t embed a table of contents because it is just using the automatic TOC in the document navigation frame from GitHub.

casey:
Dope, thank you so much!

What was the original `pandoc` command that you used to convert from asciidoc to markdown? I want to be able to run the command, and then diff the output against the `.md` file in the PR so that I can properly review the changes.

> Modified most external references to use named references to improve the readability of the un-rendered version.

I hate to say it, because I know it must have been a lot of work to switch the references to external references, but I think I would prefer in-line references. I agree that external references are more readable, but in-line references are easier to write, are what most people are used to (so PRs that change the readme are likely to use in-line references, if they don't notice that there are a lot of external references), and don't introduce the possibility that a reference name is misspelled, leading to a broken link. Would you mind changing it back to in-line references?

If you look at the original `adoc` source, I actually format the table so that each row is separated by a blank line. I don't know if that's possible for markdown tables, but it helps with readability, even though it doesn't look great.

> Made justfile example indentation consistent (four spaces in all cases, although my preference would be two).

I also prefer two-space indentation, so if you feel like changing it to two, I'd be in favor, but also not a huge deal.

halostatue:
> What was the original pandoc command that you used to convert from asciidoc to markdown? I want to be able to run the command, and then diff the output against the .md file in the PR so that I can properly review the changes.

I basically followed the instructions [in a blog post](https://karinavarela.me/2020/05/06/converting-asciidoc-to-markdown/), with the change that instead of telling pandoc to convert with `markdown_strict`, I converted with `gfm`. The _problem_ is that what I had to use (the `iconv pipeline`)…erred on several fronts. Some of the conversions made automatically were problematic in multiple ways:

- Certain things like \`make\` were represented as \\\`\\\`make\\\`\\\`.
- Most of the fenced syntax blocks were _wrapped_ instead of preserved (infuriating, and part of the reason I had to compare the two side-by-side).
- Some of the content in the fenced syntax blocks were corrupted such that `'Quoted Text'` was presented as `_Quoted Text`.

All in all, what I _thought_ would be a ~30 minute task turned into much more than that (but not overall unpleasant).

> > Modified most external references to use named references to improve the readability of the un-rendered version.

> I hate to say it, because I know it must have been a lot of work to switch the references to external references, but I think I would prefer in-line references. I agree that external references are more readable, but in-line references are easier to write, are what most people are used to (so PRs that change the readme are likely to use in-line references, if they don't notice that there are a lot of external references), and don't introduce the possibility that a reference name is misspelled, leading to a broken link. Would you mind changing it back to in-line references?

I _can_, but I recommend against it for at least a couple of places and for various reasons. I’ve been using Markdown for a long time and while I _did_ have this happen a few times during the conversion phase (part of the reason that there’s a number of force pushes above), all of the items were easy to find and fix (it’s easier to find and fix with the _full_ named reference syntax, e.g., `[link text][link name]`, because a bad name renders as that in HTML, which makes it _really_ easy to find (it does not appear that [markdownlint](https://github.com/DavidAnson/markdownlint) offers this as a lint, but it might be worth adding as that is available as a GitHub Action).

There are a couple of places where I recommend the named reference because the link is used multiple times.

> If you look at the original adoc source, I actually format the table so that each row is separated by a blank line. I don't know if that's possible for markdown tables, but it helps with readability, even though it doesn't look great.

Yeah. That’s the one place where I _won’t_ make the link changes, because the table becomes unmanageably wide. All cells of Markdown tables, at least those supported by GFM, must be on the same horizontal line. By default, you don’t _have_ to make the cells the same width, but any modern Markdown formatter _automatically_ (like the one I use with Vim or that often gets added to VS Code) pads the table to the width of each cell.

Like I said, I have no problem changing it back in most places (it is pretty easy to write a script to undo what I did), but one of the things I believe _fervently_ about Markdown documents is that they should be readable as plaintext as well as rendered. A table that is ~200 characters wide or more is unusable (I have another PR open on another project that had one that I could not keep unwrapped and fully visible at 5pt font size; they _also_ want me to change the table back to the inline link format; I will let the PR expire, TBH). For this one, I won’t make those changes for the table, but I _have_ made this editable by project committers, so you are welcome to do that yourself for comparison.

casey:
I agree that a wide table is less than ideal. In that case, how about changing the links outside the table back to inline references, and for the table, using external references, but putting them right under the table instead of at the bottom of the document. That way, you don't have to scroll to the bottom of the document to find them, and it'll be easier to keep them in sync.

halostatue:
That I can do. My general preference is to put all references at the bottom, but references can be anywhere.

halostatue:
Done. I have also updated the `justfile` so that `render-readme` works with this (and uses `pandoc` if it thinks it can find it; a boolean `has_executable('foo')` function would be _really_ useful here). The logic I’m using may be too clever by half.

casey:
I started reviewing this, by comparing the differences between what is output by `pandoc` and what's in the `md` file, and there were too many diffs to make for a reasonable review. I think that a lot of the changes that you made are great, but reviewing them together with the very large base diff of converting to markdown, which makes it impossible to review the manual diffs individually.

I'm going to close this in favor of #1091, which is admittedly worse in some ways. I used pandoc to do the conversion, and fixed the poor conversions in ways that were easier to review. (I added attributes to the code blocks, which made them not use indented code blocks, etc.) I didn't wrap the text, and I didn't use external links in the table, both of which I think are definitely improvements, but I felt like they should be left to future PRs.

I'm sorry to close this PR, since a lot of effort went into it, but I wanted to get the expedient thing done, especially to avoid piling up merge conflicts.

halostatue:
No problem. I’ll look at doing some comparisons and submit a pull request for the differences now that `README.md` is present.

PR #1093, opened 2022-02-07
Various README fixes

halostatue:
Depending on your preferences, @casey, I can open this as multiple PRs. I believe it will be faster to get them in place with one PR. As each change builds on the last but is otherwise isolated, I recommend reviewing this PR a commit at a time. What follows is a summary of the commits based on the commit messages for each.

### Prefer ATX headings over SETEXT headings

The two heading types are equivalent, but SETEXT headings only support two levels (`===` [h1] and `---` [h2]), and there are enough other headings to make it worthwhile to use ATX headings exclusively.

### Replace Unicode punctuation with ASCII punctuation

- Replace `’` with `'` globally. The Unicode version is more correct, but most modern Markdown converters apply SmartyPants conversions to automatically convert `'` to `’`.

- Replace instances of `…` in code blocks with `...`. Left the Unicode `…` characters in place in prose as I don’t think that SmartyPants conversions handle those.

### Changed in-document URLs to not go remote

Removed the prefix `https://github.com/casey/just` from in-document URLs so that the links work in local Markdown viewers without redirecting to GitHub.

As an example, `[much more](https://github.com/casey/just#manual)` becomes `[much more](#manual)`.

### Automatic `prettier` formatting

- This fixes various small formatting issues.

### Make the styling of `justfile` consistent

Modified the styling of `justfile` to always be presented in code format, including for the plural as `justfile`s. There are a couple of places where this has not been followed, such as in the function group title `#### Justfile and Justfile Directory`.

### Make `justfile` syntax indentation consistent

Changed indentation of `justfile` syntax to two spaces consistently.

### Fix various errors

- Fixed the use of the en-dash (`–`) where the em-dash (`—`) belongs.
- Fixed the unnecessary use of backticks inside a syntax block.
- Fixed a couple of sentence errors.
- Fixed cases where text should have been been wrapped in code blocks and weren’t or were wrapped in other syntax markers.

### Clean up the installation table

Convert the in-line links to reference links. Note that the references immediately below the table are not sorted, but are presented in the order of appearance in the table.

### Move `Dotenv Integration` section

This was placed under the `Functions` section, but belongs under the `Features` section. I moved it to a place that _seems_ to make sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ You can also set the shell using command-line arguments. For example, to use Pow
 ### Packages
 
 | Operating System                                                                                                                                                | Package Manager                                                                | Package                                                                                                  | Command                                                                                 |
-|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------|
+| --------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------ | -------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------- |
 | [Various](https://forge.rust-lang.org/release/platform-support.html)                                                                                            | [Cargo](https://www.rust-lang.org)                                             | [just](https://crates.io/crates/just)                                                                    | `cargo install just`                                                                    |
 | [Microsoft Windows](https://en.wikipedia.org/wiki/Microsoft_Windows)                                                                                            | [Scoop](https://scoop.sh)                                                      | [just](https://github.com/ScoopInstaller/Main/blob/master/bucket/just.json)                              | `scoop install just`                                                                    |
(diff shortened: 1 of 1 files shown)


casey (changes requested):
Awesome! A couple things that I would change:

- I prefer using setext headings where possible, since I find them easier to read.
- I like replacing ’ with ', but I'd prefer to keep `…` instead of `...`, I like the way that `…` looks

Everything else is great. Thanks for splitting it up into separate commits, that's makes it much easier to review.

halostatue:
I’ll revert back to SETEXT later (I have to disable Neoformat because `prettier` automatically changes SETEXT headings to ATX headings), but I’d like you to look at the commit I’m pushing up now where I am doing a replace of all `...` with `…`. This _only_ affects code wrapped text (<code>\`text\`</code>) and fenced blocks (<code>\`\`\`…\`\`\`</code>) to see if there are spots where you don’t want `…` instead of `...`.

casey:
Sweet, the change you just pushed looks good to me.

halostatue:
OK. I’ll get the headings changed and finalize this PR in a few hours.

halostatue:
This should finalize the cleanup PR.

casey (approved):
Sweet, looks great. Thank you for persevering! This makes the readme much more consistent.

Issue #166, opened 2017-03-30
Consider renaming `recipe` to `task`


casey:
The word "recipe" comes from make's terminology. However, since `just` is not a build system that is intended to produce artifacts, the term is a bit lacking. The best alternative to is probably `task`.

I'm waffling over whether or not this change is a good idea, so 👍 if you prefer the name `task`, and 👎 if would rather that we keep the existing `recipe` terminology.

kbknapp:
`action`?

casey:
Action is okay, although I haven't seen it used in that sense before. I like that they would be "just actions" though. It has kind of a poetic ring to it.

Ezku:
Since the first comment had multiple suggestions, can I 👍  "just tasks"? :)

TeddyDD:
I like term "recipe". It sounds... tasty 🍰

runeimp:
I like `task`, `action`, and `chore` better than `recipe`, myself. And in that order.

dionjwa:
I would go with `command`. Here are some extremely subjective and personal reasons (add "for me" to every statememt :-)):

 - `task` has other connotations, e.g. compute workflows (with tasks), CI systems with build tasks, and the like
 - `command` puts the emphasis on the user of `just`. I see `just` as a way of nicely packaging and specifying commands, that I as a user am invoking. `task` implies some automated procedures
 - I use `just` as a kind of "command-ops", as an extension of "git-ops". "task-ops" doesn't capture the power and end user ease and UX enjoyment compared to "command-ops". 
 - `command` sounds cooler (I am very scientific)
 - replace "task" with "action" and my above preferences stand

I can't believe I have such strong preferences but `just` is changing the way I organize complex mono-repos, and I have manually changed the default to "Commands:" myself. Thanks for changing my life @casey !

casey:
I think the ship has sailed on this one.

PR #1095, opened 2022-02-10
Allow duplicate recipes

### LINKED ISSUE #383: Modules and subcommands
This is apropos of @valscion's comment in #208, where he talks about some of the annoyances of using `just` to emulate subcommands. This is also somewhat related to #237 and #367, where sub-justfiles are discussed.

I'd like to come up with something that enables all of these things. Some ideas:

A statement `mod foo` parses `foo.just` or `foo/mod.just` as a submodule. Recipes in `foo.just` / `foo/mod.just` can now be called on the command line with `just foo RECIPE`.

If the submodule source file is `foo.just`, the current working directory for recipes in that submodule will be the current directory. If the it is in `foo/mod.just`, the current working directory will be `foo/`.

Questions:

- What do we do about directories with weird characters? I think punting is reasonable. But eventually we might want to support something like `mod "path with spaces/bar"`. Any solution should be based on user feedback.

- How do we handle dependencies between recipes and variables in different submodules?

- Can you have a submodule in a submodule, and call those commands?

---

Notes for a first path for out-of-line modules only:

- make a module loading step after parsing
- look at parsed module and load all other modules, recursively
- module map has all modules in it
- pass module map to the analyzer
- analyzer produces justfile

---

- will need to modify error messages to include file path
- will need to modify argument parser to successfully call submodules
- will need to disallow recipes and modules to have same name

lutostag:
Hey there @casey first off just wanted to thank you for an awesome tool. I personally love using it in all my projects.

This is meant to address the allowing duplicate recipes via a setting. This should help partially resolve [the conversation in how to override recipes](https://github.com/casey/just/issues/383#issuecomment-882002853) when using `justprep`.

I know this conversation was a while ago, and your reasoning may have changed (I really want modules or recipes that can have prereqs from other files), but `justprep` is the closest we can get to having modular "dependencies" at the moment, and this is sort of the last piece that would help our workflows.

Why I chose to use a setting rather than a cli flag: if there are duplicate recipes in the justfiles, they will not parse, rather than pass around the config flags everywhere it seemed logical to have that be a setting in the file (that if used in an older version will properly barf not knowing about the `allow-duplicates` setting.

Some tradeoffs made: settings are parsed in order and duplicates are only allowed after the setting is used. I know this breaks the nice unordered property of justfiles generally, but the compromise seemed fair while I was there.

It looks like this now:

```
set allow-duplicates

# default that might come from `justprep`
foo:
  @echo "basic recipe"

foo:
  @echo "more specific overridden recipe"
```

```shell
$ just
more specific overridden recipe
```

and without the `set allow-duplicates` line will error as normal:
```
error: Recipe `foo` first defined on line 3 is redefined on line 6
  |
6 | @foo:
  |  ^^^
```

Open to any feedback (both on approach and code) -- or if you want to go a different direction entirely at this point it would be understandable too.

I can also write up the docs as needed, but didn't want to go overboard until got your general approval to refine more

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,14 +519,14 @@ foo:
 
 #### Table of Settings
 
-| Name                      | Value              | Description                                                    |
-| ------------------------- | ------------------ | -------------------------------------------------------------- |
+| Name                      | Value              | Description                                                                                   |
+| ------------------------- | ------------------ | --------------------------------------------------------------------------------------------- |
 | `allow-duplicate-recipes` | boolean            | Allow recipes appearing later in a `justfile` to override earlier recipes with the same name. |
-| `dotenv-load`             | boolean            | Load a `.env` file, if present.                                |
-| `export`                  | boolean            | Export all variables as environment variables.                 |
-| `positional-arguments`    | boolean            | Pass positional arguments.                                     |
-| `shell`                   | `[COMMAND, ARGS…]` | Set the command used to invoke recipes and evaluate backticks. |
-| `windows-powershell`      | boolean            | Use PowerShell on Windows as default shell.                    |
+| `dotenv-load`             | boolean            | Load a `.env` file, if present.                                                               |
+| `export`                  | boolean            | Export all variables as environment variables.                                                |
+| `positional-arguments`    | boolean            | Pass positional arguments.                                                                    |
+| `shell`                   | `[COMMAND, ARGS…]` | Set the command used to invoke recipes and evaluate backticks.                                |
+| `windows-powershell`      | boolean            | Use PowerShell on Windows as default shell.                                                   |
 
 Boolean settings can be written as:
 
--- a/README.md
+++ b/README.md
@@ -519,13 +519,14 @@ foo:
 
 #### Table of Settings
 
-| Name                   | Value              | Description                                                    |
-| ---------------------- | ------------------ | -------------------------------------------------------------- |
-| `dotenv-load`          | boolean            | Load a `.env` file, if present.                                |
-| `export`               | boolean            | Export all variables as environment variables.                 |
-| `positional-arguments` | boolean            | Pass positional arguments.                                     |
-| `shell`                | `[COMMAND, ARGS…]` | Set the command used to invoke recipes and evaluate backticks. |
-| `windows-powershell`   | boolean            | Use PowerShell on Windows as default shell.                    |
+| Name                      | Value              | Description                                                    |
+| ------------------------- | ------------------ | -------------------------------------------------------------- |
+| `allow-duplicate-recipes` | boolean            | Allow duplicate recipes (last definition is used).             |
+| `dotenv-load`             | boolean            | Load a `.env` file, if present.                                |
+| `export`                  | boolean            | Export all variables as environment variables.                 |
+| `positional-arguments`    | boolean            | Pass positional arguments.                                     |
+| `shell`                   | `[COMMAND, ARGS…]` | Set the command used to invoke recipes and evaluate backticks. |
+| `windows-powershell`      | boolean            | Use PowerShell on Windows as default shell.                    |
 
 Boolean settings can be written as:
 
@@ -537,6 +538,25 @@ Which is equivalent to:
 
 ```mf
 set NAME := true
+```
+
+#### Allow Duplicate Recipes
+
+If `allow-duplicate-recipes` is `true`, recipes can be redefined, (the last definition will be used). Defaults to `false`.
+
+```make
+set allow-duplicate-recipes
+
+@foo:
+  echo foo
+
+@foo:
+  echo bar
+```
+
+```sh
+$ just foo
+bar
 ```
 
 #### Dotenv Load
--- a/README.md
+++ b/README.md
@@ -542,7 +542,7 @@ set NAME := true
 
 #### Allow Duplicate Recipes
 
-If `allow-duplicate-recipes` is `true`, recipes can be redefined, (the last definition will be used). Defaults to `false`.
+If `allow-duplicate-recipes` is `true`, recipes can be redefined and the last definition of duplicate recipes will be used. Defaults to `false`.
 
 ```make
 set allow-duplicate-recipes
--- a/README.md
+++ b/README.md
@@ -544,7 +544,7 @@ set NAME := true
 
 #### Allow Duplicate Recipes
 
-If `allow-duplicate-recipes` is `true`, defining multiple recipes with the same name is not an error, and the last definition is used. Defaults to `false`.
+If `allow-duplicate-recipes` is set to `true`, defining multiple recipes with the same name is not an error and the last definition is used. Defaults to `false`.
 
 ```make
 set allow-duplicate-recipes
--- a/README.md
+++ b/README.md
@@ -521,7 +521,7 @@ foo:
 
 | Name                      | Value              | Description                                                    |
 | ------------------------- | ------------------ | -------------------------------------------------------------- |
-| `allow-duplicate-recipes` | boolean            | Allow duplicate recipes (last definition is used).             |
+| `allow-duplicate-recipes` | boolean            | Allow recipes appearing later in a `justfile` to override earlier recipes with the same name. |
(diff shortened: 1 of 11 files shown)


casey:
Thanks for the PR and the thoughtful comments! 

I'm not opposed to a setting. A few comments:

- I think it should be called `allow-duplicate-recipes`, for clarity. `allow-duplicates` makes me think that maybe it also allows duplicate variables. (Which might be a desirable thing to add later, since `justprep` might want to also allow overriding variables.)

- The setting should be able to appear anywhere in the `justfile` and have the same effect. Otherwise, the docs have to say "All the settings are order independent, except for `allow-duplicate-recipes`…". I haven't thought about it in depth, but `Analyzer::recipes` can probably be turned into a `Table<'src, Vec<UnresolvedRecipe<'src>>>`, and the check for duplicates can be deferred until after all the recipes are analyzed, and depend on whether `allow-duplicate-recipes` was set or not.

- There should be an integration test that actually runs a recipe that was overridden.

lutostag:
Thanks for the quick response and feedback!

Went through and made the suggested changes.

Was able to implement the duplicate overriding from within the analyzer relatively cleanly (after reading the settings).

Not sure I added the integration test to the right spot, but put one into `tests/misc.rs`

Also added quick documentation to the Readme (but unsure if I missed other spots).

Any feedback/suggestions are welcome (even small nits). Thanks!

README.md
```diff
@@ -519,13 +519,14 @@ foo:
...
-| Name                   | Value              | Description                                                    |
-| ---------------------- | ------------------ | -------------------------------------------------------------- |
-| `dotenv-load`          | boolean            | Load a `.env` file, if present.                                |
-| `export`               | boolean            | Export all variables as environment variables.                 |
-| `positional-arguments` | boolean            | Pass positional arguments.                                     |
-| `shell`                | `[COMMAND, ARGS…]` | Set the command used to invoke recipes and evaluate backticks. |
-| `windows-powershell`   | boolean            | Use PowerShell on Windows as default shell.                    |
+| Name                      | Value              | Description                                                    |
+| ------------------------- | ------------------ | -------------------------------------------------------------- |
+| `allow-duplicate-recipes` | boolean            | Allow duplicate recipes (last definition is used).             |
```
casey, line 524:
```suggestion
| `allow-duplicate-recipes` | boolean            | Allow recipes to override earlier ones with the same name.             |
```

```suggestion
| `allow-duplicate-recipes` | boolean            | Allow recipes appearing later in a `justfile` to override earlier recipes with the same name. |
```

README.md
```diff
@@ -539,6 +540,25 @@ Which is equivalent to:
 set NAME := true
 ```
 
+#### Allow Duplicate Recipes
+
+If `allow-duplicate-recipes` is `true`, recipes can be redefined and the last definition of duplicate recipes will be used. Defaults to `false`.
```
```suggestion
If `allow-duplicate-recipes` is `true`, defining multiple recipes with the same name is not an error, and the last definition is used. Defaults to `false`.
```

src/analyzer.rs
```diff
@@ -82,6 +66,41 @@ impl<'src> Analyzer<'src> {
...
+      if let Some(original) = self.recipes.get(recipe.name.lexeme()) {
+        if settings.allow_duplicate_recipes {
+            self.recipes.remove(recipe.name.lexeme());
+        } else {
+          return Err(recipe.name.token().error(DuplicateRecipe {
+            recipe: original.name(),
+            first: original.line_number(),
+          }));
+        }
+      }
```
casey: Since insert will overwrite something already in the table, this can just be:

```suggestion
        if !settings.allow_duplicate_recipes {
          return Err(recipe.name.token().error(DuplicateRecipe {
            recipe: original.name(),
            first: original.line_number(),
          }));
        }
      }
```

src/analyzer.rs
```diff
@@ -102,12 +121,6 @@ impl<'src> Analyzer<'src> {
   }
 
   fn analyze_recipe(&self, recipe: &UnresolvedRecipe<'src>) -> CompileResult<'src, ()> {
-    if let Some(original) = self.recipes.get(recipe.name.lexeme()) {
-      return Err(recipe.name.token().error(DuplicateRecipe {
-        recipe: original.name(),
-        first: original.line_number(),
-      }));
-    }
```
```suggestion
```

tests/json.rs
```diff
@@ -243,6 +247,52 @@ fn dependency_argument() {
...
+
+#[test]
+fn duplicate_recipes() {
+  test(
+    "
+      set allow-duplicate-recipes
+      alias f := foo
+
+      foo:
+      foo:
```
casey: This test is slightly improved if the two `foo`s are different, so you can see the correct one in the output:

```suggestion
      foo:
      foo bar:
```

tests/misc.rs
```diff
@@ -1221,6 +1221,14 @@ test! {
...
 }
 
+test! {
+  name:     allow_duplicate_recipes,
+  justfile: "b:\n  echo foo\nb:\n  echo bar\nset allow-duplicate-recipes",
+  args:     ("b"),
+  stdout: "bar\n",
+  stderr: "echo bar\n",
+}
+
```
casey: I try not to add new tests to `misc.rs`, since it's way too long as is. You can create a new file, `tests/allow_duplicate_recipes.rs` and put the test in there. Also, I try not to add new macro-based tests, i.e., using the `test!` macro.  Check out one of the struct-based tests that uses `Test::new`.

casey (changes requested):
Great! Added a few comments.

tests/misc.rs
lutostag: Moved to new file as suggested.

lutostag:
I think I addressed your comments. As an extra thought however just noting that `allow-duplicate-recipes` I believe would be the first justfile codification of the word "recipe" to it. It lives throughout the documentation/code and in the unstable json dump area but would not be easy to change in a backwards compatible manner. Looking at https://github.com/casey/just/issues/166 particularly. It might actually make sense if we only enable `allow-duplicate-recipes` setting if we enable the `unstable` flag in the CLI.

I know that would be more work (and go back to pulling in config options into the analyzer which I tried to avoid... but would keep it behind the feature flag). Also because the only other code effected would be in the json dump code that is also unstable, we wouldn't need to pass the settings around everywhere. (Perhaps the tricky part would be documenting it clearly in the README).

@casey I don't mean to delay a feature I asked for I guess only want to make sure we bolt on new features wisely -- especially with 1.0 around the corner.

I'm fine either merging or doing the work to only allow this when `--unstable` is passed, up to you entirely but curious on your thoughts.

README.md
```diff
@@ -541,6 +542,25 @@ Which is equivalent to:
 set NAME := true
 ```
 
+#### Allow Duplicate Recipes
+
+If `allow-duplicate-recipes` is `true`, defining multiple recipes with the same name is not an error, and the last definition is used. Defaults to `false`.
```
```suggestion
If `allow-duplicate-recipes` is set to `true`, defining multiple recipes with the same name is not an error and the last definition is used. Defaults to `false`.
```

casey:
> I think I addressed your comments. As an extra thought however just noting that `allow-duplicate-recipes` I believe would be the first justfile codification of the word "recipe" to it. It lives throughout the documentation/code and in the unstable json dump area but would not be easy to change in a backwards compatible manner. Looking at #166 particularly.

Good thought! I'm actually not too worried. It wouldn't be hard to just soft-deprecate the current setting name, at the same time as we introduce an alias. I'm also on the fence about #166 anyways.

Regarding 1.0, I actually think that anything that we want to change in the future can be introduced in a backwards compatible way with settings. And, if eventually there are a bunch of settings that we want to make the default, we can do so with something like Rust's edition system, where you would put something like `set version := 2` at the top of the justfile, which would opt you into a new set of defaults.

So I'm fine including this as is and worrying about it later if we want to change it.

casey:
There was a clippy error, but it was simple so I went ahead and fixed it. Looks good!

casey (approved):

casey:
I just published 0.11.1, which includes this change.

PR #1106, opened 2022-02-17
Path_exists() function

### LINKED ISSUE #1102: `path_exists()` Path manipulation function
Hi! Do you think it would be beneficial to add `path_exists()` to a path manipulation function assortment?

Edit: I added that function, I will submit a PR tomorrow.

heavelock:
Hey! 
So here is promised PR for `path_exists()`. 
I also added tests for dirs and files, should I also add tests for hard/soft links? And sorry for their general clumsiness, I'm writing rust basically for the first time and I still have to wrap my head around it.

Closes: #1102


tests/functions.rs
```diff
@@ -382,3 +382,69 @@ fn join_argument_count_error() {
...
     .run();
 }
+
+#[test]
+fn test_path_exists_filepath_exist() {
+  let test = Test::new();
+  let testfilepathbuf = test.tempdir.path().join("testfile.txt");
+  let testfilepath = testfilepathbuf.as_path();
+  let _file = std::fs::File::create(testfilepath);
+  test
```
casey: You can use `Test::tree` to populate the tempdir the test runs in with files:

```suggestion
  Test::new()
    .tree(tree! {
      testfile: ""
    })
```

tests/functions.rs
```diff
@@ -382,3 +382,69 @@ fn join_argument_count_error() {
...
+    .justfile(format!(
+      "x := path_exists(\"{}\")",
+      testfilepath.to_str().unwrap()
+    ))
```
casey: You should just be able to do:

```suggestion
    .justfile("x := path_exists('testfile')")
```

tests/functions.rs
```diff
@@ -382,3 +382,69 @@ fn join_argument_count_error() {
...
+    .args(&["--evaluate", "x"])
+    .stdout("true")
+    .unindent_stdout(false)
```
casey:
```suggestion
    .stdout("true\n")
```

tests/functions.rs
```diff
@@ -382,3 +382,69 @@ fn join_argument_count_error() {
...
+    .run();
```
casey, line 395: These comments also apply for the rest of the tests.

casey (changes requested):
> Hey! So here is promised PR for `path_exists()`. I also added tests for dirs and files, should I also add tests for hard/soft links?

Since we're using `Path::exists`, which is provided to us, we don't need fully comprehensive tests. In general, I think it's safe to assume that you're dependencies are correct, especially when it comes from the standard library. So for this, I think we can just have one test when a file exists, and another when it doesn't.

> And sorry for their general clumsiness, I'm writing rust basically for the first time and I still have to wrap my head around it.

No worries, rust is hard, and the way just runs tests is ideosyncratic!

heavelock:
Thanks for your comments! I applied your suggestions and removed the tests for directories since we don't necessarily need to test for them.

Should I somehow refer in docs that this function is using internally rust's std behaviour?

casey:
I think it probably isn't necessary to reference the stdlib function. I looked at the documentation for Path::exists, and we could probably include some of the details. I also think it's probably a good idea to put the function under a more descriptive heading, maybe, so something like:


```
#### Filesystem Access

- `path_exists(path)` - Returns `true` if the path points at an existing entity and `false` otherwise. Traverses symbolic links, and returns `false` if the path is inaccessible or points to a broken symlink.
```

heavelock:
I changed the readme entry according to your suggestion.

casey:
Nice, merged! Thank you for the PR! I think there are probably a lot of other useful boolean functions that can be added now.

Issue #945, opened 2021-08-12
Add settings to control environment file name


casey:
Something like:

```
set dotenv-filenames := [".env", "other-dotenv-file"]
```

Celeo:
This should be like @xorander00 explained in #941, right?

> have a setting to specify the file(s) to load and the order in which to load them ... look for and load from left-to-right (later loaded dotenv files overriding values from previously loaded ones).

And if one isn't found, silently (?) skip it?

casey:
Yeah, that's right. If a user wants it to be an error if an environment file isn't found, we could add another setting, like `set dotenv-require := true`, which would default to false, and make `just` return an error if an environment file isn't found

xorander00:
Just saw this, sweet!  :) Will build from latest commit & give it a shot this weekend and report back if I run into any issues.  
  
@casey I think that would be a good setting, yes.  Although personally, I'd consider making it a numeric (or enum-like) state flag instead of a boolean toggle.  Not sure how it would work in just, but something like `dotenv-missing := 0 | 1 | 2`.  0 being silently ignore and continue, 1 being warn and continue (default), and 2 being fail and exit.  This offers more flexibility for the user as well as for the future (additional cases can just be new numbers).  Just my personal preference though and not a huge deal.

casey:
> @casey I think that would be a good setting, yes. Although personally, I'd consider making it a numeric (or enum-like) state flag instead of a boolean toggle. Not sure how it would work in just, but something like `dotenv-missing := 0 | 1 | 2`. 0 being silently ignore and continue, 1 being warn and continue (default), and 2 being fail and exit. This offers more flexibility for the user as well as for the future (additional cases can just be new numbers). Just my personal preference though and not a huge deal.

I think that would be reasonable. And it can always be a keyword, like `dotenv-missing := continue | warn | error`, so it's easier to remember.

hustcer:
@casey Can we change .env file in recipe? Like this:
```bash
# Start application in dev mode with .dev-env supported
start-dev:
  set dotenv-filenames := [".env", ".dev-env"]

# Start application in test mode with .test-env loaded
start-test:
  set dotenv-filenames := [".env", ".test-env"]
```

casey:
> @casey Can we change .env file in recipe? Like this:

I think it's probably impossible to change a `.env` file once a recipe runs. Different parts of a justfile could read from different environment files, depending on when they ran, which would be super confusing.

hustcer:
[...]

PR #1111, opened 2022-02-22
Draft: Custom dotenv names

heavelock:
Hey!

I started working on possibility of defining multiple `.env` files to be loaded with just. I basically followed an idea of that issue:
https://github.com/casey/just/issues/945
and added an array for storing filenames.

What do you think about that approach? 

It's still WIP, tests are still broken and the default value is not properly defined but I wanted to ask for your opinion before I sink more time into that.


casey:
In general, it's a bit hard to review something without tests. With tests, I can look at the new tests, and see what they're doing, which lets me answer the most important question: "Does this change have the right semantics, and is it working?". The code itself is a bit secondary, and can be fixed in review. I think the best way to avoid wasting time is actually to open a PR with new tests that are failing, and no actual implementation. That way the reviewer can confirm that what is being done is correct, and then the author can fill in the implementation.

peterjmorgan:
> In general, it's a bit hard to review something without tests. With tests, I can look at the new tests, and see what they're doing, which lets me answer the most important question: "Does this change have the right semantics, and is it working?". The code itself is a bit secondary, and can be fixed in review. I think the best way to avoid wasting time is actually to open a PR with new tests that are failing, and no actual implementation. That way the reviewer can confirm that what is being done is correct, and then the author can fill in the implementation.

I'm a huge fan of this tool, and reading your comment gives me a better perspective of how OSS maintainers think. This makes a ton of sense after reading, thank you!

casey:

PR #1116, opened 2022-02-23
Improve --init/--fmt compatibility (#1115)

### LINKED ISSUE #1115: Compatibility between `--init` & `--fmt`
The file produced by `--init` should generate a justfile which is formatted as expected by `--fmt`. Currently it is not.

```
❯ just --init
Wrote justfile to `/tmp/justfile`

❯ just --check --unstable --fmt
 default:
-       echo 'Hello, world!'
+    echo 'Hello, world!'
error: Formatted justfile differs from original.

TheLocehiliosan:
The justfile produced by `--init` should be formatted as expected by `--fmt`.

This fixes #1115.

### DIFF
--- a/src/subcommand.rs
+++ b/src/subcommand.rs
@@ -1,6 +1,6 @@
 use crate::common::*;
 
-const INIT_JUSTFILE: &str = "default:\n\techo 'Hello, world!'\n";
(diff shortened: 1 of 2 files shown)


casey:
Nice, great catch, and thanks for adding a test! ez merge :)

PR #1121, opened 2022-02-27
Add `absolute_path` function

Laura7089:
I asked on the discord and was told there was no function to get the absolute path from a relative path, so here one is!

I haven't written tests for this, please let me know what form the should take.


src/function.rs
```diff
@@ -59,6 +60,14 @@ impl Function {
   }
 }
 
+fn abs_path(context: &FunctionContext, path: &str) -> Result<String, String> {
+  let corrected_path = context.invocation_directory.join(path);
```
Laura7089: Is this necessary? I saw that `path_exists` is just called directly on the path var.

casey: Good point, `path_exists` has a bug, it needs to be made relative to `search.working_directory`. This should also be joined with search.working_directory.

src/function.rs
```diff
@@ -59,6 +60,14 @@ impl Function {
...
+  match corrected_path.canonicalize() {
```
casey: Call `.lexiclean()` instead of canonicalize. The paths that canonicalize produces are a bit weird.

src/function.rs
```diff
@@ -59,6 +60,14 @@ impl Function {
...
+    Ok(p) => Ok(p.to_string_lossy().into()),
```
casey: Calling to_string_lossy will produce an unusable path, if it isn't valid unicode. I would first convert search.working_directory to unicode:

```
  search.working_directory.to_str().ok_or_else(|| {
    format!(
      "Working directory path is not valid unicode: {}",
      search.working_directory.to_string_lossy()
    )
  })
```

And then turn it into a Utf8Path and join it with `path`

src/function.rs
Laura7089: Changed!

src/function.rs
Laura7089: Changed!

src/function.rs
Laura7089: I've refactored a little to include this - let me know if it's appropriate.

src/function.rs
```diff
@@ -59,6 +60,17 @@ impl Function {
...
 
+fn abs_path(ctx: &FunctionContext, path: &str) -> Result<String, String> {
+  let abs_path_unchecked = ctx.search.working_directory.join(path).lexiclean();
+  match abs_path_unchecked.to_str() {
+    Some(wd) => Ok(wd.to_string()),
+    None => Err(format!(
+      "Directory path is not valid unicode: {}",
+      abs_path_unchecked.to_string_lossy()
+    )),
+  }
```
casey: If there's an error, it will be due to the working directory not being valid unicode, so I think it's better if the error message only mentions the working directory, like so:

```suggestion
  Ok(
    Utf8Path::new(context.search.working_directory.to_str().ok_or_else(|| {
      format!(
        "Working directory is not valid unicode: {}",
        context.search.working_directory.display()
      )
    }))
    .join(path)
    .lexiclean()
    .to_string()
  )
```

src/function.rs
```diff
@@ -59,6 +60,17 @@ impl Function {
   }
 }
 
+fn abs_path(ctx: &FunctionContext, path: &str) -> Result<String, String> {
```
casey: I think `absolute_path` is a little clearer, and the context should be named `context` to match the other functions.

```suggestion
fn absolute_path(context: &FunctionContext, path: &str) -> Result<String, String> {
```

casey: To match the other functions.

README.md
```diff
@@ -988,6 +988,8 @@ The executable is at: /bin/just
 
 ##### Fallible
 
+- `abs_path(path)` - Absolute path to relative `path` in the invocation directory. `abs_path("./bar.txt")` in directory `/foo` is `/foo/bar.txt`.
```
```suggestion
- `absolute_path(path)` - Absolute path to relative `path` in the invocation directory. `absolute_path("./bar.txt")` in directory `/foo` is `/foo/bar.txt`.
```

casey (changes requested):
Some comments, also this needs tests. New tests should be written using the `Test` struct, and not the `test!` macro.  There are some examples in `tests/functions.rs`, where tests for this function could also go.

src/function.rs
Laura7089: Is `LexiClean` implemented for `Utf8Path` or `Utf8PathBuf`?

casey: Ah, no, it isn't, good point. I'll add an implementation soon. For now you can do something like:

```rust
if context.search.working_directory.to_str().is_none() {
      return Err(format!(
        "Working directory is not valid unicode: {}",
        context.search.working_directory.display()
      ));
}

Ok(
    context.search.working_directory
    .join(path)
    .lexiclean()
    .to_str()
    .unwrap()
    .to_owned()
)
```

Laura7089: Is there any concern worth having over the doubled-up call to `to_str()`?

Laura7089: (I've updated the PR to use a match statement which reports the working directory as being incorrect if the final join is bad)

src/function.rs
```diff
@@ -59,6 +60,17 @@ impl Function {
   }
 }
 
+fn absolute_path(context: &FunctionContext, path: &str) -> Result<String, String> {
+  let abs_path_unchecked = context.search.working_directory.join(path).lexiclean();
+  match abs_path_unchecked.to_str() {
+    Some(wd) => Ok(Utf8Path::new(wd).to_string()),
```
casey: We can just call `to_owned` here, `Utf8Path::new` is a noop:

```suggestion
    Some(absolute_path) => Ok(absolute_path.to_owned()),
```

tests/functions.rs
```diff
@@ -403,3 +403,37 @@ fn test_path_exists_filepath_doesnt_exist() {
     .stdout("false")
     .run();
 }
+
+#[test]
+fn test_absolute_path_resolves() {
+  let test_object = Test::new()
```
casey, line 409: Take a look at other tests that use the `tree` function to populate the tmpdir with files. You should be able to use that to do everything in one expression.

src/function.rs
casey: Calling `to_str()` twice shouldn't be an issue. A `just` process is going to spend the vast vast majority of its time executing recipe shell commands. Rust is already fast, and parsing, lexing, and doing local memory allocations are orders of magnitude faster than I/O, so I/O (launching external processes, reading their output, etc.) is going to be the bottleneck. So things like calling `to_str` are unlikely to make any difference in the perceived speed of `just` invocations.

That all being said, the updated version which does `to_str` once, but then only includes the current directory in the error message is fine too.

casey (changes requested):
Getting there!

tests/functions.rs
Laura7089: I'm not quite sure how to do that - could you explain how to add the `Tree` object to the `Test` then reference it in the `stdout()` builder call without getting borrow checker errors?

casey: Actually, looking at this again, I think that this would be tricky to do the way I had in mind, I think it's fine as-is.

tests/functions.rs
```diff
@@ -404,6 +404,41 @@ fn test_path_exists_filepath_doesnt_exist() {
...
+    .stdout(
+      tempdir
+        .parent()
+        .unwrap()
+        .join("test_file")
+        .to_str()
+        .unwrap()
+        .to_owned(),
+    )
+    master
```
casey: Whoops! Added this while fixing a merge conflict.

```suggestion
```

casey (approved):

casey:
For some reason the CI job didn't start, so I pushed an empty commit. This should merge soon. If this is functionality that you need, I'm happy to cut a release, just let me know.

Laura7089:
I'm in no desperate need :)

This has been my first contribution to a FOSS project, so thanks for giving me the opportunity and helping me through it!

casey:
Nice! Very glad to hear it!

PR #1152, opened 2022-04-08
Add cross-platform example

presidento:
Sometimes the file structure and the executables are very different in Windows and Unix operating systems. With this example it is possible to have one identical `justfile`. (If the same shell setting can be used in every system.)

I tried some other alternatives, but this one seemed to be the working one. I missed this information from the documentation. Just is cross-platform, but writing cross-platform `justfile` is another thing.

PS: I really love `just`, I use it in all of my pet projects on Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -893,7 +893,7 @@ $ just system-info
 This is an x86_64 machine
 ```
 
-`os_family()` can be used to create cross-platform `justfile` if the project has contributors using various operating systems. For example see  [cross-platform.just](examples/cross-platform.just) file.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -893,6 +893,8 @@ $ just system-info
 This is an x86_64 machine
 ```
 
+`os_family()` can be used to create cross-platform `justfile` if the project has contributors using various operating systems. For example see  [cross-platform.just](examples/cross-platform.just) file.
```
```suggestion
The `os_family()` function can be used to create cross-platform `justfile`s that work on various operating systems. For an example, see [cross-platform.just](examples/cross-platform.just) file.
```

casey:
Awesome, great example! Merging. And glad you like just! For my own pet projects is why I created it in the first place!

Issue #1018, opened 2021-11-04
Add `uuid()` function


casey:
This came up in #1011. @page-down, what is your use case for a `uuid()` function? Would it return the same thing every time in a given invocation, or would it return a new UUID every time? Would something that generated a hex string of a certain length be more or less useful, or is there a specific advantage to UUIDs?

page-down:
It is true that this suggestion is somewhat limited and should not be restricted to uuid. To be more useful, this function should actually be a pseudo-random number, pseudo-random string generator. I suggest changing the title of this issue.

The usage scenario is to generate a unique identifier during this execution to be shared by all processes executed by recipes.

For UUIDs, there should be at least standard format, hex format output. And at least UUID v1, UUID v4 versions should be available.

Each execution should output a new value, and the user should save the value to a variable.

The UUID is proposed because this is the format with a uniform standard. In other cases, the random string length, whether it covers numbers, alphabets, symbols, can be decided by the user by parameters.

Justfile can be made more cross-platform portable if useful tool functions are added without making just bloated.

runeimp:
This seems like an odd request to me and would be a rare use case. Something that should just be a call to a command line tool which is baked-in or easily added to any OS that `just` runs on. Or wait until `just` implements Python or something.

page-down:
> Something that should just be a call to a command line tool which is baked-in or easily added to any OS that just runs on.

Just like the functions that are currently available, they can actually be done with external tools. And `just` is not possible to provide all the utils that are similar to the standard library in a programming language.

I think to evaluate whether a feature is worth adding, you should judge whether it is generic enough, or it can solve some cross-platform problems, or it can improve script performance by reduce the overhead of starting processes.

If this function can unify the generation of pseudo-random strings across different platforms, can help users write scripts in a concise and clear manner, and is executed very efficiently, then it is good.

One of the core aspects of `just` is to help users efficiently execute common commands consisting of various tools and software. `just` can't go out and implement various tools for users. It is endless.

runeimp:
[...]

Issue #1011, opened 2021-10-29
Consider ability to reload .env files between recipes


tomasol:
Maybe there is a better way, but I would like to modify the `.env` file from within a recipe and use it afterwards.
The only working solution I found was following:
```
modify:
    cp .env.template .env
    echo "TIME=$(date +%s)" >> .env
    cat .env

read:
    echo "Expected: $(grep TIME .env)"
    echo $TIME

# could this just be "all: modify read"?
all: modify
    bash -c 'unset TIME && just --dotenv-filename .env read'
```
Ideally there would be a flag to auto reload .env files between recipes. Right now they are read at startup and even calling `just` recursively does not work as the old environment variables will have precedence before the .env file.

casey:
Would it be better if there was some way to modify variables without modifying the `.env` file?

tomasol:
Thanks for quick reply. I would like to be able to run the recipes one by one, which works now, but also together. I need the persistence to keep an execution ID around. Can it be done without modifying the `.env` file?

casey:
> Thanks for quick reply. I would like to be able to run the recipes one by one, which works now, but also together. I need the persistence to keep an execution ID around. Can it be done without modifying the `.env` file?

I see, so persistence is a part of it.

You could do something like this:

```
id := `cat .id`

init:
  echo xyz > .id
```

You would have to run `just init` first, to initialize `.id`, but then on subsequent runs, you'd have it in the `id` variable.

Does that work?

tomasol:
Not really. The solution works same way as an env file - it works if recipes are invoked separately, but breaks when invoked in a single run:
```
time := `cat .time`

init:
  echo $(date +%s) > .time
  cat .time

dep: init
  echo {{time}}
```

```sh
$ touch .time
$ just dep
echo $(date +%s) > .time
cat .time
1635542204
echo

$ just dep
echo $(date +%s) > .time
cat .time
1635542212
echo 1635542204
1635542204
```
What I really want is to generate and pass a persistent variable that works on separate recipes and on a chain.

casey:
Yeah, that's tough. I can't think of a great way to do this, but I'll leave the issue open in case anyone thinks of something clever.

page-down:
> ... I would like to modify the .env file from within a recipe and use it afterwards.
> ... I need the persistence to keep an execution ID around. Can it be done without modifying the .env file?
> ... generate and pass a persistent variable that works on separate recipes and on a chain.

[...]

PR #1170, opened 2022-05-01
UUID and sha256 digest functions

mbodmer:
issue #1018 mentioned a uuid function would be helpful.

Here is my attempt to deliver this. Also I've included sha256 hashing functions for strings and files.

To create a meaningful unittest for uuid() I had to refactor the Test struct in test.rs, I hope this is acceptable. The challenge was to keep the current interface but allow for custom verification. Well, here is my attempt. I'm new to rust, so bear with me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1032,9 +1032,9 @@ These functions can fail, for example if a path does not have an extension, whic
 
 #### UUID and Hash Generation
 
+- `sha256(textdata)` Generates a sha256 hash of the given text data. Example: `sha256("just hello")`
+- `sha256_file(path)` Generates a sha256 hash of the given file. Example: `sha256_file("/path/to/file")`
 - `uuid()` - Randomly generates a hyphenated UUID string.
-- `digest(textdata)` Generates a sha256 hash of the given text data. Example: `digest("just hello")`
-- `digest_file(path)` Generates a sha256 hash of the given file. Example: `digest_file("/path/to/file")`
 
 ### Command Evaluation Using Backticks
 
--- a/README.md
+++ b/README.md
@@ -1032,9 +1032,9 @@ These functions can fail, for example if a path does not have an extension, whic
 
 #### UUID and Hash Generation
 
-- `sha256(textdata)` Generates a sha256 hash of the given text data. Example: `sha256("just hello")`
-- `sha256_file(path)` Generates a sha256 hash of the given file. Example: `sha256_file("/path/to/file")`
-- `uuid()` - Randomly generates a hyphenated UUID string.
+- `sha256(string)` - Return the SHA-256 hash of `string` as a hexadecimal string.
+- `sha256_file(path)` - Return the SHA-256 hash of the file at `path` as a hexadecimal string.
+- `uuid()` - Return a randomly generated UUID.
 
 ### Command Evaluation Using Backticks
 
--- a/README.md
+++ b/README.md
@@ -1030,7 +1030,7 @@ These functions can fail, for example if a path does not have an extension, whic
 
 - `error(message)` - Abort execution and report error `message` to user.
 
-#### UUID an Hash Generation
+#### UUID and Hash Generation
 
 - `uuid()` - Randomly generates a hyphenated UUID string.
 - `digest(textdata)` Generates a sha256 hash of the given text data. Example: `digest("just hello")`
--- a/README.md
+++ b/README.md
@@ -1030,6 +1030,12 @@ These functions can fail, for example if a path does not have an extension, whic
 
 - `error(message)` - Abort execution and report error `message` to user.
 
+#### UUID an Hash Generation
+
+- `uuid()` - Randomly generates a hyphenated UUID string.
+- `digest(textdata)` Generates a sha256 hash of the given text data. Example: `digest("just hello")`
+- `digest_file(path)` Generates a sha256 hash of the given file. Example: `digest_file("/path/to/file")`
+
 ### Command Evaluation Using Backticks
 
 Backticks can be used to store the result of commands:
--- a/src/function.rs
+++ b/src/function.rs
@@ -17,6 +17,8 @@ lazy_static! {
     ("absolute_path", Unary(absolute_path)),
     ("arch", Nullary(arch)),
     ("clean", Unary(clean)),
+    ("digest", Unary(digest)),
+    ("digest_file", Unary(digest_file)),
     ("env_var", Unary(env_var)),
     ("env_var_or_default", Binary(env_var_or_default)),
     ("error", Unary(error)),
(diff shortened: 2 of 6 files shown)


mbodmer:
#1011

casey:
Thanks for this PR! There's already `Test::stderr_regex`, and I think it might be cleaner to add a `Test::stdout_regex` to test that the output matches a pattern, instead of adding a new kind of validator. for UUIDs, it should be okay to check that it matches the pattern `/........-....-....-....-............/`

For the hash functions, can you use the implementation from https://github.com/RustCrypto/hashes? I think it's better maintained.

mbodmer:
Hi, thanks for your review. I have now added stdout_regex in the sense of the existing stderr_regex. I agree, much less complexity that way. Also I have replaced the hashing dependency with rust-crypto. Hope this helps.

mbodmer:
I'm not sure whether the general name digest()/digest_file() for the functions is too broad. Any requests for a better more concise name? :-)
eg: sha256()/sha256_file()

casey:
> I'm not sure whether the general name digest()/digest_file() for the functions is too broad. Any requests for a better more concise name? :-) eg: sha256()/sha256_file()

I think I agree, `sha256()` and `sha256_file()` are probably better, so if you're reading a justfile and you don't know the function you can see what it does without having to look up documentation.

mbodmer:
Ok, they are now renamed to the more specific `sha256()`, `sha256_file()`

README.md
```diff
@@ -1016,6 +1016,12 @@ These functions can fail, for example if a path does not have an extension, whic
 
 - `error(message)` - Abort execution and report error `message` to user.
 
+#### UUID and Hash Generation
+
+- `sha256(textdata)` Generates a sha256 hash of the given text data. Example: `sha256("just hello")`
```
```suggestion
- `sha256(string)` Return the SHA-256 hash of `string` as a hexadecimal string.
```

README.md
```diff
@@ -1016,6 +1016,12 @@ These functions can fail, for example if a path does not have an extension, whic
...
+- `sha256_file(path)` Generates a sha256 hash of the given file. Example: `sha256_file("/path/to/file")`
```
```suggestion
- `sha256_file(path)` Return the SHA-256 hash of the file at `path` as a hexadecimal string.
```

README.md
```diff
@@ -1016,6 +1016,12 @@ These functions can fail, for example if a path does not have an extension, whic
...
+- `uuid()` - Randomly generates a hyphenated UUID string.
```
```suggestion
- `uuid()` - Return a randomly generated UUID.
```

src/function.rs
```diff
@@ -247,6 +250,23 @@ fn replace(_context: &FunctionContext, s: &str, from: &str, to: &str) -> Result<
...
+  hasher.update(s);
+  let hash = hasher.finalize();
+  Ok(format!("{:x}", hash))
+}
+
+fn sha256_file(_context: &FunctionContext, path: &str) -> Result<String, String> {
+  use sha2::{Digest, Sha256};
+  let mut hasher = Sha256::new();
+  let mut file = std::fs::File::open(path).unwrap();
+  std::io::copy(&mut file, &mut hasher).unwrap();
```
casey: These unwraps should both be handled. Something like:

```suggestion
  let mut file = std::fs::File::open(path).map_err(|err| format!("Failed to open file at `{}`: {}", path, err))?;
  std::io::copy(&mut file, &mut hasher).map_err(|err| format!("Failed to read file at `{}`: {}", path, err)?;
```

tests/functions.rs
```diff
@@ -476,3 +476,36 @@ fn path_exists_subdir() {
...
+    .justfile("x := sha256('5943ee37-0000-1000-8000-010203040506')")
+    .args(&["--evaluate", "x"])
+    .stdout("2330d7f5eb94a820b54fed59a8eced236f80b633a504289c030b6a65aef58871")
+    .run();
+}
+
+#[test]
+fn sha256_file() {
+  let mut tmpfile = tempfile::NamedTempFile::new().unwrap();
+  writeln!(tmpfile, "just is great").unwrap();
```
casey: ❤️

tests/functions.rs
```diff
@@ -476,3 +476,36 @@ fn path_exists_subdir() {
...
+  let tmppath = tmpfile.into_temp_path();
+
+  let justfile_content = format!("x := sha256_file('{}')", tmppath.to_str().unwrap());
+
+  Test::new()
+    .justfile(justfile_content)
```
casey: You can use the somewhat odd `tree` function to populate the tempdir with files:

```suggestion
    .justfile("x := sha256_file('file')")
    .tree(tree! {
      file: "just is great",
    })
```

casey (changes requested):
Nice, looking good. I added a few minor comments.

mbodmer:
Thanks for the extensive review, I think I've covered all your issues now, feel free to check again

casey (changes requested):
Nice, looks good, although I just realized that there's a bug. `just` doesn't change the working directory, so you'll need to join the `path` in `sha256_file` with `context.search.working_directory`, i.e. `context.search.working_directory.join(path)`.

Could you also add a test for this? It should fail on the current PR, and then pass once `path` is joined with the working directory. You can use `Test::current_dir` to make just start in a sub-directory of the tmpdir. You can try to do `sha256_file` on a file in tempdir.

mbodmer:
Sorry for force pushing, I've had to rewrite the history to rebase the branch, since you've also modified the lockfile and stupid me checked it in, in a commit together with other changes.
Can you give me more details, I don't understand where, how and what causes the bug you've mentioned.

casey:
Lets say you have the following files:

```
/justfile
/bar
/foo/
```

And you're currently in `/foo/`. `/justfile` contains:

```
baz:
  echo {{sha256_file("bar")}}}
```

You do:

```
$ just
```

The `just` executable looks for the justfile and sees that it's in `/`, so it sets `search.working_directory` to `/`. However, it does not call `env::set_current_dir` with `/`, so the current directory is still `/bar`. So then `sha256_file("bar")` will fail, because there is no `/foo/bar`.

mbodmer:
Ok, I think now I understand, thanks. Will try

src/function.rs
```diff
@@ -247,6 +250,26 @@ fn replace(_context: &FunctionContext, s: &str, from: &str, to: &str) -> Result<
...
+  use sha2::{Digest, Sha256};
+  let mut hasher = Sha256::new();
+  hasher.update(s);
+  let hash = hasher.finalize();
+  Ok(format!("{:x}", hash))
+}
+
+fn sha256_file(_context: &FunctionContext, path: &str) -> Result<String, String> {
+  use sha2::{Digest, Sha256};
+  let justpath = _context.search.working_directory.join(path);
```
casey: The `_` can be removed now that we're using context. The `_` is just to make rust not complain about an unused variable.

```suggestion
fn sha256_file(context: &FunctionContext, path: &str) -> Result<String, String> {
  use sha2::{Digest, Sha256};
  let justpath = context.search.working_directory.join(path);
```

casey (approved):
Looks good! I went ahead and removed a `_` that is now unnecessary.

mbodmer:
Thanks

PR #1205, opened 2022-05-29
Translate part of README.md to zh-CN based on commit 95de4eb

hustcer:
Translate part of README.md to zh-CN based on commit 95de4eb

I will fix all the links after the translation finished.


casey:
Nice, merged!

PR #1206, opened 2022-05-30
Translate editor support and quick start to zh-CN based on c87909c

hustcer:
Translate editor support and quick start to zh-CN based on commit c87909c


casey:
Nice, thank you for catching the typo!

PR #1208, opened 2022-06-01
Translate part of features to Chinese based on commit c87909c

hustcer:
Translate part of features to Chinese
It's really a long chapter


casey:
真牛

hustcer:
> invocation_directory() - Retrieves the absolute path to the current directory when just was invoked, before just changed it (chdir'd) prior to executing commands.

I'm not quite sure about this part "before just changed it (chdir'd) prior to executing commands" can you explain it with more detail?

casey:
> I'm not quite sure about this part "before just changed it (chdir'd) prior to executing commands" can you explain it with more detail?

Yeah, definitely. So initially, the user invokes just in some directory, let's say `~/src/project/foo`. That's the initial working directory. Then just looks for the justfile, let's say it's in `~/src/project/justfile`, and changes the working directory to the directory that contains the justfile, in this case `~/src/project`. `invocation_directory()` returns the directory that just was invoked in, in this case ~/src/project/foo`.

hustcer:
Ah, I see, thanks

PR #1226, opened 2022-06-12
Self update support

mihaigalos:
This PR contributes the capability to self-update the `just` binary in-place.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


mihaigalos:
Hi maintainers, please have a look and provide feedback if this feature is desired and how it can be improved.
The current limitation is that it hard-codes updates to `x86_64-unknown-linux-musl` only but we can improve it later.

Also, if we go ahead with it, please advise how I can contribute tests here.
It's a bit tricky since the binary itself needs to be called, self-update and tested for the newer version.

casey:
This is very cool! What happens if `just` is installed as a system package, e.g. with homebrew or another package manager? Will `just --update` overwrite the installed system package?

mihaigalos:
> This is very cool! What happens if `just` is installed as a system package, e.g. with homebrew or another package manager? Will `just --update` overwrite the installed system package?

Yes, the binary will be overwritten directly. I am not sure if metadata is updated in the package manager (any way to programmatically update it?).
In that case, I wouldn't expect users to manually call `just --update`, but if they to, the metadata would be out-of-sync.

casey:
I'm on the fence about this!

On the one hand, it's cool and convenient. On the other hand, the current install script is pretty easy to use, and won't work when just is installed with a package manager.

I checked out the PR branch and did a little testing, and `self_update` is a pretty heavy dependency. It pulls in reqwest, tokio, and hyper, which are pretty big. This brings clean compile times on my machine from around 40 seconds to around 70 seconds, and incremental compiles from around 0.5 seconds to around 1.0 seconds. The clean compile time increase doesn't bother me that much, but the incremental compile time increase doesn't seem worth it. Especially since my machine is pretty fast, so it'll be worse on slower machines.

So I think on balance this isn't worth it, mostly because of the increase in compile times. Thank you for this PR though!

PR #1240, opened 2022-06-20
Fix dead link "much more"

wdroz:
the link https://just.systems/man/ is 404, so I changed it to https://just.systems/man/en/


casey:
Awesome, thank you!

PR #1246, opened 2022-06-21
restore fuzz harness

evanrichter:
the current fuzz harness is out of date and wouldn't build, so I updated it back to working.

I used the normal `cargo fuzz init` template, then changed some of the crate visibility rules gated by `cfg(fuzzing)`

now, for fuzzing, you can just use [cargo fuzz](https://github.com/rust-fuzz/cargo-fuzz) from the top directory: `cargo fuzz run compile` ("compile" is the name of the harness, because the harness compiles a JustFile from plaintext)


casey:
Nice, looks great!

casey:
Whoops, meant to merge, not close 😅

Issue #1251, opened 2022-06-25
Opaque error when conditional is missing an `else` block


nk9:
## Problem
Currently, when a `justfile` has this code:

```
TEST := if path_exists('/bin/bash') == "true" {"yes"}
```

The error is this:

```
error: Expected identifier, but found end of line
```

This doesn't say which identifier was expected, which makes it difficult to understand.

## Diagnosis
The conditional is parsed in [`parse_conditional`](https://github.com/casey/just/blob/a46be416991b83a16028becf724d4b382fd9559d/src/parser.rs#L429). The `else` is expected [here](https://github.com/casey/just/blob/a46be416991b83a16028becf724d4b382fd9559d/src/parser.rs#L449). This [`expect_keyword`](https://github.com/casey/just/blob/a46be416991b83a16028becf724d4b382fd9559d/src/parser.rs#L159) function looks like it is set up to provide a useful error:

```
    if expected == found {
      Ok(())
    } else {
      Err(identifier.error(CompileErrorKind::ExpectedKeyword {
        expected: vec![expected],
        found,
      }))
    }
```

However, this code is never run because the call to `self.expect()` on the first line of the function produces [its own, more generic error](https://github.com/casey/just/blob/a46be416991b83a16028becf724d4b382fd9559d/src/parser.rs#L144).

I don't have a good feel for what the right solution here is. Perhaps there really are situations where the `else` block of `expect_keyword` is actually being reached? I'm happy to help, but could use some direction at this point.

casey:
This is a dupe of #1236, but closing that in favor of this, because this issue has a better description.

I think that avoiding calling `self.expect(Identifier)?` would be good, and instead examining the next token manually to see if it's both an identifier and the expected keyword.

PR #1252, opened 2022-06-26
Useful error when conditional is missing `else`

nk9:
To address #1251

I modeled this on the following method, `presume_keyword()`. The tests are all succeeding, and I've added a new one.

One question: is it desired that the error says ``but found identifier ` `` and just shows the rendered newline? I'd think either `\n` or just a straight up `but found a newline` would be a bit clearer.


casey:
Nice! Thanks for the PR.

I think the best option is to change `found` to a `TokenKind`, which has a good display implementation.

There should also be a test which checks the error message when an `if` clause is followed by the end of the file.

nk9:
Thanks! I spent some time investigating that, and here's where I got to:

```rust
    let next = self.advance()?;
    let found = next.lexeme();

    if next.kind != Identifier {
      Err(next.error(CompileErrorKind::ExpectedKeyword {
        expected: vec![expected],
        found: next.kind,
      }))
    } else if expected != found {
      Err(next.error(CompileErrorKind::ExpectedKeyword {
        expected: vec![expected],
        found,
      }))
    } else {
      Ok(())
    }
```

That doesn't work becaue `found:` needs to be a `&str`. But I couldn't figure out how to turn it into that. The closest I got generated this error:

```rust
error[E0515]: cannot return value referencing temporary value
   --> src/parser.rs:165:7
    |
164 |         let found_str = &format!("{}", next.kind).to_string();
    |                          ------------------------------------ temporary value created here
165 | /       Err(next.error(CompileErrorKind::ExpectedKeyword {
166 | |         expected: vec![expected],
167 | |         found: found_str,
168 | |       }))
    | |_________^ returns a value referencing data owned by the current function
```

Do you already have code which renders a `TokenKind` that I can look at? Sorry for the newbie question!

casey:
You can change the type of `found` from a `&str` to a `TokenKind`. Check out `CompileErrorKind` [here](https://github.com/casey/just/blob/a46be416991b83a16028becf724d4b382fd9559d/src/compile_error_kind.rs#L43).

nk9:
Oh, cool. So I did that, and it does indeed work for the `TokenKind`-is-not-`Identifier` case!

```rust
error: Expected keyword `else` but found identifier `end of line`
  |
1 | TEST := if path_exists('/bin/bash') == 'true' {'yes'}
  |                                                      ^
  ```

But that change also means that the error message actually becomes less clear when the problem is an unexpected identifier:

```rust
error: Expected keyword `else` but found identifier `identifier`
  |
1 | foo := if '' == '' { '' } arlo { '' }
  |                           ^^^^
```

If I could keep `found` as a string and actually pass the formatted name of the Kind, I think it might be the best of both worlds.

casey:
Gotcha. I guess maybe we should use an `UnexpectedKeyword` error, with `found` being a string, in the case that it's an identifier with the wrong lexeme (string value) but use `UnexpectedToken` if it's not an identifier at all.

nk9:
OK, I will look at adding a new error then.

nk9:
Oh, or did you mean `UnexpectedToken` and `ExpectedKeyword`? Then I wouldn't have to create a new error.

casey:
Whoops! Yup, that's right.

casey:
Agh, didn't mean to close.

nk9:
Hmm, so `UnexpectedToken` takes a `TokenKind` as `expected`. Unsurprisngly, when I pass `TokenKind::Identifier`, I get this error:

```rust
error: Expected identifier, but found end of line
  |
1 | TEST := if path_exists('/bin/bash') == 'true' {'yes'}
  |                                                      ^
```

So kind of back to square one. But if I just pass `expected` directly, it's a `Keyword` and so won't accept it:

```rust
error[E0308]: mismatched types
   --> src/parser.rs:165:24
    |
165 |         expected: vec![expected],
    |                        ^^^^^^^^ expected enum `token_kind::TokenKind`, found enum `keyword::Keyword`
```

Given this, would it be better to use an `ExpectedKeyword` error after all, and just convert the `lexeme` to a string as I was trying to do before?

nk9:
Latest code behaves as desired, but I'm going to guess that using any function called `leak` is probably not a long-term solution!

casey:
Okay, here's an idea, what if `ExpectedKeyword` looked like this:

```
ExpectedKeyword {
    expected: Vec<Keyword>,
    found: Token<'src>,
}
```

Then, the display implementation, in `compile_error.rs`, can do the following:

```
ExpectedKeyword { expected, found } => if found.kind == Identifier {
  write!(
    f,
    "Expected keyword {} but found identifier `{}`",
    List::or_ticked(expected),
    found.lexeme
  )?,
} else {
  write!(
    f,
    "Expected keyword {} but found `{}`",
    List::or_ticked(expected),
    found.kind
  )?,
}
```

nk9:
Yes, that works well. Much cleaner, thanks! I think this is ready to go.

casey:
Nice! And the test coverage looks great. The tests are failing on Windows, but they might just be flaky. I'll rerun, and if they're still failing we can figure it out.

casey:
Merged! Thanks for tackling this!

PR #1259, opened 2022-07-06
Add mdbook deps for build-book recipe (#1258)

### LINKED ISSUE #1258: build-book needs to install dependency of mdbook
if you don't have the `mdbook` binary installed, the `build-book` target fails. I would make a PR to just run `cargo install mdbook` (which fixes the problem), however, there's half a dozen places this could go. So this is really a design problem more than a technical one.
- You could put it in the `build-book` target. 
  - but you might want to avoid running the install every time. 
    - I would normally address this with something like `[[ -x $(which mdbook 2> /dev/null) ]] || cargo install mdbook` but that builds a dependency on bash and a *nix environment, which is probably undesirable. 
- You could put it in the `install-dev-dependencies` target
  - But that might irritate developers that just want dependencies for the binary and extend CI builds and so on.
- You could build a new target like `install-all-dev-dependencies` or `install-book-dev-dependencies` and put it in there.
- There are probably other places you could put it that I can't conceive of because I'm not familiar with the project. 

Thought I'd throw this out there. It's a minor problem, but one I can't just push a quick fix to unless I know the what the maintainers would prefer.

TopherIsSwell:
added mdbook dependencies needed to run recipe `build-book`. Tested in the alpine-based `docker.io/library/rust` container that running the `install-dev-deps` recipe and then the `build-book` recipe results in no errors. 

github issue #1258

### DIFF
--- a/justfile
+++ b/justfile
@@ -97,8 +97,9 @@ install-dev-deps:
   rustup install nightly
   rustup update nightly
   cargo +nightly install cargo-fuzz
-  cargo install cargo-check
+  cargo install cargo-check 
   cargo install cargo-watch
+  cargo install mdbook mdbook-linkcheck
 
 # install system development dependencies with homebrew
 install-dev-deps-homebrew:
--- a/justfile
+++ b/justfile
@@ -97,7 +97,7 @@ install-dev-deps:
   rustup install nightly
   rustup update nightly
   cargo +nightly install cargo-fuzz
-  cargo install cargo-check 
+  cargo install cargo-check
   cargo install cargo-watch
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -97,8 +97,9 @@ install-dev-deps:
   rustup install nightly
   rustup update nightly
   cargo +nightly install cargo-fuzz
-  cargo install cargo-check
+  cargo install cargo-check
```
casey:
```suggestion
  cargo install cargo-check
```

casey:
Nice, looks good!

PR #1264, opened 2022-07-11
add asdf to `Packages` under README

julian-avar:


casey:
Nice, thank you!

I don't know how asdf works, but is this command right:


```
asdf install asdf <version>
```

Should it be:

```
asdf install just <version>
```

Also, is it possible to do `asdf install just <version>` without doing `asdf plugin add just`? (Wondering if the command can be made shorter.)

julian-avar:
In my experience it's best to think of asdf as a package manager manager.

`asdf plugin add just` is actually the shortened version of `asdf plugin add just https://github.com/ggilmore/asdf-just.git`. The reason there is a shortcut is because someone added it to the [asdf community plugins repo](https://github.com/asdf-vm/asdf-plugins).

`asdf plugin add just` to add the alias. `asdf list all just` to view possible versions. `asdf install just <version>` to get the specific version you want, and `asdf global just <version>` to make change the global `just` version.

So from that perhaps you could make the command there longer to include `&& asdf global just <version>`, but most asdf users are used to that already. And so I believe that the command in my revision should be the shortest "friendly" command to get `just` installed.

casey:
Gotcha, thanks for the explanation!

Should it be `asdf install just <version>` then? (Currently it's `asdf install asdf <version>`.)

julian-avar:
Whoops i see what you mean now i made a small typo. You are 100% correct.

julian-avar:
Sorry that took a while, just got home :D. I hope it's useful to other asdf users.

casey:
Sweet, merging! Thank you!

Issue #1266, opened 2022-07-13
Allow multi-line recipe signatures


casey:
This came up in #1265. We can do this without requiring a continuation of any kind, just ignore newlines when parsing recipe parameters.

Great first issue!

neunenak:
Hey @casey , I thought I'd take a crack at this. https://github.com/casey/just/pull/1272 seems to work with at least one test justfile I've made - I'm still figuring out how the just codebase works, how tests work, etc., and I'm not sure if this is the best way to solve this parsing problem, so I don't want to call this ready for review just yet. I'd appreciate it if you had any suggestions about what to tell the parser to do here.

casey:
Nice!

I think the difficulty with the approach in #1272 is that we want to ignore newlines in grammatical sub constructs. So here:

```
    loop {
        if self.next_is(Identifier) || self.next_is(Dollar) {
            positional.push(self.parse_parameter(ParameterKind::Singular)?);
        } else if self.next_is(Eol) {
            self.accept(Eol)?;
        } else {
            break;
        }
    }
```

We also want to ignore newlines in parameters, which would require a change to parse-parameter.

One potential solution is to switch the parser between eol-sensitive and non eol-sensitive mode. In non EOL sensitive modes, parser functions that checked for tokens/returned tokens would switch the TokenKind of line endings whitespace.

I'm not really sure which approach is better, it really depends on how gross the code is for each.

jraygauthier:
One of the main pain point for me.

laniakea64:
@jraygauthier Why is it a main pain point to use [\ or parentheses](https://just.systems/man/en/chapter_47.html#outside-recipe-bodies)?

jraygauthier:
@laniakea64: Lol, because I wasn't aware this has been fixed yet (as it appears in v1.15) and was using v1.13:

```bash
$ just build-debug-features
error: Unknown start of token:
   |
33 | build-debug-features: \
   |                       ^

$ just --version
just 1.13.0
```

I will upgrade a more recent > v1.15 version.

Now, to me this ticket should be marked as fixed by PR https://github.com/casey/just/pull/1551, right? It might be a good idea to update Q/A https://github.com/casey/just/discussions/1265 as well.

casey:
Yup, I think this can be closed.

PR #1272, opened 2022-07-22
Allow newlines when parsing recipes

neunenak:
Addresses https://github.com/casey/just/issues/1266


casey:
Took a look at this, and I think that if this works, we should do it, even if it isn't a perfect solution. I looked at modifying the parser to switch between line-break sensitive and insensitive modes, and it seemed complicated.

casey:
I updated this, and added a test, but I'm running into an issue where there's an indent before the parameter, and then no indent in the recipe body, which messes this up.

laniakea64:
Given that the issue this PR aimed to resolve [has been fixed and closed](https://github.com/casey/just/issues/1266#issuecomment-1885988013), is this PR still needed?

Requiring the \ continuations here is better for human-readability of justfiles.  And this change would be hard to support properly in `vim-just` (if even possible?).

What would be the rationale for this change in current `just`?

casey:
I think it's slightly nicer to not have to add `\`, but I agree that it's not necessary, now that there are `\` line continuations, so I'll go ahead and close this.

PR #1275, opened 2022-07-22
Update just-install reference on README.md

0xradical:
Reference was pointing to the wrong page, causing a 404 error


casey:
Nice, merging!

Issue #429, opened 2019-05-06
Package just for Debian and get it accepted into the main package repository: $200 bounty


slisznia:
Casey: Highjacking the original post for visibility. I'd like to see Just packaged for Debian, but it's something I'm unlikely to do  myself, so I'm offering a $200 bounty to the person who both packages it *and* gets it accepted into the main Debian package repository.

For anyone interested in tackling this, rust packages are packaged by the Debian Rust team. Some useful links:
- https://wiki.debian.org/Teams/RustPackaging
- https://salsa.debian.org/rust-team/debcargo-conf

casey:
I would love to see `just` available in Debian and Ubuntu! I have no idea what the process for getting software into those distributions is though. For example:

- I know that Ubuntu is a derivative of Debian. If `just` is packaged for Debian, will it automatically make its way into Ubuntu, or does it also need to be packaged for Ubuntu?
- Are there any requirements for packages that `just` doesn't meet? If there are, I'd be very happy to work to rectify those. (For example, do binaries need to have a man page?)

I don't think not being 1.0 is an issue. And if it was, I would just bump to 1.0 right now. I'm not planning any huge, breaking changes in the future.

If anyone out there is willing to help get just into Debian and Ubuntu (or any other distro, linux or not) and I can do something to make the process easier, let me know!

slisznia:
The path to Ubuntu is (usually) through Debian. Here is introductory info
to be familiar with:

https://wiki.ubuntu.com/UbuntuDevelopment/NewPackages#NEW_packages_through_Debian

casey:
Thanks for the link, that's super helpful. I probably won't have time in the near future to package it myself, but I'd be very happy to help anyone who does.

NickeZ:
I have some work in progress on packaging just for debian here: https://salsa.debian.org/nickez-guest/debcargo-conf/commits/package-just

I believe that if we get into debian we get into ubuntu automatically.

NickeZ:
It is possible to create a simple man page with help2man. We can do that until we have a proper one.

```
help2man --name just --no-info --no-discard-stderr --version-string 4.4.0 just > just.1
```

```
JUST(1)                          User Commands                         JUST(1)

NAME
       just - just

DESCRIPTION
       just  v0.4.4  Casey  Rodarmor  <casey@rodarmor.com> ???? Just a command
       runner - https://github.com/casey/just

   USAGE:
              just [FLAGS] [OPTIONS] [--] [ARGUMENTS]...

[...]

PR #1280, opened 2022-07-24
Add MPR/Prebuilt-MPR installation instructions to README.md

hwittenborn:
This adds installation instructions for both the MPR and Prebuilt-MPR to the README. Both of these function as installation methods for Debian/Ubuntu system though, and I was thinking users might not know what to choose if they see both at once, so I was thinking if there was only one listed it would probably be the Prebuilt-MPR simply because it uses an APT repository instead of requiring additional tools to be installed.

The Prebuilt-MPR method is awaiting the completion of https://drone.hunterwittenborn.com/makedeb/prebuilt-mpr/324.

Partial solution for https://github.com/casey/just/issues/429.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,7 +228,7 @@ You can also set the shell using command-line arguments. For example, to use Pow
     <td><a href="https://mpr.makedeb.org/packages/just">just</a></td>
     <td>
       <code>git clone 'https://mpr.makedeb.org/just'</code><br>
-      <code>cd just/</code>
+      <code>cd just</code><br>
       <code>makedeb -si</code>
     </td>
   </tr>
--- a/README.md
+++ b/README.md
@@ -222,6 +222,25 @@ You can also set the shell using command-line arguments. For example, to use Pow
       <code>asdf install just &lt;version&gt;</code>
     </td>
   </tr>
+  <tr>
+    <td><a href="https://debian.org">Debian derivatives</td>
+    <td><a href="https://mpr.makedeb.org">MPR</a></td>
+    <td><a href="https://mpr.makedeb.org/packages/just">just</a></td>
+    <td>
+      <code>git clone 'https://mpr.makedeb.org/just'</code><br>
+      <code>cd just/</code>
+      <code>makedeb -si</code>
+    </td>
+  </tr>
+  <tr>
+    <td><a href="https://debian.org">Debian derivatives</a></td>
+    <td><a href="https://docs.makedeb.org/prebuilt-mpr">Prebuilt-MPR</a></td>
+    <td><a href="https://mpr.makedeb.org/packages/just">just</a></td>
+    <td>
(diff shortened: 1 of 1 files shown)


casey:
Whoops, didn't mean to close!

casey:
How about `Debian derivatives` instead of `Debian/Unbuntu`? Or is it only a Debian and Ubuntu thing? If so, how about `Debian & Ubuntu`, I like that a little better than the slash.

For the prebuilt MBR, I feel like the install command should mention that the prebuild MBR setup needs to be completed, otherwise just seeing `sudo apt install just` may make people skip that step. Maybe include a link to the getting started page?

So something like:

```
`sudo apt install just` [After adding the MPR](https://docs.makedeb.org/prebuilt-mpr/getting-started/)
```

casey:
I'm merging #1291, which converts the package table to HTML. The package table has been getting crazier and crazier, and I think the time has come to just convert it to an HTML table, which is more verbose, but much more flexible. You'll get merge conflicts, but the table is much easier to format, and you won't have to use out-of-line links.

hwittenborn:
Hey, thanks for the heads up @casey! I'm at a band camp thingy until Friday, so I can't make any guarantees on when I can look at this before then, though I plan on getting this completed right after

casey:
No worries and no rush, enjoy camp!

hwittenborn:
Sorry, was doing a force push with no changes to get my branch back to upstream and it closed the PR, I'll have something going again shortly.

hwittenborn:
> How about `Debian derivatives` instead of `Debian/Unbuntu`?

I added that, but I feel like it doesn't fit well, and might make less experienced users who don't know that Ubuntu is based on Debian not know that just is available. The packages in the MPR/Prebuilt-MPR are also applicable to any distros based on Debian and Ubuntu, but I feel having both Debian and Ubuntu listed would be a good idea considering how popular Ubuntu itself is.

README.md
```diff
@@ -222,6 +222,25 @@ You can also set the shell using command-line arguments. For example, to use Pow
       <code>asdf install just &lt;version&gt;</code>
     </td>
   </tr>
+  <tr>
+    <td><a href="https://debian.org">Debian derivatives</td>
+    <td><a href="https://mpr.makedeb.org">MPR</a></td>
+    <td><a href="https://mpr.makedeb.org/packages/just">just</a></td>
+    <td>
+      <code>git clone 'https://mpr.makedeb.org/just'</code><br>
+      <code>cd just/</code>
```
```suggestion
      <code>cd just</code><br>
```

casey:
I think you're right. How about `Debian and Ubuntu derivatives`?

hwittenborn:
> I think you're right. How about `Debian and Ubuntu derivatives`?

Works for me

casey:
Nice, looks great, merged! Thanks for sticking with this PR 🙇‍♂️

Issue #618, opened 2020-03-20
Add instructions to readme for setting up completions for supported shells


casey:
- [ ] zsh vanilla
- [ ] zsh oh-my-zsh
- [ ] zplug
- [ ] fish
- [ ] bash
- [ ] elvish
- [ ] powershell

kenden:
For zsh, using .oh-my-zsh, this works, but this is just a hack. If someone has a cleaner way, please share!

(edited March 28, 2023)
```bash
mkdir ~/.oh-my-zsh/plugins/just
just --completions zsh > ~/.oh-my-zsh/plugins/just/_just
```


Edit ~/.zshrc to add just in the plugin list, e.g.
```zsh
plugins=(
        git
        docker
        kubectl
        cargo
        rust
        just
)
```

And also add this to ~/.zshrc get the plugin to work (not needed in test from March 28 2023)

```zsh
autoload -Uz compinit && compinit -C  # hack to load just autocomplete
```

Note: the hack breaks some oh-my-zsh plugins, like 'go'.

lucasholder:
+1 , can't figure out how to use it with zplug :(  https://github.com/zplug/zplug/issues/547

schodet:
For zsh, put the completion script somewhere in your fpath, and be sure to have your fpath defined before calling compinit.

For example:

```
mkdir -p .zsh/completion
just --completions=zsh > .zsh/completion/_just
```

And in your .zshrc:

```
fpath=(~/.zsh/completion $fpath)
…
autoload -U compinit
compinit
```

kenden:
The code in my [last comment](https://github.com/casey/just/issues/618#issuecomment-601824467) had an issue; I found it actually works directly with ohmyzsh.

I created a pull request to make it easier to add to ohmyzsh:
https://github.com/ohmyzsh/ohmyzsh/pull/9287
-> edit march28 2023. This seems broken

casey:
@kenden Nice! Please let me know when that PR lands, so it can be added to the readme.

vmiheer:
@kenden I tried below with your commit (ohmyzsh/ohmyzsh#9287), looking at [omz wiki](https://github.com/ohmyzsh/ohmyzsh#custom-plugins-and-themes):

```
mkdir $ZSH_CUSTOM/plugins/just
[...]

PR #1285, opened 2022-07-25
Add info on zsh completion to README

nk9:
[As requested](https://github.com/casey/just/issues/618#issuecomment-1186597327), here's a summary of the zsh completion/Homebrew situation for the README.


casey:
Awesome, this looks great! Thank you for the detailed explanation!

PR #1295, opened 2022-07-28
Don't hardcode homebrew prefix

casey:
Hey @nk9, does this look better? This avoids hard-coding the homebrew prefix of `/opt/homebrew`, and just calling `brew shellenv`. I think this might be necessary on systems where brew is in `/usr/local`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2143,7 +2143,7 @@ $ just --completions zsh > just.zsh
 
 ```zsh
 # Init Homebrew, which adds environment variables
-eval "$(/opt/homebrew/bin/brew shellenv)"
+eval "$(brew shellenv)"
 
 fpath=($HOMEBREW_PREFIX/share/zsh/site-functions $fpath)
 


nk9:
I haven't forgotten about this! Would like to make this change, but I wanted to do a bit of research on Homebrew first. Will reply soon.

nk9:
The problem with this suggestion is that `brew` isn't in the user's `PATH` yet. The `shellenv` script updates `PATH` to add it, but before that is run the user has to specify the full path to the binary.

However, I have learned that currently, Homebrew can be installed in [THREE different places](https://docs.brew.sh/Installation), depending on platform:

> /usr/local for macOS Intel, /opt/homebrew for Apple Silicon and /home/linuxbrew/.linuxbrew for Linux

Sooo… I guess that needs to be mentioned in the README text too, since this code will not work for the majority of users right now who are on Intel Macs.

nk9:
See #1304

PR #1296, opened 2022-07-28
update suggested default recipe to include a justfile option

jpbochi:
This is an update to documentation, with a better recipe suggestion.

---

I had a `justfile` with the following content:

```
# as suggested at [Listing Available Recipes](https://just.systems/man/en/chapter_22.html)
default:
  @just --list

pwd:
  pwd
```

Then, I tried to run `.j` (aliased to `just -f ~/.justfile -d .` as suggested at [User justfiles](https://just.systems/man/en/chapter_63.html)), and got the following error:

```
$ .j
error: No justfile found
error: Recipe `default` failed on line 4 with exit code 1
```

Running `.j pwd` worked. The issue if that `.j` would run the `default` recipe, which executes `just --list` without passing the `--justfile` option again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -576,11 +576,10 @@ If you'd like `just` to default to listing the recipes in the `justfile`, you ca
 
 ```make
 default:
-  @# list recipes from the current justfile
-  @just --justfile {{justfile()}} --list
+  @just --list
 ```
 
-Note that the `--justfile {{justfile()}}` above is needed to make sure the listed recipes are from the current `justfile`. Without it, if you executed `just -f /some/distant/justfile` or `just -f ./non-standard-justfile`, the plain `just --list` call in the `default` recipe would not use the file you provided. It would try to find a justfile in your current path, maybe even resulting in a `No justfile found` error.
+Note that you may need to add `--justfile {{justfile()}}` to the line above above. Without it, if you executed `just -f /some/distant/justfile` or `just -f ./non-standard-justfile`, the plain `just --list` inside the recipe would not necessarily use the file you provided. It would try to find a justfile in your current path, maybe even resulting in a `No justfile found` error.
 
 The heading text can be customized with `--list-heading`:
 
--- a/README.md
+++ b/README.md
@@ -576,6 +576,7 @@ If you'd like `just` to default to listing the recipes in the `justfile`, you ca
 
 ```make
 default:
+  @# list recipes from the current justfile
   @just --justfile {{justfile()}} --list
 ```
 
--- a/README.md
+++ b/README.md
@@ -580,7 +580,7 @@ default:
   @just --justfile {{justfile()}} --list
 ```
 
-Note that the `-f {{justfile()}}` above is needed to make sure the listed are from the current `justfile`. If you executed `just -f /path/to/justfile`, you could list recipes from somewhere else or get a `No justfile found` error.
+Note that the `--justfile {{justfile()}}` above is needed to make sure the listed recipes are from the current `justfile`. Without it, if you executed `just -f /some/distant/justfile` or `just -f ./non-standard-justfile`, the plain `just --list` call in the `default` recipe would not use the file you provided. It would try to find a justfile in your current path, maybe even resulting in a `No justfile found` error.
 
 The heading text can be customized with `--list-heading`:
 
--- a/README.md
+++ b/README.md
@@ -576,8 +576,10 @@ If you'd like `just` to default to listing the recipes in the `justfile`, you ca
 
 ```make
 default:
-  @just --list
+  @just -f {{justfile()}} --list
 ```
+
+Note that the `-f {{justfile()}}` above is needed to make sure the listed are from the current `justfile`. If you executed `just -f /path/to/justfile`, you could list recipes from somewhere else or get a `No justfile found` error.
 
 The heading text can be customized with `--list-heading`:
 
--- a/README.md
+++ b/README.md
@@ -579,7 +579,7 @@ default:
   @just --list
 ```
 
-Note that you may need to add `--justfile {{justfile()}}` to the line above above. Without it, if you executed `just -f /some/distant/justfile` or `just -f ./non-standard-justfile`, the plain `just --list` inside the recipe would not necessarily use the file you provided. It would try to find a justfile in your current path, maybe even resulting in a `No justfile found` error.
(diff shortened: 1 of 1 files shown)


casey:
Good idea!

README.md
```diff
@@ -557,9 +557,11 @@ If you'd like `just` to default to listing the recipes in the `justfile`, you ca
 
 ```make
 default:
-  @just --list
+  @just -f {{justfile()}} --list
```
casey: I like using use long options in documentation and examples, since it's clearer than short:

```suggestion
  @just --justfile {{justfile()}} --list
```

README.md
```diff
@@ -557,9 +557,11 @@ If you'd like `just` to default to listing the recipes in the `justfile`, you ca
...
 ```
 
+Note that the `-f {{justfile()}}` above is needed to make sure the listed are from the current `justfile`. If you executed `just -f /path/to/justfile`, you could list recipes from somewhere else or get a `No justfile found` error.
```
casey: I think this could use some work. It's not clear *why* you wouldn't get the right justfile, so I think it should mention something about why you could get the wrong justfile.

README.md
```suggestion
  @just --justfile {{justfile()}} --list
```

README.md
```diff
@@ -557,9 +557,11 @@ If you'd like `just` to default to listing the recipes in the `justfile`, you ca
 
 ```make
 default:
-  @just --list
+  @just --justfile {{justfile()}} --list
```
```suggestion
  @# list recipes from the current justfile
  @just --justfile {{justfile()}} --list
```

README.md
casey: I think the thing it should mention is that if you're using a justfile with a non-standard name, you might not find it.

README.md
jpbochi: I prefer long options, too, but made the wrong assumptions about your preference. :)

README.md
jpbochi: it's either a non-standard name or a distant path, which was my case. Let me see if I can rephrase it better.

README.md
```diff
@@ -557,9 +557,12 @@ If you'd like `just` to default to listing the recipes in the `justfile`, you ca
 
 ```make
 default:
-  @just --list
+  @# list recipes from the current justfile
+  @just --justfile {{justfile()}} --list
 ```
 
+Note that the `--justfile {{justfile()}}` above is needed to make sure the listed recipes are from the current `justfile`. Without it, if you executed `just -f /some/distant/justfile` or `just -f ./non-standard-justfile`, the plain `just --list` call in the `default` recipe would not use the file you provided. It would try to find a justfile in your current path, maybe even resulting in a `No justfile found` error.
```
jpbochi: @casey Is this better?

casey: Is `just -f /some/distant/justfile` a problem? I think just will switch its working directory and find the justfile.

jpbochi: in the case of that `.j` alias, the directory is already set. I don't think there's other solution to combine both (the alias and the list recipe, I mean).

jpbochi: A more generic feature that `just` may need is an alternative to this [`$(MAKE) $(MFLAGS)`](https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html) for calling a "sub" `just`.

Maybe, a `just()` function could be equivalent to `{{just_executable()}} --justfile {{justfile()}}`. Then, recommend that any recursive just calls are made with `{{just()}}` instead of plainly `just`.

casey: We definitely need something like. I want some way of running a just command in the middle of a recipe, but we haven't figured out the syntax yet.

casey: I'm having trouble reproducing the behavior you mentioned. I created a justfile in foo/.justfile, which has `just --list` in the first recipe, calling `just -f foo/.justfile` seems to work, it successfully calls just and lists the recipes in the justfile.

jpbochi: The output below may help.

Notice that:
1. the `jj` alias includes a `-d .` option, as is suggested at https://just.systems/man/en/chapter_63.html#forwarding-alias;
2. all executions of `jj` did echo `just --list`, meaning it's running the recipe from `~/justfile` 
3. when `pwd` is inside my home (e.g. `~/src/`), `jj` works as intended.
4. when `pwd` is outside of `~`, I get a `No justfile found` _coming from the default recipe_;
5. when `pwd` is in a directory with another justfile, it lists recipes from that directory instead.

The only way I see to prevent the unintended behavior (4 and 5) is by changing the recipe to say `just --justfile {{justfile()}} --list`.

```
$ cat ~/justfile
default:
  just --list

pwd:
  pwd

$ type jj
jj is aliased to `just -f ~/justfile -d .'

$ cd ~/src
$ jj
just --list
Available recipes:
    default
    pwd

$ cd /
$ jj
just --list
error: No justfile found
error: Recipe `default` failed on line 2 with exit code 1

$ cd ~/src/jp/just/
$ jj
just --list
Available recipes:
    build
    build-book
    changes                   # add git log messages to changelog
    check
    c                         # alias for `check`
    …
```

casey:
Ah, okay, sorry, I misunderstood. I thought that it was just the `-f` flag that would make things not work, but it's the `-d` flag. I'm actually leaning towards not changing the recipe, and just adding the note, mentioning that if you have a non-default justfile name or different directory, you'll need to to add `--justfile {{justfile()}}`.

jpbochi:
@casey I'm cool with that, too. The current example would work as is for the majority of cases, and there's value in keeping them simple.

I updated the code with that now.

README.md
```diff
@@ -579,6 +579,8 @@ default:
   @just --list
 ```
 
+Note that you may need to add `--justfile {{justfile()}}` to the line above above. Without it, if you executed `just -f /some/distant/justfile` or `just -f ./non-standard-justfile`, the plain `just --list` inside the recipe would not necessarily use the file you provided. It would try to find a justfile in your current path, maybe even resulting in a `No justfile found` error.
```
```suggestion
Note that you may need to add `--justfile {{justfile()}}` to the line above above. Without it, if you executed `just -f /some/distant/justfile -d .` or `just -f ./non-standard-justfile`, the plain `just --list` inside the recipe would not necessarily use the file you provided. It would try to find a justfile in your current path, maybe even resulting in a `No justfile found` error.
```

casey:
Nice! I added the `-d` flag to the first example, since that's needed to get it to break.

PR #1303, opened 2022-08-04
allow bash completion to complete tasks in other directories

jpbochi:
Today, I learned about just's ability to [invoke justfiles in other directories](https://just.systems/man/en/chapter_50.html#invoking-justfiles-in-other-directories).

I moved part of my tasks into `./ci/justfile`. Then, with bash completion on, typed `just ci/` + _tab_. It completed with files, not with `just` tasks.

This change here solves that problem.

If the current text being completed has a `/` char, we'll run a different `just --summary` command.

### DIFF
--- a/completions/just.bash
+++ b/completions/just.bash
@@ -25,7 +25,15 @@ _just() {
                     COMPREPLY=( $(compgen -W "${opts}" -- "${cur}") )
                     return 0
                 elif [[ ${COMP_CWORD} -eq 1 ]]; then
-                    local recipes=$(just --summary --color never 2> /dev/null)
+                    if echo "${cur}" | grep -qF '/'; then
+                        local path_prefix=$(echo "${cur}" | sed 's/[/][^/]*$/\//')
+                        # TODO: Replace tr and awk below with --list-prefix "${path_prefix}". Need to update the --summary option to respect that first.
+                        local recipes=$(just --summary --color never 2> /dev/null -- "${path_prefix}" \
+                                | tr '[:blank:]' '\n' | awk "{print \"${path_prefix}\"\$0}")
+                    else
+                        local recipes=$(just --summary --color never 2> /dev/null)
+                    fi
+
                     if [[ $? -eq 0 ]]; then
                         COMPREPLY=( $(compgen -W "${recipes}" -- "${cur}") )
                         return 0
--- a/completions/just.bash
+++ b/completions/just.bash
@@ -25,13 +25,12 @@ _just() {
                     COMPREPLY=( $(compgen -W "${opts}" -- "${cur}") )
                     return 0
                 elif [[ ${COMP_CWORD} -eq 1 ]]; then
+                    local recipes=$(just --summary --color never 2> /dev/null)
+
                     if echo "${cur}" | grep -qF '/'; then
                         local path_prefix=$(echo "${cur}" | sed 's/[/][^/]*$/\//')
-                        # TODO: Replace tr and awk below with --list-prefix "${path_prefix}". Need to update the --summary option to respect that first.
-                        local recipes=$(just --summary --color never 2> /dev/null -- "${path_prefix}" \
-                                | tr '[:blank:]' '\n' | awk "{print \"${path_prefix}\"\$0}")
-                    else
-                        local recipes=$(just --summary --color never 2> /dev/null)
+                        local recipes=$(just --summary --color never 2> /dev/null -- "${path_prefix}")
+                        local recipes=$(printf "${path_prefix}%s" $recipes)
                     fi
 
                     if [[ $? -eq 0 ]]; then
--- a/completions/just.bash
+++ b/completions/just.bash
@@ -30,7 +30,7 @@ _just() {
                     if echo "${cur}" | grep -qF '/'; then
                         local path_prefix=$(echo "${cur}" | sed 's/[/][^/]*$/\//')
                         local recipes=$(just --summary 2> /dev/null -- "${path_prefix}")
-                        local recipes=$(printf "${path_prefix}%s" $recipes)
+                        local recipes=$(printf "${path_prefix}%s\t" $recipes)
                     fi
 
                     if [[ $? -eq 0 ]]; then
--- a/completions/just.bash
+++ b/completions/just.bash
@@ -25,11 +25,11 @@ _just() {
                     COMPREPLY=( $(compgen -W "${opts}" -- "${cur}") )
                     return 0
                 elif [[ ${COMP_CWORD} -eq 1 ]]; then
-                    local recipes=$(just --summary --color never 2> /dev/null)
+                    local recipes=$(just --summary 2> /dev/null)
 
                     if echo "${cur}" | grep -qF '/'; then
                         local path_prefix=$(echo "${cur}" | sed 's/[/][^/]*$/\//')
-                        local recipes=$(just --summary --color never 2> /dev/null -- "${path_prefix}")
+                        local recipes=$(just --summary 2> /dev/null -- "${path_prefix}")
                         local recipes=$(printf "${path_prefix}%s" $recipes)
                     fi
 
--- a/tests/completions/just.bash
+++ b/tests/completions/just.bash
@@ -1,13 +1,26 @@
 #!/usr/bin/env bash
 
-source ./completions/just.bash
+# --- Shared functions ---
 
-exit_code='0'
+cdroot() {
+  cd "$(git rev-parse --show-toplevel)" > /dev/null
+}
+
+setup() {
+  cdroot
+  mkdir -p ./tmp > /dev/null
+  echo 'install:' > tmp/justfile
+  echo 'test:' >> tmp/justfile
+  echo 'deploy:' >> tmp/justfile
+  echo 'push:' >> tmp/justfile
+  echo 'publish:' >> tmp/justfile
+}
 
 cleanup() {
   unset COMP_WORDS
   unset COMP_CWORD
   unset COMPREPLY
+  rm -f ./tmp/justfile
 }
 
 reply_equals() {
@@ -26,6 +39,14 @@ reply_equals() {
   fi
 }
 
+# --- Initial Setup ---
+cdroot
+source ./completions/just.bash
+PATH="$(git rev-parse --show-toplevel)/target/debug:$PATH"
+exit_code='0'
+
+# --- Tests ---
+
 test_just_is_accessible() {
   if just --version > /dev/null; then
     echo "${FUNCNAME[0]}: ok"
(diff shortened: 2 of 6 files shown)


completions/just.bash
```diff
@@ -25,7 +25,15 @@ _just() {
                     COMPREPLY=( $(compgen -W "${opts}" -- "${cur}") )
                     return 0
                 elif [[ ${COMP_CWORD} -eq 1 ]]; then
-                    local recipes=$(just --summary --color never 2> /dev/null)
+                    if echo "${cur}" | grep -qF '/'; then
+                        local path_prefix=$(echo "${cur}" | sed 's/[/][^/]*$/\//')
+                        # TODO: Replace tr and awk below with --list-prefix "${path_prefix}". Need to update the --summary option to respect that first.
```
jpbochi: naturally, I'd like to implement this. About that:

1. I don't know enough rust to implement it [well].
2. I don't know if there's a good reason for `--summary` to really ignore `--list-prefix`.

casey: Sorry for the delay in checking this out! `--summary` is intended for consumption by scripts. How would `--list-prefix` be used?

jpbochi:
COMPREPLY needs to have a full string to be completed. For example, when typing "ci/dep" then tab, COMPREPLY needs to say "ci/deploy" for completion to work. Right now, `just --summary ci/` would only say "deploy".

Using --list-prefix is one way to add "ci/" there. In retrospect, it's probably not the best way. Perhaps, --summary should include that automatically? This would be a breaking change for some scripts out there, unfortunately.

-------- Original Message --------

completions/just.bash
jpbochi: COMPREPLY needs to have a full string to be completed. For example, when typing "ci/dep" then tab, COMPREPLY needs to say "ci/deploy" for completion to work. Right now, `just --summary ci/` would only say "deploy".

Using --list-prefix is one way to add "ci/" there. In retrospect, it's probably not the best way. Perhaps, --summary should include that automatically? This would be a breaking change for some scripts out there, unfortunately.

casey: See this answer for an alternative approach: https://stackoverflow.com/a/48663800

jpbochi: ah, neat! That printf solution is less awk-ward than awk. :D

There's still the question of whether `just --summary` itself should include that prefix or not. What are your thoughts on that?

For now, I'll remove the TODO and change this to use printf.

casey: > There's still the question of whether just --summary itself should include that prefix or not. What are your thoughts on that?

Hmm, good question. So basically, the prefix is the directory prefix to get to the justfile, and thus invoke the recipe in question? I think it probably shouldn't, and it should be up to the caller who invoked just to add it.

completions/just.bash
```diff
@@ -26,6 +26,13 @@ _just() {
                     return 0
                 elif [[ ${COMP_CWORD} -eq 1 ]]; then
                     local recipes=$(just --summary --color never 2> /dev/null)
+
+                    if echo "${cur}" | grep -qF '/'; then
+                        local path_prefix=$(echo "${cur}" | sed 's/[/][^/]*$/\//')
+                        local recipes=$(just --summary --color never 2> /dev/null -- "${path_prefix}")
```
casey: I think summary never colors its output:

```suggestion
                        local recipes=$(just --summary 2> /dev/null -- "${path_prefix}")
```

jpbochi: FWIW, that was already there before. If it's not needed I'll do the change, of course.

completions/just.bash
jpbochi: it would still be interesting to have an option to include the pull path, I suppose.

jpbochi:
@casey Can you take a look at this one again? I believe I addressed all your previous points.

casey:
Sorry for the delay! Just checked it out, and I think there's a bug. One the left is the result of me typing `just <tab>` in the PR branch, and on the right is me typing `just <tab>` on master. Both of these are after having done `source completions/just.bash` to pick up the completion script. It looks like something is going wrong with the terminal width detection in the PR branch's completion script.

Screen Shot 2022-08-17 at 10 40 18 PM

jpbochi:
@casey That's very odd. A bash completion only writes to the `COMPREPLY` variable. It's up to bash itself to deal with terminal width.

Here are some tests I did from this branch:

```bash
$ source ./completions/just.bash

$ COMP_WORDS=(just) && COMP_CWORD=1 _just just && declare -p COMPREPLY
declare -a COMPREPLY=([0]="build" [1]="build-book" [2]="changes" [3]="check" [4]="ci" [5]="clippy" [6]="done" [7]="filter" [8]="fmt" [9]="forbid" [10]="fuzz" [11]="generate-completions" [12]="install" [13]="install-dev-deps" [14]="install-dev-deps-homebrew" [15]="man" [16]="polyglot" [17]="pr" [18]="publish" [19]="push" [20]="pwd" [21]="quine" [22]="render-readme" [23]="replace" [24]="run" [25]="sloc" [26]="test" [27]="test-quine" [28]="view-man" [29]="watch" [30]="watch-readme")

$ COMP_WORDS=(just i) && COMP_CWORD=1 _just just && declare -p COMPREPLY
declare -a COMPREPLY=([0]="install" [1]="install-dev-deps" [2]="install-dev-deps-homebrew")

$ COMP_WORDS=(just r) && COMP_CWORD=1 _just just && declare -p COMPREPLY
declare -a COMPREPLY=([0]="render-readme" [1]="replace" [2]="run")
```

And now with just 1.4.0:

```bash
$ eval "$(just --completions bash)"

$ COMP_WORDS=(just) && COMP_CWORD=1 _just just && declare -p COMPREPLY
declare -a COMPREPLY=([0]="build" [1]="build-book" [2]="changes" [3]="check" [4]="ci" [5]="clippy" [6]="done" [7]="filter" [8]="fmt" [9]="forbid" [10]="fuzz" [11]="generate-completions" [12]="install" [13]="install-dev-deps" [14]="install-dev-deps-homebrew" [15]="man" [16]="polyglot" [17]="pr" [18]="publish" [19]="push" [20]="pwd" [21]="quine" [22]="render-readme" [23]="replace" [24]="run" [25]="sloc" [26]="test" [27]="test-quine" [28]="view-man" [29]="watch" [30]="watch-readme")

$ COMP_WORDS=(just i) && COMP_CWORD=1 _just just && declare -p COMPREPLY
declare -a COMPREPLY=([0]="install" [1]="install-dev-deps" [2]="install-dev-deps-homebrew")

$ COMP_WORDS=(just r) && COMP_CWORD=1 _just just && declare -p COMPREPLY
declare -a COMPREPLY=([0]="render-readme" [1]="replace" [2]="run")
```

The version of bash I'm using is `GNU bash, version 5.1.16(1)-release (aarch64-apple-darwin21.1.0)`, for the record.

jpbochi:
I found a different bug, and fixed it.

```
$ mkdir ci

$ echo 'install:' >> ci/justfile

$ echo 'deploy:' >> ci/justfile

$ source ./completions/just.bash && COMP_WORDS=(just ci/) && COMP_CWORD=1 _just just && declare -p COMPREPLY
declare -a COMPREPLY=([0]="ci/deploy" [1]="ci/install")

$ echo 'another:' >> ci/justfile

$ source ./completions/just.bash && COMP_WORDS=(just ci/) && COMP_CWORD=1 _just just && declare -p COMPREPLY
declare -a COMPREPLY=([0]="ci/another" [1]="ci/deploy" [2]="ci/install")
```

It would be great to add those as unit tests somehow.

casey:
If you could add unit tests for this, that would be amazing. So far I haven't because I'm not sure how to do that 😅

indigoviolet:
Sorry if this is a tangent, but I noticed that the completion script won't complete filenames in some contexts: for example, after `just -f<TAB>` or `just --dotenv-filename<TAB>`. Is it possible to ensure that always happens?

jpbochi:
I've used BATS (https://bats-core.readthedocs.io/) in the past. It's focused on bash, instead of zsh or plain sh, but it may be flexible enough to allow tests for other shells, too. It has been in active development for many years.

A recent alternative is https://shellspec.info, but I'm not too familiar with it. Its "BDD" API seems pretty weird to me.

One alternative is to write our own test script for now.

jpbochi:
@casey I added a custom script to test completion of bash. That could be extended later for other shells. And could be converted to BATS, if needed.

jpbochi:
for future reference, I found out abotu that way to test bash completion here: https://stackoverflow.com/questions/3520936/accessing-bash-completions-for-specific-commands-programmatically

casey:
Sorry for the delay once again! Taking a look again, and the tests seem to still be failing locally for me:

```
: ./tests/completions/just.bash                                                                                                                                                                                      ~/src/just ⟦jp/completion⟧
test_just_is_accessible: ok
test_complete_all_recipes: failed! Completion for `just` does not match.

--- expected
+++ actual
@@ -1 +1 @@
-declare -a COMPREPLY=([0]="build" [1]="build-book" [2]="changes" [3]="check" [4]="ci" [5]="clippy" [6]="done" [7]="filter" [8]="fmt" [9]="forbid" [10]="fuzz" [11]="generate-completions" [12]="install" [13]="install-dev-deps" [14]="install-dev-deps-homebrew" [15]="man" [16]="polyglot" [17]="pr" [18]="publish" [19]="push" [20]="pwd" [21]="quine" [22]="render-readme" [23]="replace" [24]="run" [25]="sloc" [26]="test" [27]="test-quine" [28]="view-man" [29]="watch" [30]="watch-readme")
+declare -a COMPREPLY='([0]="build" [1]="build-book" [2]="changes" [3]="check" [4]="ci" [5]="clippy" [6]="done" [7]="filter" [8]="fmt" [9]="forbid" [10]="fuzz" [11]="generate-completions" [12]="install" [13]="install-dev-deps" [14]="install-dev-deps-homebrew" [15]="man" [16]="polyglot" [17]="pr" [18]="publish" [19]="push" [20]="pwd" [21]="quine" [22]="render-readme" [23]="replace" [24]="run" [25]="sloc" [26]="test" [27]="test-quine" [28]="view-man" [29]="watch" [30]="watch-readme")'

test_complete_recipes_starting_with_i: failed! Completion for `just i` does not match.

--- expected
+++ actual
@@ -1 +1 @@
-declare -a COMPREPLY=([0]="install" [1]="install-dev-deps" [2]="install-dev-deps-homebrew")
+declare -a COMPREPLY='([0]="install" [1]="install-dev-deps" [2]="install-dev-deps-homebrew")'

test_complete_recipes_starting_with_r: failed! Completion for `just r` does not match.

--- expected
+++ actual
@@ -1 +1 @@
-declare -a COMPREPLY=([0]="render-readme" [1]="replace" [2]="run")
+declare -a COMPREPLY='([0]="render-readme" [1]="replace" [2]="run")'

test_complete_recipes_from_subdirs: failed! Completion for `just tmp/` does not match.

--- expected
+++ actual
@@ -1 +1 @@
-declare -a COMPREPLY=([0]="tmp/deploy" [1]="tmp/install")
+declare -a COMPREPLY='([0]="tmp/deploy" [1]="tmp/install")'

Some test[s] failed.
```

Also, the tests should run against a test justfile, since otherwise the tests will need to be updated every time the main justfile changes.

jpbochi:
@casey interesting... My guess is that this is caused by a different bash version. Which `bash --version` do you have?

I'm on `GNU bash, version 5.1.16(1)-release (aarch64-apple-darwin21.1.0)`, while CI has `GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)`. And those tests pass on both of them.

Naturally, I'll try to adapt them to pass regardless.

casey:
Ah, good point. I have the super old `bash` that ships with MacOS. I installed a newer one and it worked, so don't worry about making it pass with the old version.

jpbochi:
@casey This should be good to be merged then. The change works (for me , at least), and tests are passing.

I think that code should even work on the older bash you had before. The issue was only on the test diff.

jpbochi:
@casey can you take a look at this again? I think it's ready to be merged, unless you have some pending concerns I missed.

casey:
Sorry for letting this sit for so long! Looks good, although can you make it use a dummy justfile that's only used for the test, instead of the project's justfile? If it uses the main justfile, then it will break when it changes.

jpbochi:
> ... can you make it use a dummy justfile that's only used for the test, instead of the project's justfile? If it uses the main justfile, then it will break when it changes.

sounds reasonable. I'll make the change.

jpbochi:
@casey updated. Tests don't rely on the main justfile anymore.

casey:
Nice! Overall looks good. Instead of creating a tempdir and temp justfile, how about committing the test justfile to `tests/completions` and cd-ing into that directory to perform the tests?

jpbochi:
@casey done. I created and commited two test justfiles, and used those. :)

tests/completions/just.bash
```diff
@@ -0,0 +1,102 @@
...
+
+cdroot() {
+  cd "$(git rev-parse --show-toplevel)" > /dev/null
+}
+
+setup() {
+  cdroot
+  cd tests/completions > /dev/null
+}
+
```
casey, line 13:
```suggestion
cd tests/completions
```

casey:
Nice, merged! Thanks for sticking with this!

PR #1304, opened 2022-08-05
Add different platform paths to .zshrc setup instructions

nk9:
Annoyed this has to be spelled out, but this is documentation after all!


casey:
I'm actually not sure we should be documenting this. After all, for a user to successfully use homebrew at all, they have to put it in their path, and telling them how to do that is probably outside the scope of the just readme.

nk9:
OK, that's fair. You're right, this should already be done. As long as they put the README code at the bottom of their .zshrc/.zprofile, it should work.

PR #1311, opened 2022-08-15
Add an example of using a variable in a recipe parameter

papertigers:


README.md
```diff
@@ -1441,6 +1441,18 @@ build target:
   cd {{target}} && make
 ```
 
+Passing a variable as an argument to a dependency, does not require interpolation:
```
```suggestion
Variables can also be passed as arguments to dependencies:
```

casey:
Nice, merged!

Issue #1298, opened 2022-07-28
Path operator can't be used to construct absolute paths


wezm:
It seemed fairly natural (to me) to try to construct an absolute path with the `/` operator, however this does not work.

E.g.

```just
remote_project_path :=  / "home" / user / project
```

yields

```
error: Expected backtick, identifier, '(', or string, but found '/'
   |
12 | remote_project_path := / "home" / user / project
   |                        ^
```

I tried wrapping it in parens in case that helped but it did not.

casey:
Seems natural to me too! If anyone wants to grab this, it would make a great first issue.

erikkrieg:
I took a stab at this change here: https://github.com/casey/just/pull/1320

wezm:
Thanks so much for tackling this issue @erikkrieg 🎉

casey:
Just released 1.5.0, which includes this change. Release binaries should be available shortly: https://github.com/casey/just/releases

PR #1320, opened 2022-08-26
makes slash operator default lhs value to an empty string

erikkrieg:
I was looking for a _good first issue_ and came across https://github.com/casey/just/issues/1298

~The change is that the parser will now default an empty string literal on the left hand side of the `/` operator if it fails unwrapping a value for the lhs. This makes it possible to create a path with an expression like: `/ "Users" / user`.~

~Lemme know if there is anything that should be changed!~

Edit:
The approach was changed to wrapping the left-hand side (lhs) value of a join expression (using the `/` operator) to an `Option`.

### DIFF
--- a/src/evaluator.rs
+++ b/src/evaluator.rs
@@ -176,13 +176,11 @@ impl<'src, 'run> Evaluator<'src, 'run> {
         }
       }
       Expression::Group { contents } => self.evaluate_expression(contents),
-      Expression::Join { lhs, rhs } => {
-        let lhs = match lhs {
-          Some(lhs) => self.evaluate_expression(lhs)?,
-          None => String::new(),
-        };
-        Ok(lhs + "/" + &self.evaluate_expression(rhs)?)
-      }
+      Expression::Join { lhs: None, rhs } => Ok("/".to_string() + &self.evaluate_expression(rhs)?),
+      Expression::Join {
+        lhs: Some(lhs),
+        rhs,
+      } => Ok(self.evaluate_expression(lhs)? + "/" + &self.evaluate_expression(rhs)?),
     }
   }
 
--- a/src/evaluator.rs
+++ b/src/evaluator.rs
@@ -177,10 +177,10 @@ impl<'src, 'run> Evaluator<'src, 'run> {
       }
       Expression::Group { contents } => self.evaluate_expression(contents),
       Expression::Join { lhs, rhs } => {
-        let lhs = if lhs.is_some() {
-          self.evaluate_expression(lhs.as_ref().unwrap())?
+        let lhs = if let Some(lhs) = lhs {
+          self.evaluate_expression(lhs)?
         } else {
-          "".to_string()
+          String::new()
         };
         Ok(lhs + "/" + &self.evaluate_expression(rhs)?)
       }
--- a/src/evaluator.rs
+++ b/src/evaluator.rs
@@ -177,10 +177,9 @@ impl<'src, 'run> Evaluator<'src, 'run> {
       }
       Expression::Group { contents } => self.evaluate_expression(contents),
       Expression::Join { lhs, rhs } => {
-        let lhs = if let Some(lhs) = lhs {
-          self.evaluate_expression(lhs)?
-        } else {
-          String::new()
+        let lhs = match lhs {
+          Some(lhs) => self.evaluate_expression(lhs)?,
+          None => String::new(),
         };
         Ok(lhs + "/" + &self.evaluate_expression(rhs)?)
       }
--- a/src/evaluator.rs
+++ b/src/evaluator.rs
@@ -177,7 +177,12 @@ impl<'src, 'run> Evaluator<'src, 'run> {
       }
       Expression::Group { contents } => self.evaluate_expression(contents),
       Expression::Join { lhs, rhs } => {
-        Ok(self.evaluate_expression(lhs)? + "/" + &self.evaluate_expression(rhs)?)
+        let lhs = if lhs.is_some() {
+          self.evaluate_expression(lhs.as_ref().unwrap())?
+        } else {
+          "".to_string()
+        };
+        Ok(lhs + "/" + &self.evaluate_expression(rhs)?)
       }
     }
   }
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -45,30 +45,17 @@ impl<'src> Expression<'src> {
   pub(crate) fn variables<'expression>(&'expression self) -> Variables<'expression, 'src> {
     Variables::new(self)
   }
-
-  pub(crate) fn empty_string_literal() -> Self {
-    Expression::StringLiteral {
-      string_literal: StringLiteral {
-        kind: StringKind::new(string_kind::StringDelimiter::QuoteDouble, false),
-        raw: "",
-        cooked: "".to_string(),
-      },
-    }
-  }
 }
 
 impl<'src> Display for Expression<'src> {
   fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
     match self {
       Expression::Backtick { token, .. } => write!(f, "{}", token.lexeme()),
-      Expression::Join { lhs, rhs } => write!(
-        f,
-        "{} / {}",
-        lhs
-          .as_ref()
-          .unwrap_or(&Box::new(Expression::empty_string_literal())),
-        rhs
-      ),
+      Expression::Join { lhs: None, rhs } => write!(f, "/ {}", rhs),
+      Expression::Join {
+        lhs: Some(lhs),
+        rhs,
+      } => write!(f, "{} / {}", lhs, rhs),
       Expression::Concatenation { lhs, rhs } => write!(f, "{} + {}", lhs, rhs),
       Expression::Conditional {
         lhs,
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -45,6 +45,16 @@ impl<'src> Expression<'src> {
   pub(crate) fn variables<'expression>(&'expression self) -> Variables<'expression, 'src> {
     Variables::new(self)
   }
+
+  pub(crate) fn empty_string_literal() -> Self {
+    Expression::StringLiteral {
+      string_literal: StringLiteral {
+        kind: StringKind::new(string_kind::StringDelimiter::QuoteDouble, false),
+        raw: "",
+        cooked: "".to_string(),
+      },
+    }
+  }
 }
 
 impl<'src> Display for Expression<'src> {
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -32,7 +32,7 @@ pub(crate) enum Expression<'src> {
   Group { contents: Box<Expression<'src>> },
   /// `lhs / rhs`
   Join {
-    lhs: Box<Expression<'src>>,
+    lhs: Option<Box<Expression<'src>>>,
     rhs: Box<Expression<'src>>,
   },
   /// `"string_literal"` or `'string_literal'`
@@ -61,7 +61,14 @@ impl<'src> Display for Expression<'src> {
   fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
     match self {
       Expression::Backtick { token, .. } => write!(f, "{}", token.lexeme()),
-      Expression::Join { lhs, rhs } => write!(f, "{} / {}", lhs, rhs),
+      Expression::Join { lhs, rhs } => write!(
+        f,
+        "{} / {}",
+        lhs
+          .as_ref()
+          .unwrap_or(&Box::new(Expression::empty_string_literal())),
+        rhs
+      ),
       Expression::Concatenation { lhs, rhs } => write!(f, "{} + {}", lhs, rhs),
       Expression::Conditional {
         lhs,
--- a/src/node.rs
+++ b/src/node.rs
@@ -118,14 +118,11 @@ impl<'src> Node<'src> for Expression<'src> {
       } => Tree::string(cooked),
       Expression::Backtick { contents, .. } => Tree::atom("backtick").push(Tree::string(contents)),
       Expression::Group { contents } => Tree::List(vec![contents.tree()]),
-      Expression::Join { lhs, rhs } => Tree::atom("/")
-        .push(
-          lhs
-            .as_ref()
-            .unwrap_or(&Box::new(Expression::empty_string_literal()))
-            .tree(),
-        )
-        .push(rhs.tree()),
+      Expression::Join { lhs: None, rhs } => Tree::atom("/").push(rhs.tree()),
+      Expression::Join {
+        lhs: Some(lhs),
+        rhs,
+      } => Tree::atom("/").push(lhs.tree()).push(rhs.tree()),
     }
   }
 }
--- a/src/node.rs
+++ b/src/node.rs
@@ -118,7 +118,14 @@ impl<'src> Node<'src> for Expression<'src> {
       } => Tree::string(cooked),
       Expression::Backtick { contents, .. } => Tree::atom("backtick").push(Tree::string(contents)),
       Expression::Group { contents } => Tree::List(vec![contents.tree()]),
-      Expression::Join { lhs, rhs } => Tree::atom("/").push(lhs.tree()).push(rhs.tree()),
+      Expression::Join { lhs, rhs } => Tree::atom("/")
+        .push(
+          lhs
+            .as_ref()
+            .unwrap_or(&Box::new(Expression::empty_string_literal()))
+            .tree(),
+        )
+        .push(rhs.tree()),
     }
   }
 }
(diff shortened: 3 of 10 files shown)


src/parser.rs
```diff
@@ -405,18 +405,27 @@ impl<'tokens, 'src> Parser<'tokens, 'src> {
...
-      let value = self.parse_value()?;
-
+      let value = self.parse_value();
       if self.accepted(Slash)? {
+        // Enables Slash operator to build absolute paths, like `/ "Users"`, by
+        // setting a default empty string on the lhs if the resulting expression
+        // is not Ok.
+        let value = value.unwrap_or(Expression::StringLiteral {
+          string_literal: StringLiteral {
+            kind: StringKind::new(string_kind::StringDelimiter::QuoteDouble, false),
```
erikkrieg: I also noticed that something like this is possible:

```rs
kind: StringKind::from_token_start("\"").unwrap()
```

I didn't do it, however, because it looked a bit hacky to me, although it would mean that nothing extra from the `string_kind` module would need to be made public.

erikkrieg:
Hey, @casey. Anything I am missing or should do to make this ready for review?

src/parser.rs
```diff
@@ -405,18 +405,27 @@ impl<'tokens, 'src> Parser<'tokens, 'src> {
...
       self.parse_conditional()?
     } else {
-      let value = self.parse_value()?;
-
+      let value = self.parse_value();
       if self.accepted(Slash)? {
+        // Enables Slash operator to build absolute paths, like `/ "Users"`, by
+        // setting a default empty string on the lhs if the resulting expression
+        // is not Ok.
+        let value = value.unwrap_or(Expression::StringLiteral {
```
casey: I think this `unwrap_or` will swallow errors. What if `self.parse_value()` returns an error after skipping over some tokens?

Maybe something like this instead?

```rust
    let expression = if self.accepted_keyword(Keyword::If)? {
      self.parse_conditional()?
    } else if self.accepted(Slash)? {
        let lhs = None;
        let rhs = Box::new(self.parse_expression()?);
        Expression::Join { lhs, rhs }
    } else {
      let value = self.parse_value()?;
      
      if self.accepted(Slash)? {
        let lhs = Some(Box::new(value));
        let rhs = Box::new(self.parse_expression()?);
        Expression::Join { lhs, rhs }
      } else if self.accepted(Plus)? {
        let lhs = Box::new(value);
        let rhs = Box::new(self.parse_expression()?);
        Expression::Concatenation { lhs, rhs }
      } else {
        value
      }
    };
```

src/parser.rs
casey: You can avoid needing to construct an empty string token by letting `lhs` be an `Option`:

```
  Join {
    lhs: Option<Box<Expression>>,
    rhs: Box<Expression>,
  },
```

casey:
Sorry for the delay! I was out of town for the weekend, and just now catching up to my GitHub notifications. Thanks for taking this on! I left some comments on the code. The `unwrap_or` will might leave the parser in a bad state, so best to avoid it, and I think it's cleaner to make `lhs` an `Option` instead of the empty string.

I'd also like to see a few more tests. For example, in what context could the `unwrap_or` cause problems? Is there a test that would have caught it?

src/parser.rs
erikkrieg: Okay, this is closer to what I first imagined, but I didn't realize `Expression::Join` would accept an option (I shoulda looked further into that).

casey: It doesn't actually accept an option, but you can change it in this PR so it does!

erikkrieg:
Thanks for the feedback! I'll take a look tonight.

erikkrieg:
I went on quite the journey trying to make `lhs` an optional expression for `Join` before putting a pin in that approach and going back to something close to my first pass. I went through the following files while changing `lhs` to an option:
- src/assignment_resolver.rs
- src/evaluator.rs
- src/expression.rs
- src/node.rs
- src/parser.rs
- src/string_kind.rs
- src/summary.rs
- src/variables.rs

The number of places to make the change wasn't so much a problem (tho a bit messy, I think) but the real issue was that I hit a snag with lifetimes on a borrow in `src/variables.rs` when needing to push a ref to a boxed expression onto the stack. In the case where `lhs.is_none()` I needed to provide an expression containing a string literal and I couldn't figure out a way to create one in a way that satisfied the lifetime requirements.

So I came back around full circle, except without the potentially breaking `unwrap` on `parse_value`.

casey:
How about this:

```
        // if `lhs` is `Some`:
        Expression::Concatenation { lhs, rhs } | Expression::Join { Some(lhs), rhs } => {
          self.stack.push(rhs);
          self.stack.push(lhs);
        }
       // If `lhs` is `None`:
        Expression::Join { None, rhs } => {
          self.stack.push(rhs);
        }
```

Variables is an iterator that must return all variables (and thus traverse all sub-expressions) in an expression. If `lhs` is `None`, it contains no variables, so nothing needs to be pushed onto the stack.

src/parser.rs
```diff
@@ -1989,6 +1993,7 @@ mod tests {
         Identifier,
         ParenL,
         ParenR,
+        Slash,
```
erikkrieg, line 1998: I wasn't able to confidently understand what these tests mean. My best guess was that these are the tokens that can be found at the beginning of an expression without causing an exception.

This was the output from one of the two failed tests:
```
---- parser::tests::unclosed_parenthesis_in_interpolation stdout ----
thread 'parser::tests::unclosed_parenthesis_in_interpolation' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
 CompileError {
     token: Token {
         offset: 15,
         length: 2,
         line: 1,
         column: 12,
         src: "a:\n echo {{foo(}}",
         kind: InterpolationEnd,
     },
     kind: UnexpectedToken {
         expected: [
             Backtick,
             Identifier,
             ParenL,
             ParenR,
<            Slash,
             StringToken,
         ],
         found: InterpolationEnd,
     },
 }

', src/parser.rs:907:9
```

erikkrieg: All that said, I am wondering if my understanding was correct and whether adding `Slash` to the list was the appropriates change.

casey: Yup, I think that's correct. When the parser hits an unexpected token, the error message includes which tokens might have been accepted. Since `/` can now appear at the beginning of an expression, the error message now includes `/`, so adding it to the `expected` token list is correct.

erikkrieg:
Okay, your suggestion for variables makes sense. I have an another branch locally with those changes, so I can give that a shot and see if that was truly the last compiler error for the option approach.

The question, then, is do you prefer the `None` case being handled in all those modules downstream of the parser? Or creating an empty string literal expression at the parser and leaving the rest unchanged?

casey:
I like handling the `None` case better, it seems more explicit.

src/summary.rs
```diff
@@ -258,7 +258,11 @@ impl Expression {
         rhs: Box::new(Expression::new(rhs)),
       },
       Join { lhs, rhs } => Expression::Join {
-        lhs: Box::new(Expression::new(lhs)),
+        lhs: if lhs.is_some() {
+          Some(Box::new(Expression::new(lhs.as_ref().unwrap())))
+        } else {
+          None
+        },
```
erikkrieg: I tried cleaning this up a bit by using the `if let Some()` syntax here but was getting some odd warnings and hints from rust-analyzer.

Screen Shot 2022-09-02 at 12 02 03 AM

I was able to pass the tests using that syntax but held off on committing that version. Not sure if there is an issue with the Rust LSP implementation or an actual issue with using that syntax in this context.

wezm: The suggestion to use [`Option::map`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map) that hint makes (which comes from [clippy](https://github.com/rust-lang/rust-clippy)) is probably what you want since it's equivalent and more succinct:

```rust
lhs: lhs.as_ref().map(|lhs| Box::new(Expression::new(lhs)))
```

erikkrieg: Hmm, if this handles the `None` case then I suppose it is more succinct, but I'd also argue more cryptic. I'll test it out, though. And if it works, I am happy to use that, if it is preferred.

erikkrieg: Yup, that does it. Haven't come across the method on Option before. Was also a bit thrown off because I am more used to thinking of `map` as a method for lists.

wezm: > I'll test it out, though. And if it works, I am happy to use that, if it is preferred.

I'd say it's very common in Rust code along with the other combinatorial methods on `Option` and `Result` (`and_then`, `or_else`, `map_err`, etc.)

erikkrieg: I ended up reading through the rest of the Option methods to see if there was anything else I could pare down. Tried using `and_then` in a couple places, but because of the Results being unwrapped with `?` it didn't end up looking like an improvement to me (though I definitely might have missed something).

src/evaluator.rs
```diff
@@ -177,7 +177,12 @@ impl<'src, 'run> Evaluator<'src, 'run> {
       }
       Expression::Group { contents } => self.evaluate_expression(contents),
       Expression::Join { lhs, rhs } => {
-        Ok(self.evaluate_expression(lhs)? + "/" + &self.evaluate_expression(rhs)?)
+        let lhs = if let Some(lhs) = lhs {
+          self.evaluate_expression(lhs)?
+        } else {
+          String::new()
+        };
```
wezm: Minor, feel free to ignore. You might consider `match` for this one:

```rust
        let lhs = match lhs {
            Some(lhs) => self.evaluate_expression(lhs)?,
            None => String::new(),
        };
```

erikkrieg: I am down with it 👍

src/evaluator.rs
```diff
@@ -177,7 +177,11 @@ impl<'src, 'run> Evaluator<'src, 'run> {
       }
       Expression::Group { contents } => self.evaluate_expression(contents),
       Expression::Join { lhs, rhs } => {
-        Ok(self.evaluate_expression(lhs)? + "/" + &self.evaluate_expression(rhs)?)
+        let lhs = match lhs {
```
casey: How about:

```
Expression::Join { None, rhs } => {
  Ok("/" + self.evaluate_expression(lhs)?) // not sure about the + here, might need String::from("/") + or something else
}
Expression::Join { Some(lhs), rhs } => {
  Ok(lhs + "/" + &self.evaluate_expression(rhs)?)
}
```

In general, I think it's cleaner to just handle the cases separately, rather than creating an empty string to join with.

src/expression.rs
```diff
@@ -45,13 +45,30 @@ impl<'src> Expression<'src> {
   pub(crate) fn variables<'expression>(&'expression self) -> Variables<'expression, 'src> {
     Variables::new(self)
   }
+
+  pub(crate) fn empty_string_literal() -> Self {
```
casey: Creating empty string literals should be avoided.

casey: I think it's less clear to create empty string literals, rather than to handle the none case separately.

src/node.rs
```diff
@@ -118,7 +118,14 @@ impl<'src> Node<'src> for Expression<'src> {
       } => Tree::string(cooked),
       Expression::Backtick { contents, .. } => Tree::atom("backtick").push(Tree::string(contents)),
       Expression::Group { contents } => Tree::List(vec![contents.tree()]),
-      Expression::Join { lhs, rhs } => Tree::atom("/").push(lhs.tree()).push(rhs.tree()),
+      Expression::Join { lhs, rhs } => Tree::atom("/")
```
casey, line 121: How about:

```
Expression::Join { None, rhs } => write!(
  f,
  "/ {}",
  rhs
),
Expression::Join { Some(lhs), rhs } => write!(
  f,
  "{} / {}",
  lhs,
  rhs
),
```

src/expression.rs
```diff
@@ -45,13 +45,30 @@ impl<'src> Expression<'src> {
...
+    }
+  }
 }
 
 impl<'src> Display for Expression<'src> {
   fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
     match self {
       Expression::Backtick { token, .. } => write!(f, "{}", token.lexeme()),
-      Expression::Join { lhs, rhs } => write!(f, "{} / {}", lhs, rhs),
+      Expression::Join { lhs, rhs } => write!(
```
casey: How about:

```
Expression::Join { None, rhs } => write!(
  f,
  "/ {}",
  rhs
),
Expression::Join { Some(lhs), rhs } => write!(
  f,
  "{} / {}",
  lhs,
  rhs
),
```

src/node.rs
```diff
@@ -118,7 +118,14 @@ impl<'src> Node<'src> for Expression<'src> {
...
+        .push(
+          lhs
+            .as_ref()
+            .unwrap_or(&Box::new(Expression::empty_string_literal()))
+            .tree(),
+        )
+        .push(rhs.tree()),
```
casey:
```suggestion
Expression::Join { None, rhs } => Tree::atom("/")
  .push(rhs.tree()),
Expression::Join { Some(lhs), rhs } => Tree::atom("/")
  .push(lhs.tree()),
  .push(rhs.tree()),
```

casey:
Added some more comments. Also, some tests using prefix `/` in different contexts would be good.

src/evaluator.rs
erikkrieg: Ah, so, I had tried this earlier but it wasn't working. Taking a look at it again I see that the enum can be used in the match statement like this: `Expression::Join { lhs: None, rhs } => ()`

erikkrieg:
> Added some more comments. Also, some tests using prefix `/` in different contexts would be good.

I added tests for when lhs is `None` that seem to mirror the pre-existing tests for the slash operator. I also added a test for  the error case when rhs values/expression is not provided. Are there any other tests besides what are in the `slash_operator` test module that I should be looking at?

src/string_kind.rs
```diff
@@ -7,7 +7,7 @@ pub(crate) struct StringKind {
 }
 
 #[derive(Debug, PartialEq, Clone, Copy, Ord, PartialOrd, Eq)]
-enum StringDelimiter {
+pub(crate) enum StringDelimiter {
```
casey: Can `pub(crate)` be removed now?

src/string_kind.rs
```diff
@@ -26,7 +26,7 @@ impl StringKind {
     Self::new(StringDelimiter::QuoteSingle, false),
   ];
 
-  const fn new(delimiter: StringDelimiter, indented: bool) -> Self {
+  pub(crate) const fn new(delimiter: StringDelimiter, indented: bool) -> Self {
```
casey: Can `pub(crate)` be removed now?

casey:
These look good! And the PR looks good too, I think just a couple of pub(crate)s that aren't necessary.

It's really interesting that now `// "foo"` is valid, and is basically `"" / "" / "foo"`.

I'm not sure if that should be valid syntax, the thought being that maybe in the future, we want to use `//` for something else, and if we allow `// "foo"` now, then using `//` for something else would be a breaking change.

If, god forbid, I ever have to implement support for arithmetic in Just, then it would be pretty natural to use `//` for division. Also, I think it's pretty uncommon to want double slashes, although I suppose you could do: "https:" // "ordinals.com" / "page" to build a URL. Also, windows has UNC paths, which start with `//`, so you could use `//` there.

What do you think?

Either way, let's not hold this PR up. Even if we decide to make `//` illegal, we can merge this PR, and I can add a PR that changes `//` into an error, as long as I do it before I cut a release.

src/string_kind.rs
erikkrieg: Yeah. I had this file marked as "viewed" in the pr UI and missed it when cleaning everything up.

erikkrieg:
That's a good question... 

**TL;DR:** I think `//"foo"` should be a valid expression.

Without having very intimate knowledge of Just's internals, here are a few ideas for arithmetic:
1. Treat operators, like `+`, `*` and `/`, differently for ints and floats so that arithmetic can be written with a syntax familiar to many other popular languages. 
2. Use functions for arithmetic, like `div(4,2)`, `add(1,2)`, `mul(2,2)`. It is more verbose, but very transparent.
3. Wrap expressions to indicate arithmetic should be performed, i.e. `[ 4 / 2 ]` works but `4 / 2` does not.

I am a less in favour of using `//` for division because I feel it is more likely to lead to confusion writing and reading math equations in Justfiles compared to the other approaches. But because I am not sure how much effort each of these options might take, I am not really considering that dimension in my ranking.

Now, that doesn't answer you main question, but I think it might be enough to consider reserving `//` **not** worthwhile enough to affect the decision much either way. So then the question maybe become whether it is useful to have `//"foo"` evaluate to `"//foo"`. I suppose the simple answer is "yes", and you gave a couple cases. It also seems a bit more intuitive to me, but then again, I have been looking at this feature for a few days, so take that with a heap of salt.

casey:
That all sounds pretty reasonable. Also, if we have arithmetic, we'll also have static typing, so that could be used to disambiguate joins from division.

I just merged master into this branch, and assuming the tests pass, is this ready to merge?

erikkrieg:
> is this ready to merge?

Yes, it is!

Appreciate all the feedback and patience.

casey:
Sorry for the delay! Just added some docs. Thanks for sticking with this, this is a nice improvement.

PR #1335, opened 2022-09-08
Refactor run.rs

neunenak:
This removes mutation of the color and verbosity variables and makes it slightly easier to figure out what's going on here.

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -1,5 +1,6 @@
 use super::*;
 
+/// Main entry point into just.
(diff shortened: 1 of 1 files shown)


src/run.rs
```diff
@@ -1,5 +1,6 @@
 use super::*;
 
+/// Main entry point into just.
```
casey:
```suggestion
/// Main entry point into just binary.
```

casey:
Nice, merging. LGTM!

PR #1336, opened 2022-09-08
Fix issues reported by clippy

neunenak:
Fixes every issue reported by running vanilla-configuration `cargo clippy`.

### DIFF
--- a/src/assignment_resolver.rs
+++ b/src/assignment_resolver.rs
@@ -45,10 +45,7 @@ impl<'src: 'run, 'run> AssignmentResolver<'src, 'run> {
         length: 0,
         kind: TokenKind::Unspecified,
       };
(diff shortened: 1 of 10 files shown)


casey:
How can I get these errors? When I run `cargo clippy` it doesn't complain.

neunenak:
Huh, I just realized, I had my default toolchain set to nightly, not stable. It looks like these are all lints that are only reported on nightly clippy.

neunenak:
I just tested it and this branch compiles and passes all tests on stable as well.

casey:
Okay, sweet. I was worried I had something weird going on with the config.

casey:
Actually, looking at these, they all look like reasonable fixes, so why not merge them.

Issue #1324, opened 2022-08-29
Look into chumsky


casey:
Error recovery is very difficult, and just's parser doesn't try to do this at all. [zesterer/chumsky](https://github.com/zesterer/chumsky/) is a parser combinator crate that supports this. I have a feeling that just's parser is good enough that we wouldn't profit much by replacing it, but it's worth considering.

neunenak:
Chumsky's a parser library I've been meaning to look into myself. I wonder if it would make implementing https://github.com/casey/just/pull/1272 easier (which I hope to get back to soon)

casey:
I never quite know if the current parser is reasonably good all things considered, or a heinous mess. One thing that is fiddly in the current parser is parsing recipe bodies. It would be pretty painful to replace, and might be hard to replicate with a more principled parser library.

PR #1343, opened 2022-09-13
Chumsky parser

neunenak:
Use the chumsky library to parse justfiles. Cf. https://github.com/casey/just/issues/1324


casey:
I took another look at this, and I think that although a structured approach has advantages, overall I think I like the current parser.

PR #1354, opened 2022-09-25
Suppress command exit code error messages

### LINKED ISSUE #1334: Supress the "error: Recipe `...` failed on line ... with exit code ..." without suppressing the exit code
If you're wrapping some binary CLI with `just`, e.g.

```sh
$ just some-command --some --flags
```

It will often return a non-zero exit code (e.g. 2) while you trying to browse through some arguments (especially if there's incomplete subcommands that have further nested subcommands etc). It's pretty annoying to have the `error: Recipe `...` failed on line ... with exit code ...` message pop up every time, making it feel somewhat alien.

Suppressing the error is not the right solution though, because it should remain to be a non-zero exit code! (especially if called from some scripts)

Is there a way to make `just` even more transparent and a bit more quiet? I.e., don't suppress exit codes but don't print anything extra to the output, at all (unless there's some fatal errors in the justfile itself).

// Apologies in advance if I'm missing something obvious here but I couldn't find anything except `@` and `-`.

gokhanettin:
When a recipe wraps cli tool and the tool exits with a non-zero code,
just adds its own extra exit error message along with the messages
from the tool. Introduce `no-exit-message` attribute to suppress this
additional message.

Closes #1334

### DIFF
--- a/README.md
+++ b/README.md
@@ -1987,6 +1987,35 @@ echo 'Bar!'
 Bar!
 ```
 
+`just` normally prints error messages when a recipe line fails. These error
+messages can be suppressed using the `[no-exit-message]` attribute. You may find
+this especially useful with a recipe that recipe wraps a tool:
+
+```make
+git *args:
+    @git {{args}}
+```
+
+```sh
+$ just git status
+fatal: not a git repository (or any of the parent directories): .git
+error: Recipe `git` failed on line 2 with exit code 128
+```
+
+Add the attribute to suppress the exit error message when the tool exits with a
+non-zero code:
+
+```make
+[no-exit-message]
+git *args:
+    @git {{args}}
+```
+
+```sh
+$ just git status
+fatal: not a git repository (or any of the parent directories): .git
+```
+
 ### Selecting Recipes to Run With an Interactive Chooser
 
 The `--choose` subcommand makes `just` invoke a chooser to select which recipes to run. Choosers should read lines containing recipe names from standard input and print one or more of those names separated by spaces to standard output.
--- a/src/attribute.rs
+++ b/src/attribute.rs
@@ -1,0 +1,13 @@
+use super::*;
+
+#[derive(EnumString)]
+#[strum(serialize_all = "kebab_case")]
+pub(crate) enum Attribute {
+  NoExitMessage,
+}
+
+impl Attribute {
+  pub(crate) fn from_name(name: Name) -> Option<Attribute> {
+    name.lexeme().parse().ok()
+  }
+}
--- a/src/compile_error.rs
+++ b/src/compile_error.rs
@@ -248,7 +248,7 @@ impl Display for CompileError<'_> {
         write!(f, "Alias `{}` has an unknown target `{}`", alias, target)?;
       }
       UnknownAttribute { attribute } => {
-        write!(f, "Unknown attribute `{attribute}`")?;
+        write!(f, "Unknown attribute `{}`", attribute)?;
       }
       UnknownDependency { recipe, unknown } => {
         write!(
--- a/src/compile_error.rs
+++ b/src/compile_error.rs
@@ -247,6 +247,9 @@ impl Display for CompileError<'_> {
       UnknownAliasTarget { alias, target } => {
         write!(f, "Alias `{}` has an unknown target `{}`", alias, target)?;
       }
+      UnknownAttribute { attribute } => {
+        write!(f, "Unknown attribute `{attribute}`")?;
+      }
       UnknownDependency { recipe, unknown } => {
         write!(
           f,
--- a/src/compile_error_kind.rs
+++ b/src/compile_error_kind.rs
@@ -98,6 +98,9 @@ pub(crate) enum CompileErrorKind<'src> {
     alias: &'src str,
     target: &'src str,
   },
+  UnknownAttribute {
+    attribute: &'src str,
+  },
   UnknownDependency {
     recipe: &'src str,
     unknown: &'src str,
--- a/src/error.rs
+++ b/src/error.rs
@@ -35,6 +35,7 @@ pub(crate) enum Error<'src> {
     recipe: &'src str,
     line_number: Option<usize>,
     code: i32,
+    suppress_message: bool,
   },
   CommandInvoke {
     binary: OsString,
@@ -166,6 +167,16 @@ impl<'src> Error<'src> {
     Self::Internal {
       message: message.into(),
     }
+  }
+
+  pub(crate) fn suppress_message(&self) -> bool {
+    matches!(
+      self,
+      Error::Code {
+        suppress_message: true,
+        ..
+      }
+    )
   }
 }
 
@@ -323,6 +334,7 @@ impl<'src> ColorDisplay for Error<'src> {
         recipe,
         line_number,
         code,
+        ..
       } => {
         if let Some(n) = line_number {
           write!(
--- a/src/justfile.rs
+++ b/src/justfile.rs
@@ -472,11 +472,13 @@ mod tests {
       recipe,
       line_number,
       code,
+      suppress_message,
     },
     check: {
       assert_eq!(recipe, "a");
       assert_eq!(code, 200);
       assert_eq!(line_number, None);
+      assert_eq!(suppress_message, false);
     }
   }
 
@@ -491,11 +493,13 @@ mod tests {
       recipe,
       line_number,
       code,
+      suppress_message,
     },
     check: {
       assert_eq!(recipe, "fail");
       assert_eq!(code, 100);
       assert_eq!(line_number, Some(2));
+      assert_eq!(suppress_message, false);
     }
   }
 
@@ -510,11 +514,13 @@ mod tests {
       recipe,
       line_number,
       code,
+      suppress_message,
     },
     check: {
       assert_eq!(recipe, "a");
       assert_eq!(code, 150);
       assert_eq!(line_number, Some(2));
+      assert_eq!(suppress_message, false);
     }
   }
 
@@ -664,13 +670,15 @@ mod tests {
     "#,
     args: ["--quiet", "wut"],
     error: Code {
-      line_number,
       recipe,
+      line_number,
+      suppress_message,
       ..
     },
     check: {
       assert_eq!(recipe, "wut");
       assert_eq!(line_number, Some(7));
+      assert_eq!(suppress_message, false);
     }
   }
 
--- a/src/justfile.rs
+++ b/src/justfile.rs
@@ -478,7 +478,7 @@ mod tests {
       assert_eq!(recipe, "a");
       assert_eq!(code, 200);
       assert_eq!(line_number, None);
-      assert_eq!(suppress_message, false);
+      assert!(!suppress_message);
     }
   }
 
@@ -499,7 +499,7 @@ mod tests {
       assert_eq!(recipe, "fail");
       assert_eq!(code, 100);
       assert_eq!(line_number, Some(2));
-      assert_eq!(suppress_message, false);
+      assert!(!suppress_message);
     }
   }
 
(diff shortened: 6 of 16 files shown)


casey:
I didn't take a close look, but this looks good. Thanks for writing tests! Let's figure out if this flag should also suppress command echoing.

casey:
Nice, let me know when this is ready to review!

gokhanettin:
@casey, I think it is ready to review. We can update the docs when we agree on the implementation.

I have a couple of concerns here. If we need to add multiple annotations, we are likely to support a syntax like `[no-exit-message, annotation2, annotation3]` which we can't represent with `annotation: Option<Annotation<'src>>` any more. It will probably be something like `annotations: BTreeSet<Annotation<'src>>`. This won't be backward compatible with the json representation. Do you think we should now have `annotations: BTreeSet<Annotation<'src>>` within the recipe struct even though we currently have only one annotation type?

casey:
> @casey, I think it is ready to review. We can update the docs when we agree on the implementation.

Nice! I'm just heading out of town for the weekend, so I won't be able to look at this until next week when I get back. Apologies for the delay.

> I have a couple of concerns here. If we need to add multiple annotations, we are likely to support a syntax like `[no-exit-message, annotation2, annotation3]` which we can't represent with `annotation: Option<Annotation<'src>>` any more. It will probably be something like `annotations: BTreeSet<Annotation<'src>>`. This won't be backward compatible with the json representation. Do you think we should now have `annotations: BTreeSet<Annotation<'src>>` within the recipe struct even though we currently have only one annotation type?

Good thinking. One note: The JSON dump format is unstable, i.e., requires the `--unstable` flag, so it can be changed in backwards incompatible ways if necessary. Although I agree it's a good idea to avoid that if possible.

I actually think that recipes probably shouldn't have a `BTreeSet` of annotations, but rather they should have a `suppress_exit_error_messages: bool` field. My thinking is that by the time you have a `Recipe` struct, the actual annotations aren't important, just their semantics.

casey (changes requested):
This is looking good, thanks for all the tests!

A few high level thoughts:

- I wasn't sure whether we should call these "attributes" or "annotations", so I did a google search for `programming "function attribute"` and `programming "function annotation"`, and the search with "attribute" got 8x more hits, so I think we should go with that and call them "attributes" instead of annotations.

- Since the json now just includes a single `suppress_exit_error_messages` boolean, which will be backwards compatible with future additions, I think we should avoid parsing the comma-separate attribute list syntax, i.e., `[a,b,c]`, for now. We can easily add it in the future if we need it.

- Since an attribute is required to be directly in front of a recipe, instead of popping an attribute off the list of items, in the main parser loop, we can parse an attribute if we see a `[`, and then immediatly parse a recipe:

  ```rust
  } else if self.accepted(BracketL)? {
    items.push(Item::AnnotationContainer(self.parse_annotations()?));
    self.expect_eol()?;
    self.parse_recipe()?
  }
  ```

  This avoids the need to check for spurious newlines or make the `pop_doc_comment` function more complicated.

- I think the signature of `Parser::parse_atribute` should be `fn parse_attribute(&mut self) -> Result<Attribute>`, and attribute can be:

  ```rust
  enum Attribute {
    NoExitMessage
  }
  ```

  This is nice because the parser returns a single, strongly-typed attribute, so downstream code doesn't need to worry about checking for invalid attributes.

gokhanettin:
Thanks for the review. I agree with all the thoughts. I will be back with the changes in a couple of days.

README.md
```diff
@@ -1985,6 +1985,37 @@ echo 'Bar!'
 Bar!
 ```
 
+Additional exit error messages from recipes can be supressed using
+`no-exit-message` attribute. You may find this especially useful when your
+recipe wraps a tool:
```
```suggestion
`just` normally prints error messages when a recipe line fails. These error messages can be
suppressed using the `[no-exit-message]` attribute. You may find this especially useful with
a recipe that recipe wraps a tool:
```

README.md
```diff
@@ -1985,6 +1985,37 @@ echo 'Bar!'
...
+
+```make
+# Wraps git
```
```suggestion
```

README.md
```diff
@@ -1985,6 +1985,37 @@ echo 'Bar!'
...
+git *args:
+    @git {{args}}
+```
+
+```sh
+$ just git status
+fatal: not a git repository (or any of the parent directories): .git
+error: Recipe `git` failed on line 3 with exit code 128
```
```suggestion
error: Recipe `git` failed on line 2 with exit code 128
```

README.md
```diff
@@ -1985,6 +1985,37 @@ echo 'Bar!'
...
+```
+
+Add the attribute to suppress the exit error message when the tool exits with a
+non-zero code:
+
+```make
+# Wraps git
```
```suggestion
```

src/attribute.rs
```diff
@@ -0,0 +1,13 @@
...
+#[derive(EnumString)]
+#[strum(serialize_all = "kebab_case")]
+pub(crate) enum Attribute {
+  NoExitMessage,
+}
+
+impl Attribute {
+  pub(crate) fn from_lexeme(lexeme: &str) -> Option<Attribute> {
+    lexeme.parse().ok()
+  }
```
casey: I think this is a little clearer, since attributes should only be parsed from naems:

```suggestion
  pub(crate) fn from_name(lexeme: Name) -> Option<Attribute> {
    name.lexeme().parse().ok()
  }
```

src/compile_error.rs
```diff
@@ -247,6 +247,9 @@ impl Display for CompileError<'_> {
       UnknownAliasTarget { alias, target } => {
         write!(f, "Alias `{}` has an unknown target `{}`", alias, target)?;
       }
+      UnknownAttribute { attribute } => {
+        write!(f, "Attribute `{}` is unknown", attribute)?;
```
```suggestion
        write!(f, "Unknown attribute `{attribute}`")?;
```

src/compile_error_kind.rs
```diff
@@ -98,6 +98,9 @@ pub(crate) enum CompileErrorKind<'src> {
     alias: &'src str,
     target: &'src str,
   },
+  UnknownAttribute {
+    attribute: &'src str,
```
casey, line 102:
```suggestion
    attribute: Name<'src>,
```

src/error.rs
```diff
@@ -167,6 +168,15 @@ impl<'src> Error<'src> {
...
   }
+
+  pub(crate) fn suppress_message(&self) -> bool {
+    match self {
+      Error::Code {
+        suppress_message, ..
+      } => *suppress_message,
+      _ => false,
+    }
+  }
```
casey: I'm not 100% sure this is the right syntax, but it should be possible to simplify this like so:

```suggestion
    matches!(self, Error::Code { suppress_message: true, ..})
```

src/justfile.rs
```diff
@@ -472,6 +472,7 @@ mod tests {
       recipe,
       line_number,
       code,
+      ..
```
casey: It's more verbose, but I think I'd extract and check these with an assert below:

```suggestion
      suppress_error_message,
```

src/parser.rs
```diff
@@ -345,13 +345,19 @@ impl<'tokens, 'src> Parser<'tokens, 'src> {
...
       } else if self.accepted(At)? {
         let doc = pop_doc_comment(&mut items, eol_since_last_comment);
-        items.push(Item::Recipe(self.parse_recipe(doc, true)?));
+        items.push(Item::Recipe(self.parse_recipe(doc, true, false)?));
+      } else if self.accepted(BracketL)? {
+        self.parse_attribute()?;
+        self.expect_eol()?;
+        let quiet = if self.accepted(At)? { true } else { false };
+        let doc = pop_doc_comment(&mut items, eol_since_last_comment);
+        items.push(Item::Recipe(self.parse_recipe(doc, quiet, true)?));
```
casey: I think it's a little confusing that `parse_attribute` consumes the `BracketR` but not the `BracketL`. How about something like this:

(Also, `if CONDITION { true } else { false }` is equivalent to `CONDITION`.)

```suggestion
      } else if self.accepted(BracketL)? {
        let attribute_name = self.parse_attribute_name()?
        self.expect(BracketR)?;
        self.expect_eol()?;
        let quiet = self.accepted(At)?;
        let doc = pop_doc_comment(&mut items, eol_since_last_comment);
        items.push(Item::Recipe(self.parse_recipe(doc, quiet, true)?));
```

src/parser.rs
```diff
@@ -816,6 +824,18 @@ impl<'tokens, 'src> Parser<'tokens, 'src> {
...
+  fn parse_attribute(&mut self) -> CompileResult<'src, Attribute> {
+    let token = self.expect(Identifier)?;
+    let lexeme = token.lexeme();
+    if let Some(attribute) = Attribute::from_lexeme(lexeme) {
+      self.expect(BracketR)?;
+      Ok(attribute)
+    } else {
+      Err(token.error(CompileErrorKind::UnknownAttribute { attribute: lexeme }))
+    }
+  }
```
casey: This is not quite right, since you also need the token to produce the error message, but something like this:

```suggestion
  /// Parse a recipe attribute name
  fn parse_attribute_name(&mut self) -> CompileResult<'src, Attribute> {
    let name = self.parse_name()?;
    Attribute::from_lexeme(name).ok_or_else(|| token.error(CompileErrorKind::UnknownAttribute { attribute: lexeme }))
    }
```

casey (changes requested):
Nice this is great! A few comments, but it's all small stuff.

src/compile_error_kind.rs
gokhanettin: Other `CompileErrorKind`s like `UnknownFunction`, `UnknownSetting` all have `&'src str`. I thought there was nothing special about the attribute. Considering your `parse_attribute_name` suggestion, do you find anything wrong or ugly with the code below @casey?

```rs
...
UnknownAttribute {
  attribute: &'src str,
},
```

```rs
  /// Parse a recipe attribute name
  fn parse_attribute_name(&mut self) -> CompileResult<'src, Attribute> {
    Attribute::from_name(name).ok_or_else(|| {
      name.error(CompileErrorKind::UnknownAttribute {
        attribute: name.lexeme(),
      })
    })
  }
```

I think we should have `&'src str` here. `CompileError` returned by  `error()` method already has a token field. We should not have another `Token` or `Name` type field within the `CompileErrorKind`.

```rs
pub(crate) struct CompileError<'src> {
  pub(crate) token: Token<'src>,
  pub(crate) kind: Box<CompileErrorKind<'src>>,
}
```

casey: Great point, makes sense to me!

gokhanettin:
I am not sure about `CompileErrorKind::UnknownAttribute` field type. Pushed all other suggestions. Also need to update GRAMMER.md.

Is this OK?

```md
recipe        : attribute? '@'? NAME parameter* variadic? ':' dependency* body?

attribute     : '[' NAME ']' eol
```

src/parser.rs
```diff
@@ -345,13 +345,20 @@ impl<'tokens, 'src> Parser<'tokens, 'src> {
...
+              items.push(Item::Recipe(self.parse_recipe(doc, false, false)?));
             }
           }
         }
       } else if self.accepted(At)? {
         let doc = pop_doc_comment(&mut items, eol_since_last_comment);
-        items.push(Item::Recipe(self.parse_recipe(doc, true)?));
+        items.push(Item::Recipe(self.parse_recipe(doc, true, false)?));
+      } else if self.accepted(BracketL)? {
+        self.parse_attribute_name()?;
```
casey: This works, but it's a bit unclear what's going on. I'd do something like:

```rust
let Attribute::NoExitMessage = self.parse_atribute_name()?;
```

This makes it clear what's going on, and it will turn into a compile error when we add another variant.

casey (changes requested):
One minor comment, otherwise looks great. Your suggestion for representing attributes in the grammar looks good to me.

casey:
Very nice! Great work, especially for a non-trivial PR that touches the parser!

PR #1358, opened 2022-10-05
Don't evaluate comments

casey:
This is a follow-up to #1333. In addition to not printing out comments, it doesn't evaluate the contents of interpolations inside comments. @neunenak, what do you think?

### DIFF
--- a/src/line.rs
+++ b/src/line.rs
@@ -1,5 +1,8 @@
 use super::*;
 
+// TODO
+// - don't evaluate comments
+
 /// A single line in a recipe body, consisting of any number of `Fragment`s.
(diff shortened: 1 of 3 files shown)


neunenak:
Huh, I didn't realize interpolations inside comments would get evaluated. Yeah, this seems like a good change.

casey:
Unfortunately they'll still get parsed, so you'll get parse errors. However, fixing that is a backwards-compatible change. However it'll be a little hairy. We'll have to either extract setting values in a first pass (not actually that hard since settings are simple) and use the settings to change how the parser works, or stipulate that the setting only affects parsing of recipes after the setting.

PR #1362, opened 2022-10-08
Add kitchen sink Justfile to test syntax highlighting

nk9:
I've added a file with all the different features I could think of, plus a line prefix in the GRAMMAR.md. The justfile is pretty messy, I mostly just copied things out of the docs while changing names as needed to get stuff parsing properly. I've definitely found some issues with the syntax highlighing during this exercise, so that's great!


casey:
Awesome, this is great! `test-suite.just` seems a little generic, so I renamed it to `kitchen-sink.just`.

PR #1368, opened 2022-10-13
Require `set fallback := true` to enable recipe fallback

casey:


nk9:
We should add `fallback` to the Sublime syntax, too. Line 185.

Issue #1300, opened 2022-07-30
Allow configuring tempdir


casey:
Sometimes TMPDIR is set to an invalid value, for example see #1269. To allow users to work around this, it should be possible to configure which tmpdir a justfile uses. A setting seems like the best approach:

```
set tmpdir := "foo"
```

It should be joined with the working directory, so it can be relative or absolute.

Smurphy000:
Hey @casey I'd be interested in implementing this if you have any pointers (familiarizing myself with the repo). Also I will not have time to start on this until next week if anybody else wants to pick this up before then.

casey:
Hi @Smurphy000, awesome! My suggestion would be to write the tests first. You can even open a draft PR with just the tests, and I can confirm that they look good. Then you can work backwards and implement the feature. First you'll get parse errors, so parse the setting, then you'll have to implement it. The actual change will be in `Recipe::run_shebang`, where the script temporary file is created.

casey:
Implemented in #1369!

PR #1369, opened 2022-10-17
allow tempdir configuration (#1300)

dmatos2012:
Hi, 

This PR addresses issue #1300 to allow configuring `tempdir` as an additional setting. 

I am fairly new at Rust, and wanted to tackle this issue as it seemed like a good challenge. Please let me know if something is just completely wrong, and I will gladly try to change it to get it right. But as far as I could see, it does what it intends on doing. Most of the changes are just to make the other tests happy again.

Looking forward to it! And thanks for this awesome tool!


casey:
Thanks for taking a crack at this! This should have tests. One suggestion is that the script could get the name of the script, extract the directory, write a file to that directory, and then after the tests assert that the file was created in the right location. The justfile for such a test will look something like this:

```
set tempdir = "."

foo:
  #!/usr/bin/env bash
  touch `dirname $0`/foo
```

dmatos2012:
I took your suggestion and I believe the two tests in `tests/misc.rs` should address what you have suggested. I was not sure exactly where to put the tests, so I can move them if you find them a more appropriate place, or maybe even its own `temp_dir.rs` test file.

Also not sure if these tests should have specific `linux` #[cfg] directives since I am not sure those will pass on windows(given use of linux only commands), but cant test that myself. 

Let me know your thoughts about the new two tests. Thanks again for your time !

src/analyzer.rs
```diff
@@ -68,6 +68,9 @@ impl<'src> Analyzer<'src> {
         Setting::WindowsShell(windows_shell) => {
           settings.windows_shell = Some(windows_shell);
         }
+        Setting::Tempdir(path) => {
+          settings.temp_dir = Some(path);
```
```suggestion
        Setting::Tempdir(tempdir) => {
          settings.tempdir = Some(tempdir);
```

src/parser.rs
```diff
@@ -786,13 +786,24 @@ impl<'tokens, 'src> Parser<'tokens, 'src> {
         value: Setting::WindowsShell(self.parse_shell()?),
         name,
       })
+    } else if name.lexeme() == Keyword::Tempdir.lexeme() {
+      Ok(Set {
+        value: Setting::Tempdir(self.parse_path()?),
```
casey: I think this can just be inlined, also the value of `Seting::Tempdir` can be a `String`:

```suggestion
        value: Setting::Tempdir(self.parse_string_literal()?.cooked),
```

src/parser.rs
```diff
@@ -786,13 +786,24 @@ impl<'tokens, 'src> Parser<'tokens, 'src> {
...
       }))
     }
   }
 
+  // Parse tempdir setting value
+  fn parse_path(&mut self) -> CompileResult<'src, StringLiteral<'src>> {
+    let path = self.parse_string_literal()?;
+    Ok(path)
+  }
+
```
```suggestion
```

src/recipe.rs
```diff
@@ -247,15 +247,24 @@ impl<'src, D> Recipe<'src, D> {
...
+        })?
+    } else {
+      tempfile::Builder::new()
+        .prefix("just")
+        .tempdir()
+        .map_err(|error| Error::TmpdirIo {
+          recipe: self.name(),
+          io_error: error,
+        })?
+    };
```
casey: Maybe something like this to deduplicate a little bit, and preserve the temporary directory prefix:

```suggestion
    let mut tempdir_builder = tempfile::Builder::new();
    
    tempdir_builder.prefix("just");

    let tempdir = match &context.settings.tempdir {
      Some(tempdir) => tempdir_builder.tempdir_in(tempdir),
      None => tempdir_builder.tempdir(),
    }.map_err(|error| Error::TmpdirIo {
      recipe: self.name(),
      io_error: error,
    })?;
```

src/setting.rs
```diff
@@ -8,6 +8,7 @@ pub(crate) enum Setting<'src> {
   Export(bool),
   PositionalArguments(bool),
   Shell(Shell<'src>),
+  Tempdir(StringLiteral<'src>),
```
casey:
```suggestion
  Tempdir(String),
```

src/setting.rs
```diff
@@ -22,6 +23,7 @@ impl<'src> Display for Setting<'src> {
       | Setting::PositionalArguments(value)
       | Setting::WindowsPowerShell(value) => write!(f, "{}", value),
       Setting::Shell(shell) | Setting::WindowsShell(shell) => write!(f, "{}", shell),
+      Setting::Tempdir(path) => write!(f, "{}", path),
```
casey:
```suggestion
      Setting::Tempdir(tempdir) => write!(f, "{}", tempdir),
```

src/settings.rs
```diff
@@ -13,6 +13,7 @@ pub(crate) struct Settings<'src> {
   pub(crate) ignore_comments: bool,
   pub(crate) positional_arguments: bool,
   pub(crate) shell: Option<Shell<'src>>,
+  pub(crate) temp_dir: Option<StringLiteral<'src>>,
```
casey:
```suggestion
  pub(crate) tempdir: Option<String>,
```

tests/json.rs
```diff
@@ -44,6 +44,7 @@ fn alias() {
         "export": false,
         "positional_arguments": false,
         "shell": null,
+        "temp_dir" : null,
```
casey:
```suggestion
        "tempdir" : null,
```

tests/misc.rs
```diff
@@ -1081,6 +1081,33 @@ test! {
...
+    standard_parent_tempdir := "/tmp"
+    foo:
+        #!/usr/bin/env bash
+        script_tempdir=$(basename "$(dirname $0)")
+        script_foo={{standard_parent_tempdir}}/$script_tempdir/foo
+        [[ -f $script_foo ]] && echo "foo_exists" || echo "foo_missing"
+    "#,
+    stdout: "foo_exists\n",
+}
+
```
casey: A few comments:

- There are still a bunch of macro-based tests, but I've been moving away from them. For new tests it's best to use the `Test` struct. Check out some of the other tests.
- These tests can go in their own file, perhaps `tests/tempdir.rs`.
- `realpath` isn't available on MacOS.
- The test can be simplified with something like this:

This will only succeed if the temporary directory is the same directory as the current directory, which will only be true if the tempdir has been set:

```
set tempdir := "."
foo:
  touch $(basename "$(dirname $0)")/bar
  cat bar
```

- I think the `standard_tempdir` test can be removed.

casey (changes requested):
Bunch of comments! Check 'em out. I think `tempdir` (no dash) is better than `temp_dir`.

casey:
Regarding running these tests on windows, we should avoid using any unix-specific commands so they can run anywhere.

dmatos2012:
Thanks for the feedback on all the files. I have changed them accordingly and added the tests which now should address what you have mentioned, and used the `Test` struct rather than the json. 

Thanks again for your time!

casey:
Nice merged! I added a small change, which is to interpret the tempdir path relative to the working directory, not the invocation directory.

dmatos2012:
Thanks so much for the guidance and adding the small tweaks. Hope to keep contributing in the near future to this repo :)

casey:
You are most welcome! Thank you for the PR!

PR #1373, opened 2022-10-21
add --only flag in order to skip recipe dependencies

### LINKED ISSUE #1314: Skip dependencies
Oftentimes when I'm writing a new recipe, I'd like to re-run that task, and _only_ that task. I don't want the dependencies to run because they might take a long time, and I know they're up-to-date.

Another way to say this is: I'd like a way to run the recipes specified explicitly given on the command line, without any dependencies. One example I can find of this feature is [Fake's `--single-target` parameter](https://fake.build/core-targets.html).

I'm aware of several workarounds for this, but it would be very nice to have this feature baked in. To list a few of the workarounds, so that I might illustrate why I would like this feature:

1. Write the dependencies with something like [entr](http://eradman.com/entrproject/) so that they don't unnecessarily re-run, as suggested in #424. This requires me to focus on having very robust dependencies before focusing on the task that I'm actually trying to work on. Sometimes I'd rather write the whole tree quickly and then go back and add such niceties.
2. Split each of my recipes into two. One which is my recipe, containing no dependencies. The second recipe includes the dependencies and includes the first recipe as the final dependency. This makes the justfile much less readable, though. Not great to maintain.
3. Comment out the dependencies as I'm developing the task. This works until I forget to uncomment them when I'm done.

sousajo-cc:
using this flag one can just re-run only that task specified in the cmd line.

Closes: https://github.com/casey/just/issues/1314

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -22,6 +22,7 @@ pub(crate) struct Config {
   pub(crate) list_heading: String,
   pub(crate) list_prefix: String,
   pub(crate) load_dotenv: bool,
+  pub(crate) only: bool,
   pub(crate) search_config: SearchConfig,
   pub(crate) shell: Option<String>,
   pub(crate) shell_args: Option<Vec<String>>,
@@ -91,6 +92,7 @@ mod arg {
   pub(crate) const JUSTFILE: &str = "JUSTFILE";
   pub(crate) const LIST_HEADING: &str = "LIST-HEADING";
   pub(crate) const LIST_PREFIX: &str = "LIST-PREFIX";
+  pub(crate) const ONLY: &str = "ONLY";
   pub(crate) const NO_DOTENV: &str = "NO-DOTENV";
   pub(crate) const NO_HIGHLIGHT: &str = "NO-HIGHLIGHT";
   pub(crate) const QUIET: &str = "QUIET";
@@ -175,6 +177,12 @@ impl Config {
           .value_name("TEXT")
           .takes_value(true),
       )
+        .arg(
+          Arg::with_name(arg::ONLY)
+              .long("only")
+              .help("Enable running only task")
+              .value_name("TASK")
+        )
       .arg(
         Arg::with_name(arg::NO_DOTENV)
           .long("no-dotenv")
@@ -592,6 +600,7 @@ impl Config {
       dotenv_filename: matches.value_of(arg::DOTENV_FILENAME).map(str::to_owned),
       dotenv_path: matches.value_of(arg::DOTENV_PATH).map(PathBuf::from),
       verbosity,
+      only: matches.is_present(arg::ONLY),
     })
   }
 
--- a/src/justfile.rs
+++ b/src/justfile.rs
@@ -316,44 +316,47 @@ impl<'src> Justfile<'src> {
     let mut evaluator =
       Evaluator::recipe_evaluator(context.config, dotenv, &scope, context.settings, search);
 
-    for Dependency { recipe, arguments } in recipe.dependencies.iter().take(recipe.priors) {
-      let arguments = arguments
-        .iter()
-        .map(|argument| evaluator.evaluate_expression(argument))
-        .collect::<RunResult<Vec<String>>>()?;
+    if !context.config.only {
+      for Dependency { recipe, arguments } in recipe.dependencies.iter().take(recipe.priors) {
+        let arguments = arguments
+          .iter()
+          .map(|argument| evaluator.evaluate_expression(argument))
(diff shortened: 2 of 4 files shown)


casey:
Nice! I'm curious, is this something that you personally needed? If so, what is the use-case?

I think that it would be better if this wasn't implemented as a option that takes an argument `just --only RECIPE`, where the `RECIPE` is an argument to the `--only` option. Instead, it should be a flag that doesn't take any arguments, but which, when used, causes `just` to skip running any dependencies that aren't explicitly mentioned on the command line. This would allow running it with multiple recipes, like `just --only foo bar`, as well as recipes that take arguments.

sousajo-cc:
> Nice! I'm curious, is this something that you personally needed? If so, what is the use-case?
>
Hello :) Yeah I have kinda of complicated git workflow at work, and sometimes I just want to redo a specific recipe not the whole thing. 

> I think that it would be better if this wasn't implemented as a option that takes an argument `just --only RECIPE`, where the `RECIPE` is an argument to the `--only` option. Instead, it should be a flag that doesn't take any arguments, but which, when used, causes `just` to skip running any dependencies that aren't explicitly mentioned on the command line. This would allow running it with multiple recipes, like `just --only foo bar`, as well as recipes that take arguments.
>

That makes a lot of sense, now that you mention it, I will try to play with it a bit!

sousajo-cc:
if that's ok @casey I will try to draft a couple of tests for it, and if you can get me a feedback on those would be great, so I'm sure that they can guide the implementation ^^

casey:
> Hello :) Yeah I have kinda of complicated git workflow at work, and sometimes I just want to redo a specific recipe not the whole thing.

Okay good, I always like to make sure that features are actually going to be used!

> That makes a lot of sense, now that you mention it, I will try to play with it a bit!

Sounds good, give it a shot, and feel free to ask for help.

> if that's ok @casey I will try to draft a couple of tests for it, and if you can get me a feedback on those would be great, so I'm sure that they can guide the implementation ^^

That's a great idea, I think it's always good to write some tests first so you can get feedback and figure out exactly how the feature behaves, even if they fail.

alexpovel:
Hi! I'm a potential user as well, I'd love this feature. In some sense, it's very antithetical to the idea of having a dependency graph in the first place. But as there are now _multiple_ (2!) people publicly voicing interest for this feature, perhaps we can let pragmatism overrule purity :^) 

That said, let me contribute a use case! In the .NET world, a (CI) workflow like [this one](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-net#using-the-net-starter-workflow) is customary:

```yaml
name: dotnet package

on: [push]

jobs:
  build:

    runs-on: ubuntu-latest
    strategy:
      matrix:
        dotnet-version: ['3.0', '3.1.x', '5.0.x' ]

    steps:
      - uses: actions/checkout@v3
      - name: Setup .NET Core SDK ${{ matrix.dotnet-version }}
        uses: actions/setup-dotnet@v3
        with:
          dotnet-version: ${{ matrix.dotnet-version }}
      - name: Install dependencies
        run: dotnet restore
      - name: Build
        run: dotnet build --configuration Release --no-restore
      - name: Test
        run: dotnet test --no-restore --verbosity normal
```

However, a corresponding `Justfile` would have the dependencies as: `test: build` and `build: restore`. To test, we need to build, and to build, we need to restore any packages we depend on (`dotnet test` works by itself as well and will build and restore by itself, but this is `just`, so we're looking into making these steps explicit and separate!). `just test` then works beautifully locally, but in CI, it's beneficial to separate jobs into individual steps (or whatever the CI systems' lingo is). That allows us to debug quicker. A `just test` job might fail for a host of reasons, _not only_ due to failing tests. Package restoration might have failed. This wouldn't have good UI/UX visibility if `just test` were the only step in a job.

Similar ideas apply to [Python pipelines as well, for example](https://github.com/alexpovel/ancv/blob/8dc46cb83252eecc4031d1a7ab438a9548ed6f60/.github/workflows/publish.yml). A `test` requires installing the environment, but that in turn requires installing the tooling to set up our environment in the first place: `test: setup-environment`, and `setup-environment: install-poetry`, where `poetry` is an environment management tool. All the same ideas as described above apply.

Without a feature like `--only`, all this is hard to express, yet very desirable in CI. Maintaining separate, CI-specific jobs, which would each not specify any dependencies, is tedious.

I'm just getting into Rust myself, let me know if you're still working on this @sousajo-cc , else I can offer to chime in!

casey:
This has gotten a little stale, feel free to update and re-open!

PR #1375, opened 2022-10-24
Add function for capitalizing strings

femnad:
Hello,

This PR adds a `capitalize` function (name inspired by the Python [string method](https://docs.python.org/3/library/stdtypes.html?highlight=capitalize#str.capitalize)) for converting a string to sentence case; first letter upper case, rest lower case.

I have a use case where a function like this would be useful and it's a small enough change so I'm creating this PR without a discussion issue but it's not a big loss if it doesn't get merged.

Thanks!

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -16,6 +16,7 @@ lazy_static! {
   pub(crate) static ref TABLE: BTreeMap<&'static str, Function> = vec![
     ("absolute_path", Unary(absolute_path)),
     ("arch", Nullary(arch)),
+    ("capitalize", Unary(capitalize)),
     ("clean", Unary(clean)),
     ("env_var", Unary(env_var)),
     ("env_var_or_default", Binary(env_var_or_default)),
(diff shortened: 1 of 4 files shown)


src/function.rs
```diff
@@ -79,6 +80,15 @@ fn arch(_context: &FunctionContext) -> Result<String, String> {
   Ok(target::arch().to_owned())
 }
 
+fn capitalize(_context: &FunctionContext, s: &str) -> Result<String, String> {
+  let mut chars = s.chars();
+  match chars.next() {
+      None => Ok(String::new()),
+      Some(c) => Ok(c.to_uppercase().collect::<String>() +
+                    &chars.as_str().to_lowercase()),
+  }
```
casey: Maybe something like this:

```suggestion
  Ok(s.chars().enumerate().flat_map(|(i, c)| if i == 0 { c.upper_case() } else { c.lower_case() }).collect())
```

casey:
Good feature, and thanks for also adding a test! Very minor comment suggesting a slightly tweaked implementation.

casey:
Also, could you add a note, like this to the function documentation: `<sup>master</sup>`. This is to indicate that the function is on master and not in a released version of just yet. I'll change it to the version it winds up being released in.

src/function.rs
femnad: So much more elegant! Applied with rustfmt and some modifications.

femnad:
Added master designation to the function documentation:

image

casey:
Merged! Thank you for the PR!

PR #1387, opened 2022-10-28
OS Recipe Attributes

casey:
This adds `[linux]`, `[macos]`, `[windows]`, and `[unix]` recipe OS attributes.

By default, without an OS attribute, a recipe is enabled on all operating systems. With an OS attribute, a recipe is only enabled when being run by a version of just build for that OS. Note that this depends on the OS the `just` binary was built for, not the OS that it's running on. So recipes with `[linux]` in a justfile run by a`just` binary built for Linux but running on the Windows Subsystem for Linux be enabled.

This allows writing cross-platform justfiles by writing multiple recipes with the same name which are enabled on different OSes. For example:

```make
[windows]
foo:
  echo bar

[unix]
foo:
  echo baz
```

On Windows, `just foo` will print `bar` and on linux `just foo` will print `baz`.

Recipes with the `[unix]` attribute are enabled on all unices, including Linux and MacOS.

Todo:

- [ ] Document in readme

### DIFF
--- a/src/attribute.rs
+++ b/src/attribute.rs
@@ -1,7 +1,10 @@
 use super::*;
 
-#[derive(EnumString, PartialEq, Debug, Clone, Serialize)]
-#[strum(serialize_all = "kebab_case")]
+#[derive(
+  EnumString, PartialEq, Debug, Copy, Clone, Serialize, Ord, PartialOrd, Eq, IntoStaticStr,
+)]
+#[strum(serialize_all = "kebab-case")]
+#[serde(rename_all = "kebab-case")]
 pub(crate) enum Attribute {
   Linux,
   Macos,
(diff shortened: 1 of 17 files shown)


src/attribute.rs
```diff
@@ -1,13 +1,34 @@
...
+  }
+}
+
+#[cfg(test)]
+mod tests {
+  use super::*;
+
+  #[test]
+  fn to_str() {
+    assert_eq!(Attribute::NoExitMessage.to_str(), "no-exit-message");
```
esiegel, line 32: Maybe add the same test for all of the enums, though not sure if it tests much.

esiegel (approved):
Super simple.  Looks good to me.

PR #1390, opened 2022-10-28
Note on how shebang recipes are executed

bkhl:
To avoid confusion due to shebang recipes not working quite like executable scripts on e.g. Linux, added a note clarifying how it is called.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1697,6 +1697,8 @@ Yo from a shell script!
 Hello from ruby!
 ```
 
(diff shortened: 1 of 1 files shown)


casey:
I don't think that shebang scripts on linux are executed by passing the script contents to standard in, although I could be wrong. Do you have a source for that?

casey:
Just checked, and shebang scripts are run by passing the path to the script to the interpreter command, so `just` works the same as normal shebangs.

PR #1393, opened 2022-10-29
Add a replace function that supports regex

### LINKED ISSUE #1389: Add a replace function that supports regex
The current `replace(s, from, to)` function doesn't seem to support regex. I think it would be helpful to also have a `replace_regex(s, from, to)` for more complex replacements.

miles170:
Add `replace_regex(s, from, to)` that supports replacing text using regular expressions.

Closes #1389.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1145,6 +1145,7 @@ The executable is at: /bin/just
 
 - `quote(s)` - Replace all single quotes with `'\''` and prepend and append single quotes to `s`. This is sufficient to escape special characters for many shells, including most Bourne shell descendants.
 - `replace(s, from, to)` - Replace all occurrences of `from` in `s` to `to`.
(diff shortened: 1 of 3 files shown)


ianpittwood:
LGTM! Thanks for taking this on!

casey:
Thanks for this, and sorry I've been so slow to review it! Can you add a test for the error case? I want to make sure the error message looks good. If the regex crate's error message doesn't show the original regex, we should include that as well.

miles170:
There is a typo in the README. (regec -> regex)

casey:
Good catch! Fixing here: #1409.

Tweaked the argument names and added an entry in the readme. I also removed the error prefix, I think the underlying error is probably descriptive enough. Thank you for the PR!

PR #1394, opened 2022-10-30
On windows powershell is named powershell.exe

asdf8dfafjk:
At least on Windows 10 21H2


casey:
Thanks! `pwsh.exe` is the newer PowerShell Core, but it isn't installed by default, so using `powershell.exe` in the docs is probably a safer bet.

Issue #1405, opened 2022-11-04
`just --list` does not evaluate default parameter expressions


epbuennig:
I have a recipe for a project where I usually pass a some args and file to a recipe, and if I don't pass anything, I use `xplr` to choose a file.

It works as expected (using `just 1.8.0`) with this:
```
FILES := "path_to_files"

# the echo inside the file expression is where I had xplr
recipe +args=('$(xplr ' + FILES + ')'):
    echo {{args}}
```

However, I expected `just --list` to substitute the variable, but it doesn't.
Current behavior, no substitution nor evaluation of the expression:
```
$ just --list
Available recipes:
    recipe +args=('$(xplr ' + FILES + ')'):
```
Expected behavior, variable is substituted and expression is shown evaluated:
```
$ just --list
Available recipes:
    recipe +args='$(xplr path_to_files)':
```

I think expressions as parameter defaults should be evaluated when showing the recipes as it shows a better picture of what the default argument means.

EDIT:
On a similar note I'd also expect `just --show` to evaluate expressions, which it currently doesn't. Maybe it `--evaluate` could be used in conjunction with `--list` and `--show` to evaluate all expressions?

casey:
Thank you for the issue! I definitely agree that it would be good from a documentation perspective. However, we would have to evaluate things like backticks and assignments, which might not be desirable when not executing recipes. I'm open to feedback on this but overall I'm leaning towards not doing this.

epbuennig:
I think it would be fine if it is opt in? Like I suggested `--evaluate --list`/`--evaluate --show`, which currently can't be used together, could be used here. Do you have any concerns with that approach?

casey:
Oh, sorry, I missed that! Yes, definitely, I think that's a great approach.

epbuennig:
I'd love to tackle that, seems fairly straight forward and like a good first issue, I'd tackle the problem this weekend.

epbuennig:
### I've done some reading and testing and have a few questions:

How much should we actually evaluate?
My initial guess was, as much as we can, but that seems like a bad idea.

Consider this example:
```
key := 'value'
recipe PARAM=key:
    echo {{ if PARAM == key { 'default: ' } else { 'arg: ' } }} {{ PARAM }}
```
How far can we evaluate `just --evaluate --show recipe`, I'd expect this as a user:
```
recipe PARAM='value':
    echo {{ if PARAM == 'value' { 'default: ' } else { 'arg: ' } }} {{ PARAM }}
```
[...]

PR #1411, opened 2022-11-12
Allow passing `--evaluate` alongside `--list` to evaluate parameter default expressions

epbuennig:
This is an initial draft for adding support to pass `--evaluate` as an option to `--list`  to evaluate expressions in parameter defaults. This PR addresses #1405.

A `justfile` like so:
```
key := "value"

recipe arg=key:
    echo {{ arg }}
```
Should then produce the following output:
```
$ just --evaluate --list
Available recipes:
    recipe arg='key'
```
Possible extensions, like evaluation for `--show` or `--dump` have been left for another PR (as discussed [here](https://github.com/casey/just/issues/1405#issuecomment-1312552643)).

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -159,7 +159,7 @@ impl Config {
       .arg(
         Arg::with_name(arg::EVAL)
           .long("eval")
-          .help("Evaluate parameter default expressions.")
+          .help("Evaluate parameter defaults.")
           .requires(cmd::LIST)
       )
       .arg(
--- a/src/config.rs
+++ b/src/config.rs
@@ -39,6 +39,7 @@ mod cmd {
   pub(crate) const COMPLETIONS: &str = "COMPLETIONS";
   pub(crate) const DUMP: &str = "DUMP";
   pub(crate) const EDIT: &str = "EDIT";
+  pub(crate) const EVALUATE: &str = "EVALUATE";
   pub(crate) const FORMAT: &str = "FORMAT";
   pub(crate) const INIT: &str = "INIT";
   pub(crate) const LIST: &str = "LIST";
@@ -53,6 +54,7 @@ mod cmd {
     COMPLETIONS,
     DUMP,
     EDIT,
+    EVALUATE,
     FORMAT,
     INIT,
     LIST,
@@ -85,6 +87,7 @@ mod arg {
   pub(crate) const DOTENV_PATH: &str = "DOTENV-PATH";
   pub(crate) const DRY_RUN: &str = "DRY-RUN";
   pub(crate) const DUMP_FORMAT: &str = "DUMP-FORMAT";
+  pub(crate) const EVAL: &str = "EVAL";
   pub(crate) const HIGHLIGHT: &str = "HIGHLIGHT";
   pub(crate) const JUSTFILE: &str = "JUSTFILE";
   pub(crate) const LIST_HEADING: &str = "LIST-HEADING";
@@ -109,26 +112,6 @@ mod arg {
   pub(crate) const DUMP_FORMAT_JSON: &str = "json";
   pub(crate) const DUMP_FORMAT_JUST: &str = "just";
   pub(crate) const DUMP_FORMAT_VALUES: &[&str] = &[DUMP_FORMAT_JUST, DUMP_FORMAT_JSON];
-}
-
-mod arg_or_cmd {
-  use super::*;
-
-  pub(crate) const EVALUATE: &str = "EVALUATE";
-
-  pub(crate) const CONFLICT_EVALUATE: &[&str] = &[
-    cmd::CHANGELOG,
-    cmd::CHOOSE,
-    cmd::COMMAND,
-    cmd::COMPLETIONS,
-    cmd::DUMP,
-    cmd::EDIT,
-    cmd::FORMAT,
-    cmd::INIT,
-    cmd::SHOW,
-    cmd::SUMMARY,
-    cmd::VARIABLES,
-  ];
 }
 
 impl Config {
@@ -172,6 +155,12 @@ impl Config {
           .default_value(arg::DUMP_FORMAT_JUST)
           .value_name("FORMAT")
           .help("Dump justfile as <FORMAT>"),
+      )
+      .arg(
+        Arg::with_name(arg::EVAL)
+          .long("eval")
+          .help("Evaluate parameter default expressions.")
+          .requires(cmd::LIST)
       )
       .arg(
         Arg::with_name(arg::HIGHLIGHT)
@@ -281,10 +270,6 @@ impl Config {
           .help("Use <WORKING-DIRECTORY> as working directory. --justfile must also be set")
           .requires(arg::JUSTFILE),
       )
-      .arg(Arg::with_name(arg_or_cmd::EVALUATE).long("evaluate").help(
-        "Evaluate and print all variables, or evaluate all variables for another command. If a \
-         variable name is given as an argument, only print that variable's value.",
-      ).conflicts_with_all(arg_or_cmd::CONFLICT_EVALUATE))
       .arg(
         Arg::with_name(cmd::CHANGELOG)
           .long("changelog")
@@ -322,6 +307,10 @@ impl Config {
           .long("edit")
           .help("Edit justfile with editor given by $VISUAL or $EDITOR, falling back to `vim`"),
       )
+      .arg(Arg::with_name(cmd::EVALUATE).long("evaluate").help(
+        "Evaluate and print all variables. If a variable name is given as an argument, only print \
+        that variable's value.",
+      ))
       .arg(
         Arg::with_name(cmd::FORMAT)
           .long("fmt")
@@ -541,7 +530,7 @@ impl Config {
       Subcommand::Init
     } else if matches.is_present(cmd::LIST) {
       Subcommand::List {
-        evaluate: matches.is_present(arg_or_cmd::EVALUATE),
+        evaluate: matches.is_present(arg::EVAL),
       }
     } else if let Some(name) = matches.value_of(cmd::SHOW) {
       Subcommand::Show {
@@ -549,10 +538,10 @@ impl Config {
       }
     } else if matches.is_present(cmd::VARIABLES) {
       Subcommand::Variables
-    } else if matches.is_present(arg_or_cmd::EVALUATE) {
+    } else if matches.is_present(cmd::EVALUATE) {
       if positional.arguments.len() > 1 {
         return Err(ConfigError::SubcommandArguments {
-          subcommand: arg_or_cmd::EVALUATE,
+          subcommand: cmd::EVALUATE,
           arguments: positional
             .arguments
             .into_iter()
@@ -1013,7 +1002,7 @@ mod tests {
 
   error! {
     name: subcommand_conflict_evaluate,
-    args: ["--summary", "--evaluate"],
+    args: ["--list", "--evaluate"],
   }
 
   error! {
@@ -1115,8 +1104,8 @@ mod tests {
   }
 
   test! {
-    name: subcommand_list_evaluate,
-    args: ["--list", "--evaluate"],
+    name: subcommand_list_eval,
+    args: ["--list", "--eval"],
     subcommand: Subcommand::List {
       evaluate: true,
     },
--- a/src/config.rs
+++ b/src/config.rs
@@ -992,11 +992,6 @@ mod tests {
   }
 
   error! {
-    name: subcommand_conflict_evaluate,
-    args: ["--list", "--evaluate"],
-  }
-
-  error! {
     name: subcommand_conflict_show,
     args: ["--list", "--show"],
   }
--- a/src/config.rs
+++ b/src/config.rs
@@ -125,6 +125,7 @@ mod arg_or_cmd {
     cmd::EDIT,
     cmd::FORMAT,
     cmd::INIT,
+    cmd::SHOW,
     cmd::SUMMARY,
     cmd::VARIABLES,
   ];
@@ -545,7 +546,6 @@ impl Config {
     } else if let Some(name) = matches.value_of(cmd::SHOW) {
       Subcommand::Show {
         name: name.to_owned(),
-        evaluate: matches.is_present(arg_or_cmd::EVALUATE),
       }
     } else if matches.is_present(cmd::VARIABLES) {
       Subcommand::Variables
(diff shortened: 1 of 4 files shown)


epbuennig:
This currently works as expected on the repo's `justfile`.

I'm not entirely sure how we should approach passing a mutable evaluator reference to the formatting of a parameter (or later to anything, really). With the current API of `DisplayColor` we'd have to use interior mutability as `FmtEval` in `src/parameter.rs` or some other way. This is obviously hacked in to illustrate the problem.

There's also the issue of `--evalaute` not showing up in the usage anymore, which is why the single commented out test in `test/command.rs` is failing, not sure how to approach that.

src/config.rs
```diff
@@ -113,6 +111,25 @@ mod arg {
   pub(crate) const DUMP_FORMAT_VALUES: &[&str] = &[DUMP_FORMAT_JUST, DUMP_FORMAT_JSON];
 }
 
+mod arg_or_cmd {
```
epbuennig: Not too keen on the naming here, but I figured this should be different to `arg` as it serves 2 purposes now.

src/parameter.rs
```diff
@@ -13,6 +13,43 @@ pub(crate) struct Parameter<'src> {
...
+    }
+    if let Some(prefix) = self.0.kind.prefix() {
+      write!(f, "{}", color.annotation().paint(prefix))?;
+    }
+    write!(f, "{}", color.parameter().paint(self.0.name.lexeme()))?;
+    if let Some(ref default) = self.0.default {
+      write!(
+        f,
+        "={}",
+        color.string().paint(&format!(
```
epbuennig: The extra `format!` is mainly because regular strings are green, including their quotes, and I didn't look too far into the coloring module to see if this could be done better.

src/parameter.rs
```diff
@@ -13,6 +13,43 @@ pub(crate) struct Parameter<'src> {
...
+    evaluator: &'eval std::cell::RefCell<Evaluator<'src, 'run>>,
+  ) -> FmtEval<'this, 'eval, 'src, 'run> {
+    FmtEval(self, evaluator)
+  }
+}
+
+pub struct FmtEval<'param, 'eval, 'src, 'run>(
+  &'param Parameter<'src>,
+  &'eval std::cell::RefCell<Evaluator<'src, 'run>>,
+);
```
epbuennig: I think even if the evaluator was included in `Wrapper`, it would probably cause major problems from the lifetimes alone.
I tried to use `... -> impl ColorDisplay + 'this`, even with explicit bounds that all other lifetimes outlive `'this`, it would complain about some lifetimes missing from the signature.

Not only that, but I imagine this would complicate the current approach of `Wrapper` which uses a trait object and probably faces similar issues.

src/parameter.rs
epbuennig: Unfortunately, I selected the wrong range here and GitHub won't let me correct it, referring to L43-L47.

src/subcommand.rs
```diff
@@ -458,9 +494,19 @@ impl Subcommand {
...
-          line_width += UnicodeWidthStr::width(
-            format!(" {}", parameter.color_display(Color::never())).as_str(),
-          );
+          line_width += if let Some(evaluator) = &mut evaluator {
+            UnicodeWidthStr::width(
+              format!(
+                " {}",
+                parameter
+                  .format_eval(evaluator)
+                  .color_display(Color::never())
```
epbuennig: It should also be noted that we may want to keep the evaluated result here to avoid another evaluation later.

src/config.rs
casey: On reflection and looking at the code, I think we should avoid overloading the `--evaluate` command flag with a second meaning. It complicates the argument parser and makes the help text weird. How about `--list --eval`?

src/parameter.rs
```diff
@@ -13,6 +13,43 @@ pub(crate) struct Parameter<'src> {
   pub(crate) name: Name<'src>,
 }
 
+impl<'src> Parameter<'src> {
+  pub(crate) fn format_eval<'this, 'eval, 'run>(
```
casey: Can we do something like this instead?

```rust
use super::*;

/// A single function parameter
#[derive(PartialEq, Debug, Clone, Serialize)]
pub(crate) struct Parameter<'src, T = Expression<'src>> {
  /// An optional default expression
  pub(crate) default: Option<T>,
  /// Export parameter as environment variable
  pub(crate) export: bool,
  /// The kind of parameter
  pub(crate) kind: ParameterKind,
  /// The parameter name
  pub(crate) name: Name<'src>,
}

impl<'src, T: Display> ColorDisplay for Parameter<'src, T> {
  fn fmt(&self, f: &mut Formatter, color: Color) -> Result<(), fmt::Error> {
    if self.export {
      write!(f, "$")?;
    }
    if let Some(prefix) = self.kind.prefix() {
      write!(f, "{}", color.annotation().paint(prefix))?;
    }
    write!(f, "{}", color.parameter().paint(self.name.lexeme()))?;
    if let Some(ref default) = self.default {
      write!(f, "={}", color.string().paint(&default.to_string()))?;
    }
    Ok(())
  }
}
```

This gives Parameter a generic parameter T which is the type of the default. T defaults to `Expression<'src>`, so everything works as normal. You can create an evaluated parameter with `Parameter<String>`.

So when `--eval` is passed, convert parameters to `Parameter<String>` using eval before printing.

src/parameter.rs
```diff
@@ -13,6 +13,43 @@ pub(crate) struct Parameter<'src> {
...
+    FmtEval(self, evaluator)
+  }
+}
+
+pub struct FmtEval<'param, 'eval, 'src, 'run>(
+  &'param Parameter<'src>,
+  &'eval std::cell::RefCell<Evaluator<'src, 'run>>,
+);
+
+impl<'this, 'eval, 'src, 'run> ColorDisplay for FmtEval<'this, 'eval, 'src, 'run> {
```
casey: I think that combining fmt and eval is probably the wrong way to go. See my comment above.

casey (approved):
Thanks for tackling this! Added a comment with an alternative approach that might be much cleaner. Let me know what you think.

casey (changes requested):
Whoops, meant to request changes.

src/parameter.rs
epbuennig: Yeah, that's what I thought too.

src/parameter.rs
epbuennig: If `Parameter` becomes generic then when we evaluate it we need to store it separately to the originals because the recipes expect `Parameter<Expression<'src>>`, right? That would add some complexity to simply printing the recipes.
What if we instead evaluate them in place into `Expression::String`? That way the type stays the same, and we can keep most of the formatting code that was already in place, I think.

EDIT:
Ah, I realize now that the only suitable expression type is `StringLiteral` which borrows for `'src`, I guess that won't work. But I'd still like to avoid having to pack up the parameters separately to the recipes when evaluating and checking the length. Would you mind if I instead added an `enum` containing either an Expression or an evaluated expression? To avoid the type of `Parameter` being different after evaluation. Although that would require mutable access.

epbuennig: I agree on the flag, I want to avoid changing something like the CLI in a breaking manner. Although, it feels a bit weird to have both `evaluate` and `eval`, do you think we can differentiate them more?

epbuennig:
I'll get to it in the next few days and should have it ready and cleaned up for another review.

epbuennig:
At the risk of this becoming stale, I've been stuck on a peculiar borrowing problem, I'll push the code as is, it does not currently compile.

src/subcommand.rs
```diff
@@ -493,19 +485,12 @@ impl Subcommand {
...
-                parameter
-                  .format_eval(evaluator)
-                  .color_display(Color::never())
-              )
-              .as_str(),
-            )
-          } else {
-            UnicodeWidthStr::width(format!(" {}", parameter.color_display(Color::never())).as_str())
+          if let Some(evaluator) = &mut evaluator {
+            let parameter = parameter.evaluate_default(evaluator)?;
```
epbuennig: According to the borrow checker, we're borrowing from `&justfile.recipes` in line 479 when we return in the `Err` case here. Despite the only lifetime being mentioned in the return value of `evalaute_default` being `'src`. I'm not sure why this is, as it works fine in `justfile.run()`.

epbuennig: Changing the `justfile` parameter of `list` to not be moved but borrowed, yields a similar result, requiring the borrow of `&Justfile<'scr>` itself to outlive `'src`, i.e., suggesting `&'src Justfile<'src>`, which is not at all satiable.

epbuennig:
So I followed your approach first to see where it leads me, other than the borrowing problem I'm stumped by, there are a few other minor problems I want to highlight:

If we evaluate in the nested loop as seen in the current state of this draft, we'll evaluate a parameter once for every alias of a command. Later in the actual formatting step we need to either do this again, doubling all evaluations, or tracking which recipe they belong to, requiring some local of type `BTreeMap<&'src str, Vec<Parameter<'src, String>>` or similar.
I personally would rather mutate the parameters in place, caching the evaluated `String` alongside its original `Expression`. Allowing us to evaluate once and once only, but the `Recipe`s are behind `Rc`s because of the aliases.

src/subcommand.rs
```diff
@@ -485,12 +487,37 @@ impl Subcommand {
...
-          line_width += UnicodeWidthStr::width(
-            format!(" {}", parameter.color_display(Color::never())).as_str(),
-          );
+          // FIXME: unwrap -> ?
+          if let Some(evaluated) = evaluator
+            .as_mut()
+            .map(|evaluator| parameter.evaluate_default(evaluator))
+            .transpose()
+            .unwrap()
+            .flatten()
```
epbuennig: The `unwrap` on L495 is a result of what I previously wrote, I have no idea why it thinks that we're borrowing from the locally owned `justfile` when returning an eval error.
So I implemented the functionality anyway by unwrapping and left the `FIXME` in there.

src/subcommand.rs
```diff
@@ -485,12 +487,37 @@ impl Subcommand {
...
+          {
+            use std::collections::btree_map::Entry;
+
+            line_width += UnicodeWidthStr::width(
+              format!(" {}", evaluated.color_display(Color::never())).as_str(),
+            );
```
epbuennig: We could technically skip this string alloc by simply counting the evaluated and name length + 3, but optimizations are not the point of this PR. I want to keep the diff minimal.

src/subcommand.rs
```diff
@@ -485,12 +487,37 @@ impl Subcommand {
...
+                let mut parameters = BTreeMap::new();
+                parameters.insert(evaluated.name.lexeme(), evaluated);
+                vacant.insert(parameters);
+              }
+              Entry::Occupied(mut occopied) => {
+                occopied
+                  .get_mut()
+                  .insert(evaluated.name.lexeme(), evaluated);
+              }
+            }
```
epbuennig: I've tested this locally with a justfile like this:
```make
key1 := `echo val11`
key2 := `echo val22`

recipe1 arg1=key1 arg2=`echo val12`:
    echo

recipe2 arg1=`echo val21` arg2=key2:
    echo
```
But I wasn't sure how to exactly write a test here to fit in with the rest, if that is even necessary.

src/config.rs
```diff
@@ -522,11 +529,15 @@ impl Config {
...
     } else if matches.is_present(cmd::LIST) {
-      Subcommand::List
+      Subcommand::List {
+        evaluate: matches.is_present(arg::EVAL),
+      }
     } else if let Some(name) = matches.value_of(cmd::SHOW) {
       Subcommand::Show {
         name: name.to_owned(),
       }
+    } else if matches.is_present(cmd::VARIABLES) {
```
casey: This should be moved back to the original location, to avoid churn in the diff.

src/config.rs
```diff
@@ -155,6 +156,12 @@ impl Config {
           .value_name("FORMAT")
           .help("Dump justfile as <FORMAT>"),
       )
+      .arg(
+        Arg::with_name(arg::EVAL)
+          .long("eval")
+          .help("Evaluate parameter default expressions.")
```
```suggestion
          .help("Evaluate parameter defaults.")
```

src/subcommand.rs
```diff
@@ -428,7 +432,35 @@ impl Subcommand {
...
   }
 
-  fn list(config: &Config, justfile: Justfile) {
+  fn list<'src>(
+    config: &Config,
+    evaluate: bool,
+    search: &Search,
+    justfile: Justfile<'src>,
+  ) -> Result<(), Error<'src>> {
+    let dotenv = if config.load_dotenv {
```
casey: A dotenv is only needed if we're evaluating parameter defaults, so we shouldn't load it if `evaluate` is false.

src/subcommand.rs
casey: I just pushed a diff that avoids this. `justfile.recipes.iter()` works for some reason.

src/subcommand.rs
casey: What is this testing?

casey (changes requested):
A few minor comments.

I'd like to avoid entangling evaluation with formatting. Check out this PR for a sketch of how to do that:

https://github.com/casey/just/commit/1700cd7d42843994c2676233d77075b606baf4fd

It splits out the formatting into `list_inner`, which should probably be called` list_evaluated`. This should make it easier to do evaluation in one place, and then list_evaluated can be the same regardless of whether it's getting evaluated or unevaluated parameters.

epbuennig:
> What is this testing?

This was simply bad code and is fixed in ff1b69f.

src/config.rs
epbuennig: Resolved in 17112ff!

src/config.rs
epbuennig: Resolved in d3dfffa!

src/subcommand.rs
epbuennig: Resolved in 154bf63!

epbuennig:
I think I've resolved the remaining comments, it currently tests fine on my machine, passing `--eval` evaluates as expected. I'll take a look at whatever obvious improvements to the code can be done and looking for unhandled edge cases before requesting review.

epbuennig:
It's been a while, but I didn't see anything while checking the places I edited. I'm going to request the final review, thank you for being so patient with me. The remaining conflict is simply an inlined format string.

PR #1416, opened 2022-11-19
Use more secure cURL options in install.sh

casey:


God-damnit-all:
Why TLS 1.2 and not 1.3?

casey:
@ImportTaste I didn't know that 1.2 was no longer the latest, opened #1481 to use 1.3.

God-damnit-all:
@casey I guess I should ask again since the commit was reverted, why TLS 1.2 and not 1.3?

PR #1417, opened 2022-11-19
Do not include shebang lines on Windows

casey:


src/shebang.rs
```diff
@@ -47,7 +47,7 @@ impl<'line> Shebang<'line> {
   }
 
   pub(crate) fn include_shebang_line(&self) -> bool {
-    !matches!(self.interpreter_filename(), "cmd" | "cmd.exe")
+    !cfg!(windows) && !matches!(self.interpreter_filename(), "cmd" | "cmd.exe")
```
pfmoore, line 50: Is this correct? It seems to be saying that the build is for Windows *and* the interpreter filename is `cmd`. Shouldn't that "and" be an "or"?

I wasn't able to check this as I'm not entirely sure what `justfile` syntax would set `interpreter_filename`.

PR #1418, opened 2022-11-19
Make include_shebang_line clearer

casey:
@pfmoore Check out this version, it should be clearer. We don't want to include the shebang line if we're on windows or if the interpreter is cmd. Strictly speaking, I don't think the second condition is required, since it's unlikely that anyone will use `cmd` anywhere but windows, but I wanted to change as little as possible.

### DIFF
--- a/src/shebang.rs
+++ b/src/shebang.rs
@@ -47,7 +47,7 @@ impl<'line> Shebang<'line> {
   }
 
   pub(crate) fn include_shebang_line(&self) -> bool {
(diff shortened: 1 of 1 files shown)


pfmoore:
Yep, that works fine. A shebang of `#!py` now (reliably) runs the script under Python on Windows[^1].

[^1]: Well, ignoring bug https://github.com/python/cpython/issues/99442 in Python 3.11, but that's their problem, not yours 🙂 And it will be fixed when 3.11.1 comes out.

casey:
Nice!

Issue #1407, opened 2022-11-08
Add quick-and-dirty justfile inclusion


casey:
Users have wanted to be able to include other justfiles for a long time, and I've wanted to support this. However, I've been worried about implementing something hacky that would bring problems and be hard to support and debug. I've had a design in my head for a module system, but it's complicated enough that I'm unlikely to implement it any time soon.

As a compromise, I think that we should add a simple `include` statement, similar to @MadBomber's [justprep](https://github.com/MadBomber/justprep).

I'm unlikely to implement this myself, because I have other projects which are taking all my time, but if someone wants to pick this up, I can review and help them get it over the line.

The design is this:

- Add include statements of the form `include "PATH_TO_JUSTFILE"`
- When the parser hits an include statement, it passes the `PATH_TO_JUSTFILE` to the loader, which returns the text with `'src` lifetime
- The parser then invokes the lexer on the returned text, and inserts those tokens into its own token stream, replacing the `include` statement with those tokens
- Parsing then continues as normal
- Users can use the `allow-duplicate-recipes` and OS attributes to override recipes in `include`s and select recipes for the current OS

The idea behind using the word `include` instead of `use` or `import` is that we want to express that we're doing quick-and-dirty inclusion of the tokens. Later, if someone is feeling super motivated, or I get around to it, we can implement a module system, and users can migrate to the module system if they want to.

Additions and improvements that shouldn't block the initial PR:

- Show correct filename in error messages
- Add attributes to include statements, so that which file is included can depend on the OS.
- Allow duplicate settings, exports, and assignments, so that you can override anything

casey:
@MadBomber what do you think of this design?

casey:
Later on we can add some extensions:

- Add attributes to `include` statements, so that which file is included can depend on the OS.
- Allow duplicate settings, exports, and assignments, so that you can override anything

MadBomber:
@casey The advantage is that you do not need both justprep and just with this design - its just just :)  Reducing the number of utilities is generally a good idea.

OTOH if you do not implement a way to isolate the errors to the place they came from, you can get into a little bit of trouble.  The justprep utility does indicate in the final justfile created from the included files via comments which section of the file comes from which included file so that any errors discovered by just are easily traced to the actual cause.

justprep also supports the fake module convention which of course can be easily implemented manually within a justfile but it is som much nicely to just say module followed by the path to the justfile.

casey:
Thanks for the feedback!

[...]

PR #1420, opened 2022-11-22
an include feature for justfiles

kbknapp:
This design takes all lines that look like this:
`@include "relative/path/to/file"`
and replaces them with the contents of the pointed-to file. The resulting buffer is passed along to be parsed and executed as before. That's it: no attempt to look at the file or do anything with the contents. This is pure concatenation.

Feel free to tweak as needed!

Implements #1407

### DIFF
--- a/examples/included.just
+++ b/examples/included.just
@@ -1,0 +1,16 @@
+# This file gets included.
+
+# ANSI color escapes.
+BOLD := '\e[1m'
+RESET := '\e[0m'
+RED := '\e[31m'
+GREEN := '\e[32m'
+BROWN := '\e[34m'
+BLUE := '\e[34m'
+PURPLE := '\e[35m'
+CYAN := '\e[36m'
+GRAY := '\e[37m'
+
+# Report the animal with the best blep.
+blep:
+  @printf "{{BOLD}}Cat{{RESET}} {{BLUE}}blep{{RESET}} is best."
--- a/examples/included.just
+++ b/examples/included.just
@@ -1,15 +1,15 @@
 # This file gets included.
 
 # ANSI color escapes.
-BOLD := '\e[1m'
-RESET := '\e[0m'
-RED := '\e[31m'
-GREEN := '\e[32m'
-BROWN := '\e[34m'
 BLUE := '\e[34m'
-PURPLE := '\e[35m'
+BOLD := '\e[1m'
+BROWN := '\e[34m'
 CYAN := '\e[36m'
 GRAY := '\e[37m'
+GREEN := '\e[32m'
+PURPLE := '\e[35m'
+RED := '\e[31m'
+RESET := '\e[0m'
 
 # Report the animal with the best blep.
 blep:
--- a/src/loader.rs
+++ b/src/loader.rs
@@ -19,8 +19,8 @@ impl Loader {
 
     // Parse include directives.
     // Include directives are lines that look like:
-    // #include "relative/path/to/file"
-    let include_regexp = Regex::new(r#"^#include "([^"]+)"$"#).unwrap();
+    // @include "relative/path/to/file"
+    let include_regexp = Regex::new(r#"^@include "([^"]+)"$"#).unwrap();
 
     let parent = path.parent().map_or_else(|| Path::new(""), |p| p);
 
--- a/src/loader.rs
+++ b/src/loader.rs
@@ -16,6 +16,34 @@ impl Loader {
       path: path.to_owned(),
       io_error,
     })?;
-    Ok(self.arena.alloc(src))
+
+    // Parse include directives.
+    // Include directives are lines that look like:
+    // #include "relative/path/to/file"
+    let include_regexp = Regex::new(r#"^#include "([^"]+)"$"#).unwrap();
+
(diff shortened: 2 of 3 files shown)


examples/included.just
```diff
@@ -0,0 +1,16 @@
...
+# ANSI color escapes.
+BOLD := '\e[1m'
+RESET := '\e[0m'
+RED := '\e[31m'
+GREEN := '\e[32m'
+BROWN := '\e[34m'
+BLUE := '\e[34m'
+PURPLE := '\e[35m'
+CYAN := '\e[36m'
+GRAY := '\e[37m'
```
mihaigalos: Let's sort these lines:
```suggestion
BLUE := '\e[34m'
BOLD := '\e[1m'
BROWN := '\e[34m'
CYAN := '\e[36m'
GRAY := '\e[37m'
GREEN := '\e[32m'
PURPLE := '\e[35m'
RED := '\e[31m'
RESET := '\e[0m'
```

src/loader.rs
```diff
@@ -16,6 +16,34 @@ impl Loader {
       path: path.to_owned(),
       io_error,
     })?;
-    Ok(self.arena.alloc(src))
+
+    // Parse include directives.
+    // Include directives are lines that look like:
+    // #include "relative/path/to/file"
+    let include_regexp = Regex::new(r#"^#include "([^"]+)"$"#).unwrap();
```
mihaigalos: Your PR description uses the syntax "@include". Here you use "#include", also in the comment. Let's unify.

src/loader.rs
```diff
@@ -16,6 +16,34 @@ impl Loader {
...
+          io_error,
+        })?;
+        buf.push_str(&contents);
+      } else {
+        buf.push_str(line);
+      }
+      buf.push('\n');
+    }
+
+    Ok(self.arena.alloc(buf))
```
mihaigalos, line 47: This file currently adds functionality and requires manual validation.
It is missing a minimal automatic test, can you please add one?

src/loader.rs
mihaigalos: Let's extract this logic to a separate, testable function which we call from here.

mihaigalos (commented):
Nice work, needs a bit of improvement.

casey:
@mihaigalos This is just a draft, no need to review. I'll probably just use it as a starting point. I'd like to integrate include into the parser, instead of as a preprocessor step.

mihaigalos:
> @mihaigalos This is just a draft, no need to review. I'll probably just use it as a starting point. I'd like to integrate include into the parser, instead of as a preprocessor step.

That's fine. I would really like this feature.

src/loader.rs
kbknapp: Typo from the implementation we used at my work which was `#include`. I'll fix it up here in a minute.

kbknapp:
Rebased on the latest `master`. Thanks @casey! Let me know if there's anything you'd like me to help with or change to make it easier for you.

casey:
Okay! I hoped to use this as a basis for a version that would parse included justfiles, and then include the token stream, instead of doing literal text inclusion, but I've been swamped and I know that this has been a long requested feature, so instead of delaying it, let's get this in.

Some comments:

- This should have a test, maybe in `tests/include.rs`
- How about `!` instead of @? `@` is used for quiet recipes, so `!` seems more unambiguous. Also open to other suggestions!
- How about `!include PATH_TO_FILE` instead of `!include "PATH_TO_FILE"`? The argument to include isn't an actual just string, e.g. you can't use single quotes, so the quotes might be more confusing than anything else.
- This should have require the `--unstable` flag. I'm open to stabilizing it as-is, since it doesn't conflict with a future version of the feature which uses different syntax, but I'd like to get some testing beforehand.

neunenak:
In the interest of getting #1407 closed and this functionality working ASAP, I've made a PR https://github.com/casey/just/pull/1470 that solves the same problem. It's loosely based off of this one, but I had to make the code a bit more complicated to handle recursive justfile inclusion.

mihaigalos:
> * How about `!include PATH_TO_FILE` instead of `!include "PATH_TO_FILE"`? The argument to include isn't an actual just string, e.g. you can't use single quotes, so the quotes might be more confusing than anything else.

Why have any characters before the `include`? Since if you don't have a colon (`:`) in the line when specifying an include, you are not defining a recipe.

Good idea on dropping the quotes, they only add noise.

casey:
> Why have any characters before the `include`? Since if you don't have a colon (`:`) in the line when specifying an include, you are not defining a recipe.

Without the `!` If the path contains a `:`, then it's ambiguous whether it's an include or a recipe named `include`, e.g.:

```
include :foo
```

mihaigalos:
> If the path contains a `:`

I'm very biased since no ':' in Linux paths afaik, but can imagine `C:\` on Windows. Ok. 😅

casey:
> I'm very biased since no ':' in Linux paths afaik, but can imagine C:\ on Windows. Ok. 😅

Ultimately, linux paths are arbitrary strings, and can include any byte except nulls.

casey:
This has been implemented by @neunenak in #1470! For now it's just in master, but I'll cut a release soon so people can play with it.

PR #1427, opened 2022-11-27
Differentiate between shell and nushell example

Dialga:


casey:
Nice, thanks for the PR!

PR #1440, opened 2022-12-07
Changelog links

nk9:
Navigating the changelog is a chore right now because the issue numbers aren't links. I've run a couple regexes to add links to the existing file.

The one annoyance with this is that it means you'll need to linkify items while adding them. However, they are all quite consistent, and I'm wondering if you don't already have a script for this that could be tweaked to generate the links? Another alternative I guess would be to have a script which runs on commit and linkifies any unlinked issues and users. This is a bit more work, and I'm not sure it's worth the effort, but LMK.


casey:
Nice, great change!

ngirard:
Indeed, nice change, much appreciated !

Thanks @nk9, @casey, and happy new year !

PR #1442, opened 2022-12-12
Eliminate lazy_static

camsteffen:
Just a little cleanup.


camsteffen:
const Mutex::new requires Rust 1.63

casey:
Thanks for the PR!

I like this change, but I don't think it's worth bumping the MSRV so aggressively. 1.65 was released on Nov. 3, so a little over a month ago. Just doesn't have a formal MSRV policy, but I think I'd like to support at least the last couple of Rust releases.

camsteffen:
It could be 1.63.

casey:
I'm rethinking this. I've tried to be conservative with my MSRV policy, but I don't have a super clear memory for why. I think I want to revisit the reasons for that conservatism, come up with a concrete policy (e.g. support last two Rust releases) and then we can figure out hat new policy allows merging this. Hang tight!

camsteffen:
I would think MSRV doesn't matter much since this isn't a library?

casey:
Just is packaged for a number of distros, and if their version of rust lags behind Just's MSRV, they won't be able to package it, and users won't be able to build it from source. I'm not sure if this is a problem in practice though.

casey:
Based on surveying the internet, it seems like a conservative MSRV isn't super important for binary crates, so I decided to just switch MSRV to stable.

Merged, with a bunch of fixes for latest clippy.

Thanks for the PR!

PR #1443, opened 2022-12-15
Move to dotenvy crate

mike-burns:
In light of [RUSTSEC-2021-0141][], it's probably prudent to move off of the `dotenv` crate. The `dotenvy` crate is a maintained fork.

[RUSTSEC-2021-0141]: https://rustsec.org/advisories/RUSTSEC-2021-0141.html


casey:
Seems reasonable to me! Nice that we can remove the the allow(deprecated), that's been bugging me for years 😂

PR #1447, opened 2022-12-15
Update readme codeblocks

### LINKED ISSUE #1372: Use `just` for readme codeblocks and mention that it works in readme
Thanks to @mihaigalos, [just language support has been added to linguist](https://github.com/github/linguist/pull/6086), GitHub's syntax highlighter. Once support lands, we should switch the readme to use `just` instead of `make` for syntax highlighting.

Very exciting!

Here's an example just codeblock, when it renders in color we can make the switch:

```just
x := "world"

foo:
  echo 'Hello, {{x}}!'
```

nicochatzi:
https://github.com/github/linguist/pull/6086 is in so this PR just replace uses of ` ```make ``` ` with ` ```just ``` ` in the READMEs. 

Closes : #1372

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ If you'd rather not install `sh`, you can use the `shell` setting to use the she
 
 Like PowerShell:
 
-```make
+```just
 # use PowerShell instead of sh:
 set shell := ["powershell.exe", "-c"]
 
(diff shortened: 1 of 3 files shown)


casey:
Nice, thanks for the PR!

PR #1448, opened 2022-12-15
Clickable images on landing page

nk9:
Every time I load the just.systems page, I want to be able to click on the enormous images that change on hover. This lets you do that.


casey:
Nice, seems reasonable to me! Thanks for the PR.

PR #1455, opened 2022-12-20
Update Sublime syntax instructions

nk9:
I've removed the `.sublime-syntax` file from the `extras` directory as well, since it's now well out of date. It uses an earlier version of the Sublime syntax definition, which is somewhat closer to the original `.tmLanguage` format. There may be a case for keeping it so that users with older versions of Sublime can fall back on it. The version in the Just package requires Sublime Text 4075+, which was [released in July 2020](https://www.sublimetext.com/dev). I think 2.5 years is enough time to give users to upgrade, but I wanted to highlight so you can make the final decision.


nk9:
Flagging change to the README for @hustcer.

casey:
Nice! Can you also link to the sublime text repo on GitHub, in addition package control? I figure that some users might not be familiar with or use Package Control, so a link directly to the repo might be helpfu.

nk9:
Sure, LMK if this is what you had in mind.

casey:
Nice, looks good to me!

nk9:
Looks like this missed the release because the branch is out of date? I don't really understand auto-merge. LMK if I need to do anything else.

casey:
Ahh, sorry, I hit auto merge, but a test failed, so it never actually merged. The test is flaky, so it's unrelated to this PR. I'll try again.

TonioGela:
Thanks @nk9! I saw the update, and now I'm using your new version :)

PR #1459, opened 2022-12-25
Add another way for the GitHub Action installation

azzamsa:
I don't think I have capability to update the [README.中文.md](https://github.com/casey/just/blob/master/README.%E4%B8%AD%E6%96%87.md) too.


casey:
Nice, merged!

PR #1464, opened 2022-12-30
Let pages run

jsoref:
Organizations can be configured to not give write permissions by default.
The pages job needs permissions to write.


casey:
Thanks for the PR! In what circumstance is this necessary? Is it so that actions can run on a private fork in an organization where write permissions are disabled by default?

jsoref:
Public fork in an org w/ restrictive permissions, e.g. here: https://github.com/check-spelling/casey-just/actions/runs/3797500893

casey:
Makes sense, thanks for the PR!

PR #1470, opened 2023-01-02
Support including a justfile from another justfile

### LINKED ISSUE #1407: Add quick-and-dirty justfile inclusion

neunenak:
This PR adds support for including the contents of another justfile verbatim, using the syntax `!include <path-to-file>`. There is also some basic checking that justfiles cannot recursively include an already-processed justfile.

Should close https://github.com/casey/just/issues/1407

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -89,7 +89,7 @@ pub(crate) enum Error<'src> {
     message: String,
   },
   IncludeMissingPath {
-    justfile: PathBuf,
+    file: PathBuf,
     line: usize,
   },
   InitExists {
@@ -500,7 +500,7 @@ impl<'src> ColorDisplay for Error<'src> {
         )?;
       }
       IncludeMissingPath {
-        justfile, line
+        file: justfile, line
       } => {
 
         write!(
--- a/src/error.rs
+++ b/src/error.rs
@@ -98,6 +98,9 @@ pub(crate) enum Error<'src> {
   Internal {
     message: String,
   },
+  InvalidDirective {
+    line: String,
+  },
   Io {
     recipe: &'src str,
     io_error: io::Error,
@@ -128,10 +131,6 @@ pub(crate) enum Error<'src> {
   TmpdirIo {
     recipe: &'src str,
     io_error: io::Error,
-  },
-  TrailingInclude {
-    justfile: PathBuf,
-    line: usize,
   },
   Unknown {
     recipe: &'src str,
@@ -523,6 +522,9 @@ impl<'src> ColorDisplay for Error<'src> {
           message
         )?;
       }
+      InvalidDirective { line } => {
+          write!(f, "Invalid directive: {}", line)?;
+      }
       Io { recipe, io_error } => {
         match io_error.kind() {
           io::ErrorKind::NotFound => write!(
@@ -597,14 +599,6 @@ impl<'src> ColorDisplay for Error<'src> {
         "Recipe `{recipe}` could not be run because of an IO error while trying to create a temporary \
          directory or write a file to that directory`:{io_error}",
       )?,
-      TrailingInclude { justfile, line } => {
-        write!(
-          f,
-          "!include statement in {} line {} occurs after the first non-blank, non-comment line",
-          justfile.display(),
-          line
-        )?;
-      }
       Unknown {
         recipe,
         line_number,
--- a/src/error.rs
+++ b/src/error.rs
@@ -523,7 +523,7 @@ impl<'src> ColorDisplay for Error<'src> {
         )?;
       }
       InvalidDirective { line } => {
-          write!(f, "Invalid directive: {}", line)?;
+          write!(f, "Invalid directive: {}", line.ordinal())?;
       }
       Io { recipe, io_error } => {
         match io_error.kind() {
--- a/src/error.rs
+++ b/src/error.rs
@@ -500,14 +500,14 @@ impl<'src> ColorDisplay for Error<'src> {
         )?;
       }
       IncludeMissingPath {
-          justfile, line
+        justfile, line
       } => {
 
         write!(
           f,
-          "!include statement in {} line {} has no argument",
+          "!include directive on line {} of `{}` has no argument",
+          line.ordinal(),
           justfile.display(),
-          line
         )?;
 
       },
--- a/src/error.rs
+++ b/src/error.rs
@@ -523,7 +523,7 @@ impl<'src> ColorDisplay for Error<'src> {
         )?;
       }
       InvalidDirective { line } => {
-          write!(f, "Invalid directive: {}", line.ordinal())?;
+        write!(f, "Invalid directive: {line}")?;
       }
       Io { recipe, io_error } => {
         match io_error.kind() {
--- a/src/error.rs
+++ b/src/error.rs
@@ -87,10 +87,6 @@ pub(crate) enum Error<'src> {
   InitExists {
     justfile: PathBuf,
   },
-  Include {
-    path: PathBuf,
-    io_error: io::Error,
-  },
   IncludeRecursive {
     cur_path: PathBuf,
     recursively_included_path: PathBuf,
@@ -496,14 +492,6 @@ impl<'src> ColorDisplay for Error<'src> {
       InitExists { justfile } => {
         write!(f, "Justfile `{}` already exists", justfile.display())?;
       }
-      Include { path, io_error } => {
-        write!(
-          f,
-          "Failed to read included justfile at `{}`: {}",
-          path.display(),
-          io_error
-        )?;
-      },
       IncludeRecursive { cur_path, recursively_included_path } => {
         write!(
           f,
--- a/src/error.rs
+++ b/src/error.rs
@@ -32,8 +32,8 @@ pub(crate) enum Error<'src> {
     io_error: io::Error,
   },
   CircularInclude {
-    cur_path: PathBuf,
-    recursively_included_path: PathBuf,
+    current: PathBuf,
+    include: PathBuf,
   },
   Code {
     recipe: &'src str,
@@ -341,10 +341,10 @@ impl<'src> ColorDisplay for Error<'src> {
           io_error
         )?;
       }
-      CircularInclude { cur_path, recursively_included_path } => {
+      CircularInclude { current, include } => {
         write!(
           f,
-          "Justfile at {} tries to recursively include {}", cur_path.display(), recursively_included_path.display()
+          "Include `{}` in `{}` is a circular include", include.display(), current.display()
         )?;
       },
       Code {
--- a/src/error.rs
+++ b/src/error.rs
@@ -88,6 +88,10 @@ pub(crate) enum Error<'src> {
     function: Name<'src>,
     message: String,
   },
+  IncludeMissingPath {
+    justfile: PathBuf,
+    line: usize,
+  },
   InitExists {
     justfile: PathBuf,
   },
@@ -126,7 +130,8 @@ pub(crate) enum Error<'src> {
     io_error: io::Error,
   },
   TrailingInclude {
-    line_number: usize,
+    justfile: PathBuf,
+    line: usize,
   },
   Unknown {
     recipe: &'src str,
@@ -495,6 +500,18 @@ impl<'src> ColorDisplay for Error<'src> {
           message
         )?;
       }
+      IncludeMissingPath {
+          justfile, line
+      } => {
+
+        write!(
+          f,
+          "!include statement in {} line {} has no argument",
+          justfile.display(),
+          line
+        )?;
+
+      },
       InitExists { justfile } => {
         write!(f, "Justfile `{}` already exists", justfile.display())?;
       }
@@ -580,11 +597,12 @@ impl<'src> ColorDisplay for Error<'src> {
         "Recipe `{recipe}` could not be run because of an IO error while trying to create a temporary \
          directory or write a file to that directory`:{io_error}",
       )?,
-      TrailingInclude { line_number } => {
+      TrailingInclude { justfile, line } => {
         write!(
           f,
-          "!include statement at line {} occurs after the first non-blank, non-comment line",
-          line_number
+          "!include statement in {} line {} occurs after the first non-blank, non-comment line",
+          justfile.display(),
+          line
         )?;
       }
       Unknown {
--- a/src/error.rs
+++ b/src/error.rs
@@ -31,6 +31,10 @@ pub(crate) enum Error<'src> {
     chooser: OsString,
     io_error: io::Error,
   },
+  CircularInclude {
+    cur_path: PathBuf,
+    recursively_included_path: PathBuf,
+  },
   Code {
     recipe: &'src str,
     line_number: Option<usize>,
@@ -87,15 +91,8 @@ pub(crate) enum Error<'src> {
   InitExists {
     justfile: PathBuf,
   },
-  IncludeRecursive {
-    cur_path: PathBuf,
-    recursively_included_path: PathBuf,
-  },
   Internal {
     message: String,
-  },
-  InvalidInclude {
-    line_number: usize,
   },
   Io {
     recipe: &'src str,
@@ -127,6 +124,9 @@ pub(crate) enum Error<'src> {
   TmpdirIo {
     recipe: &'src str,
     io_error: io::Error,
+  },
+  TrailingInclude {
+    line_number: usize,
   },
   Unknown {
     recipe: &'src str,
@@ -337,6 +337,12 @@ impl<'src> ColorDisplay for Error<'src> {
           io_error
         )?;
       }
+      CircularInclude { cur_path, recursively_included_path } => {
+        write!(
+          f,
+          "Justfile at {} tries to recursively include {}", cur_path.display(), recursively_included_path.display()
+        )?;
+      },
       Code {
         recipe,
         line_number,
@@ -492,25 +498,12 @@ impl<'src> ColorDisplay for Error<'src> {
       InitExists { justfile } => {
         write!(f, "Justfile `{}` already exists", justfile.display())?;
       }
-      IncludeRecursive { cur_path, recursively_included_path } => {
-        write!(
-          f,
-          "Justfile at {} tries to recursively include {}", cur_path.display(), recursively_included_path.display()
-        )?;
-      },
       Internal { message } => {
         write!(
           f,
           "Internal runtime error, this may indicate a bug in just: {} \
            consider filing an issue: https://github.com/casey/just/issues/new",
           message
-        )?;
-      }
-      InvalidInclude { line_number } => {
-        write!(
-          f,
-          "!include statement at line {} occurs after the first non-blank, non-comment line",
-          line_number
         )?;
       }
       Io { recipe, io_error } => {
@@ -587,6 +580,13 @@ impl<'src> ColorDisplay for Error<'src> {
         "Recipe `{recipe}` could not be run because of an IO error while trying to create a temporary \
          directory or write a file to that directory`:{io_error}",
       )?,
+      TrailingInclude { line_number } => {
+        write!(
+          f,
+          "!include statement at line {} occurs after the first non-blank, non-comment line",
+          line_number
+        )?;
+      }
       Unknown {
         recipe,
         line_number,
--- a/src/error.rs
+++ b/src/error.rs
@@ -87,6 +87,14 @@ pub(crate) enum Error<'src> {
   InitExists {
     justfile: PathBuf,
   },
+  Include {
+    path: PathBuf,
+    io_error: io::Error,
+  },
+  IncludeRecursive {
+    cur_path: PathBuf,
+    recursively_included_path: PathBuf,
+  },
   Internal {
     message: String,
   },
@@ -485,6 +493,20 @@ impl<'src> ColorDisplay for Error<'src> {
       InitExists { justfile } => {
         write!(f, "Justfile `{}` already exists", justfile.display())?;
       }
+      Include { path, io_error } => {
+        write!(
+          f,
+          "Failed to read included justfile at `{}`: {}",
+          path.display(),
+          io_error
+        )?;
+      },
+      IncludeRecursive { cur_path, recursively_included_path } => {
+        write!(
+          f,
+          "Justfile at {} tries to recursively include {}", cur_path.display(), recursively_included_path.display()
+        )?;
+      },
       Internal { message } => {
         write!(
           f,
--- a/src/error.rs
+++ b/src/error.rs
@@ -98,6 +98,9 @@ pub(crate) enum Error<'src> {
   Internal {
     message: String,
   },
+  InvalidInclude {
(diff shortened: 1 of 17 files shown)


neunenak:
This is ready for a review @casey

neunenak:
It looks like the test I added for this is failing in the macos-latest CI run, but not on my machine. With the debug print commit I added, I can see the output:

```

---- loader::tests::recursive_includes_fail stdout ----
Perform load on /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/temptree7vv8s5/justfile, seen_paths {}
Perform load on /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/temptree7vv8s5/subdir/justfile_b, seen_paths {"/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/temptree7vv8s5/justfile"}
Perform load on /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/temptree7vv8s5/justfile, seen_paths {"/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/temptree7vv8s5/subdir/justfile_b", "/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/temptree7vv8s5/justfile"}
thread 'loader::tests::recursive_includes_fail' panicked at 'assertion failed: (left ~= right)
  left: `IncludeRecursive { cur_path: "/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/temptree7vv8s5/justfile", recursively_included_path: "/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/temptree7vv8s5/subdir/justfile_b" }`
 right: `Error::IncludeRecursive { cur_path, recursively_included_path } if
cur_path == tmp.path().join("subdir").join("justfile_b") &&
    recursively_included_path == tmp.path().join("justfile")`', src/loader.rs:206:5


failures:
    loader::tests::recursive_includes_fail

test result: FAILED. 461 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.17s
```

So there's some discrepency between `/var` and `/private/var` with the way that I'm creating tempfiles in the test. @casey any suggestions here?

casey:
My guess is that it's because on macos, $TEMPDIR is a symlink, and if you canonicalize a path that includes a symlink, the path will change, so you might not get the output that you're expecting.

I thought of a potential issue here. Strings allow newlines, so it's possible for an include to be processed inside of a string. For example:

```
foo :="
!include bar
"
```

This is unlikely, but we should avoid it. We could only allow includes if there have been any non-blank, non-comment lines earlier in the file.

neunenak:
> My guess is that it's because on macos, $TEMPDIR is a symlink, and if you canonicalize a path that includes a symlink, the path will change, so you might not get the output that you're expecting.

Ah, okay. So if I make sure to call `.canonicalize()` on both code paths, that should solve the OS X test issue I think.
 
> I thought of a potential issue here. Strings allow newlines, so it's possible for an include to be processed inside of a string.

Huh, that's a good point. The best solution is probably to actually do parsing on the justfile as written before processing includes, but that would require significant redesign. Maybe it's better to leave that for modules. For now I think disallowing includes after the first non-blank, non-comment line is fine.

casey:
> Ah, okay. So if I make sure to call .canonicalize() on both code paths, that should solve the OS X test issue I think.

Not sure if it's possible, If you can avoid canonicalize, and just make the test more forgiving, maybe with a regex, that would probably be better. Canonicalized paths wind up being rather weird. I believe on windows especially, where I think it converts them to UNC paths, which strike most users as odd. 

> Huh, that's a good point. The best solution is probably to actually do parsing on the justfile as written before processing includes, but that would require significant redesign.

Definitely. I started working on a version that did just that, but it was tricky so I never finished it.

neunenak:
> Not sure if it's possible, If you can avoid canonicalize, and just make the test more forgiving, maybe with a regex, that would probably be better. Canonicalized paths wind up being rather weird. I believe on windows especially, where I think it converts them to UNC paths, which strike most users as odd.

The reason I wanted to call `canonicalize()` is because otherwise it would be possible to include the same justfile using two different paths to identify it, which would wind up being different `PathBuf`s in the hashmap and wouldn't test positive for overlap. E.g. if we have:

```
justfile_a -> !include ./subdir/justfile_b
subdir/justfile_b -> !include ./subdir2/justfile_c
subdir/subdir2/justfile_c -> !include ../../justfile_a
```

we'd have an infinite circular dependency but all of the actual paths involved are different. 

We could maybe try to be less anal about detecting circular includes, but I think this is something where we should try particularly hard to have a good error message, because the "default" behavior is recursively including the same files until just crashes, which is a bad user experience.

casey:
You could use the `Path::lexiclean` function as an alternative to `canonicalize`. It cleans paths lexically, i.e., only by looking at their contents, and doesn't look at symlinks on disk. For example:

```
Path::new("subdir/subdir2/../../justfile_a").lexiclean() == Path::new("justfile_a")
```

neunenak:
> You could use the `Path::lexiclean` function as an alternative to `canonicalize`. It cleans paths lexically, i.e., only by looking at their contents, and doesn't look at symlinks on disk. For example:
> 
> ```
> Path::new("subdir/subdir2/../../justfile_a").lexiclean() == Path::new("justfile_a")
> ```

That's a handy crate to know that I wasn't previously aware of. Modified this PR to use that! @casey

kbknapp:
A quick note, I saw this comment in #1420 talking about this PR:

>  It's loosely based off of this one, but I had to make the code a bit more complicated to handle recursive justfile inclusion.

In our private fork of Just that added this include feature we found recursive inclusions to be very error prone and confusing for the users due to how recipe overrides work (and the lack of variable overrides). Which is why we ultimately opted for non-recursive includes which helped a ton.

Just my 2 cents.

casey:
@kbknapp Thanks for the color! Can you elaborate a little bit about what was confusing about recursive includes? I'm not sure I understand.

kbknapp:
(note the issues we experienced could have been helped if we'd also implemented specific error messages, etc. but we didn't do that for expediency. Also keep in mind we were/are using Just in a large monorepo which means the _users_ of the included recipes may not have been the original authors).

The two issues we ran into frequently were "variable already declared error" (since we didn't implement overriding variables), or finding which version of an overridden recipe was actually being run.

In either case (recursive justfiles or non-recursive) we found you'd have to manually look at (or `cat` all the justfiles into a single large file) to find out in what order things were included and then manually/mentally walk through the file to figure out where the variable was declared, or recipe was being overridden.

With recursive justfiles this lead us to have small building block style include files, so we ended up with tons of small justfiles that would be combined into intermediate justfiles that were effectively _only_ includes built-up of all these building block style files. The final "leaf" justfile that people actually ran would then also include many of these intermediate include justfiles along with some additional building block justfiles and override any recipes necessary. 

While this was a very satisfying outcome, and if you knew what was happening it was pretty magical, it also confused anyone who didn't really dig into the justfiles. Or anyone who only needed to add a new recipe for their team without really worrying about all the included files. So that debugging process I described above was a ton more work and many users got frustrated because they missed an include somewhere, or because imagine you have `#include "variables.just"` so the user didn't actually look at `variables.just` because they're dealing with a recipe override, but it turned out `variables.just` contained `#include "recipes.just"`.

It was also easy to end up in an include cycle (mostly due to how we these small building blocks set up, but in our defense recursive strategies beg for that style of set up 😅)

Of course these issues could be attributed to a combination of poor naming, or bad practices but it's still things we saw regularly. 

Once we went with non-recursive justfiles those issues mostly disappeared. If you saw `#include "foo.just"` could be sure it wasn't itself including anything else you'd have to trace down.

This meant we had to do away with the building-block style justfiles, but in the end the system is much easier to grok. If you're debugging a justfile and you don't see `#include "foo.just"` then you can be sure `foo.just` isn't involved. Cycles are also impossible.

Again, just our anecdote which is a relatively small sample size all things considered.

neunenak:
> (note the issues we experienced could have been helped if we'd also implemented specific error messages, etc. but we didn't do that for expediency. Also keep in mind we were/are using Just in a large monorepo which means the _users_ of the included recipes may not have been the original authors).
> 
> The two issues we ran into frequently were "variable already declared error" (since we didn't implement overriding variables), or finding which version of an overridden recipe was actually being run.
> 
> In either case (recursive justfiles or non-recursive) we found you'd have to manually look at (or `cat` all the justfiles into a single large file) to find out in what order things were included and then manually/mentally walk through the file to figure out where the variable was declared, or recipe was being overridden.
> 
> With recursive justfiles this lead us to have small building block style include files, so we ended up with tons of small justfiles that would be combined into intermediate justfiles that were effectively _only_ includes built-up of all these building block style files. The final "leaf" justfile that people actually ran would then also include many of these intermediate include justfiles along with some additional building block justfiles and override any recipes necessary.
> 
> While this was a very satisfying outcome, and if you knew what was happening it was pretty magical, it also confused anyone who didn't really dig into the justfiles. Or anyone who only needed to add a new recipe for their team without really worrying about all the included files. So that debugging process I described above was a ton more work and many users got frustrated because they missed an include somewhere, or because imagine you have `#include "variables.just"` so the user didn't actually look at `variables.just` because they're dealing with a recipe override, but it turned out `variables.just` contained `#include "recipes.just"`.
> 
> It was also easy to end up in an include cycle (mostly due to how we these small building blocks set up, but in our defense recursive strategies beg for that style of set up sweat_smile)
> 
> Of course these issues could be attributed to a combination of poor naming, or bad practices but it's still things we saw regularly.
> 
> Once we went with non-recursive justfiles those issues mostly disappeared. If you saw `#include "foo.just"` could be sure it wasn't itself including anything else you'd have to trace down.
> 
> This meant we had to do away with the building-block style justfiles, but in the end the system is much easier to grok. If you're debugging a justfile and you don't see `#include "foo.just"` then you can be sure `foo.just` isn't involved. Cycles are also impossible.
> 
> Again, just our anecdote which is a relatively small sample size all things considered.

So what would happen if justfile_a includes justfile_b where justfile_b has its own include statements? Would they just be completely ignored by justfile_a?

casey:
@kbknapp Gotcha, thanks for all the detail! 

I can think of a few ways to address this issues. One is cycle detection, so if you recursively include the same justfile multiple times, you'll get an error.

Another is to add the ability to dump the final justfile, which would include the paths of what was included.

You might have:

justfile:
```just
!include bar.just

foo:
  echo foo
```

bar.just:
```just
bar:
  echo bar
```

And when you do `just --dump` you would get:

```just
# <bar.just>

bar:
  echo bar

# </bar.just>

foo:
  echo foo
```

Do you think that that would be enough to make dealing with recursive justfiles tractable?

neunenak:
Since it seems like there's still some discussion to be had about exactly how `!include` ought to work, I've added a commit that gates it behind an unstable option, so the exact behavior can be tweaked if necessary.

kbknapp:
> Do you think that that would be enough to make dealing with recursive justfiles tractable?

Those would definitely help a ton and I'd be happy to use them! I'd still think making includes recursive opens the complexity door a little too wide too early due to these question. Making the errors reflect any includes that are happening, and adding variable overrides would also help in these scenarios too.

Additionally, you can add recursion to includes in a backwards compatible manner, but you can't remove it in a backwards compatible manner.

Either way though, I'd be happy for the feature 😄

casey:
> Additionally, you can add recursion to includes in a backwards compatible manner, but you can't remove it in a backwards compatible manner.

We can always stabilize single-level includes, but leave recursive includes unstable.

@neunenak Let me know if this is ready to review.

neunenak:
@casey ready for review but I am baffled as to why the generate-completions script is failing in the linter CI, it works on my machine fine.

casey:
When you run `just generate-completions` locally, does it produce a diff? That CI job complains if generate-completions generates a diff.

neunenak:
> When you run `just generate-completions` locally, does it produce a diff? That CI job complains if generate-completions generates a diff.

Ah, yes, it does. So okay that's the problem, I guess this is because I added a new option. Not sure what the right way to fix this is immediately.

neunenak:
Or do I just need to commit the diff?

neunenak:
Okay that was the problem. Ready for review @casey

src/error.rs
```diff
@@ -87,9 +87,16 @@ pub(crate) enum Error<'src> {
...
   },
+  IncludeRecursive {
+    cur_path: PathBuf,
+    recursively_included_path: PathBuf,
+  },
   Internal {
     message: String,
   },
+  InvalidInclude {
+    line_number: usize,
```
casey: We should include the path and the line number, since this error might originate from an include, and not the root justfile.

src/error.rs
```diff
@@ -87,9 +87,16 @@ pub(crate) enum Error<'src> {
   InitExists {
     justfile: PathBuf,
   },
+  IncludeRecursive {
+    cur_path: PathBuf,
+    recursively_included_path: PathBuf,
+  },
```
casey: I think we can improve this error. If it only prints the current path, and the path that we've already seen, it might not be clear why we've already seen this path. Check out `CompileError::{CircularRecipeDependency, CircularVariableDependency}` for example formatting.

The ideal form would be something like:

```suggestion
  CircularInclude {
    prefix: Vec<PathBuf>,
    cycle: Vec<PathBuf>,
  },
```

`prefix` is the includes that got us to this point, and `cycle` is the circular inclusion that triggered the error.

The prefix helps the user find where the cycle started, and the cycle helps the user figure out what the actual cycle was.

src/error.rs
```diff
@@ -87,9 +87,16 @@ pub(crate) enum Error<'src> {
...
   Internal {
     message: String,
   },
+  InvalidInclude {
```
```suggestion
  TrailingInclude {
```

src/subcommand.rs
```diff
@@ -177,7 +177,12 @@ impl Subcommand {
     loader: &'src Loader,
     search: &Search,
   ) -> Result<(&'src str, Ast<'src>, Justfile<'src>), Error<'src>> {
-    let src = loader.load(&search.justfile)?;
+    let src = if config.includes {
+      config.require_unstable("--includes is currently an unstable option")?;
+      loader.load_with_includes(&search.justfile)?
+    } else {
+      loader.load_without_includes(&search.justfile)?
+    };
```
casey:
```suggestion
    let src = if config.unstable {
      loader.load_with_includes(&search.justfile)?
    } else {
      loader.load_without_includes(&search.justfile)?
    };
```

src/loader.rs
```diff
@@ -1,4 +1,9 @@
 use super::*;
+use std::collections::HashSet;
+
+// This regex defines the syntax for a Justfile include statement as `!include <file-path>`
```
casey:
```suggestion
// This regex defines the syntax for an include statement as `!include <file-path>`
```

src/loader.rs
```diff
@@ -11,11 +16,243 @@ impl Loader {
     }
   }
 
-  pub(crate) fn load<'src>(&'src self, path: &Path) -> RunResult<&'src str> {
+  pub(crate) fn load_without_includes<'src>(&'src self, path: &Path) -> RunResult<&'src str> {
```
casey: I like just calling the base case `load`:

```suggestion
  pub(crate) fn load<'src>(&'src self, path: &Path) -> RunResult<&'src str> {
```

src/loader.rs
```diff
@@ -11,11 +16,243 @@ impl Loader {
...
     let src = fs::read_to_string(path).map_err(|io_error| Error::Load {
       path: path.to_owned(),
       io_error,
     })?;
```
casey: Let's refactor this into a shared function, `Loader::load_file`  that returns a String.

src/loader.rs
```diff
@@ -11,11 +16,243 @@ impl Loader {
...
     Ok(self.arena.alloc(src))
   }
+
+  pub(crate) fn load_with_includes<'src>(&'src self, path: &Path) -> RunResult<&'src str> {
+    let src = self.perform_load(path, HashSet::new())?;
+    Ok(self.arena.alloc(src))
+  }
+
+  fn perform_load(&self, path: &Path, seen_paths: HashSet<PathBuf>) -> RunResult<String> {
```
casey: Not sure about the name, how about:

```suggestion
  fn load_with_includes_recursive(&self, path: &Path, seen_paths: HashSet<PathBuf>) -> RunResult<String> {
```

src/loader.rs
```diff
@@ -11,11 +16,243 @@ impl Loader {
...
+    let src = fs::read_to_string(path).map_err(|io_error| Error::Load {
+      path: path.to_owned(),
+      io_error,
+    })?;
```
casey:
```suggestion
    let src = Self::load_file(path)?;
```

src/loader.rs
```diff
@@ -11,11 +16,243 @@ impl Loader {
...
+  /// Note that this does not do any parsing yet (i.e. nothing stops a justfile from including a
+  /// file that is not a valid justfile), and that (currently) line numbers in error messages will
+  /// reference lines in this concatenated String rather than probably-more-useful information
+  /// about the original file an error came from.
+  ///
+  /// !include directives are allowed to appear in a justfile only before the first non-blank,
+  /// non-comment line. This is because this processing step occurs before parsing, and we don't
+  /// want to accidentally process the string "!include" when it happens to occur within a string
+  /// or in another post-parsing syntactic context where it shouldn't be.
+  fn process_includes(
```
casey: This is only called at the end of `load_with_includes_recursive`, which is itself short, so I'd just merge these two functions.

src/loader.rs
```diff
@@ -11,11 +16,243 @@ impl Loader {
...
+    current_justfile_path: &Path,
+    seen_paths: HashSet<PathBuf>,
+  ) -> RunResult<String> {
+    let has_final_newline = original.ends_with('\n');
+
+    let include_regexp = Regex::new(INCLUDE_REGEX).unwrap();
+
+    //NOTE this string-processing code can be made a bit cleaner once the Rust std lib Intersperse
+    //API is stabilized (https://doc.rust-lang.org/std/iter/struct.Intersperse.html)
+
```
casey: I tend to avoid in-tree notes like this. If it's important to track, it should be an issue.

```suggestion
```

src/loader.rs
```diff
@@ -1,4 +1,9 @@
 use super::*;
+use std::collections::HashSet;
+
+// This regex defines the syntax for a Justfile include statement as `!include <file-path>`
+// occurring at the start of a line, and as the only contents of that line
+const INCLUDE_REGEX: &str = "^!include\\s+(.+)$";
```
casey: I think we should match `!include` on its own:

```suggestion
const INCLUDE_REGEX: &str = "^!include(\\s(.+))?$";
```

Which would be an error.

src/loader.rs
```diff
@@ -11,11 +16,243 @@ impl Loader {
...
+  ///
+  /// Note that this does not do any parsing yet (i.e. nothing stops a justfile from including a
+  /// file that is not a valid justfile), and that (currently) line numbers in error messages will
+  /// reference lines in this concatenated String rather than probably-more-useful information
+  /// about the original file an error came from.
+  ///
+  /// !include directives are allowed to appear in a justfile only before the first non-blank,
+  /// non-comment line. This is because this processing step occurs before parsing, and we don't
+  /// want to accidentally process the string "!include" when it happens to occur within a string
+  /// or in another post-parsing syntactic context where it shouldn't be.
```
casey: This is a great comment, but I would remove it here and adapt it for the readme.

src/loader.rs
casey: Actually, on second thought, can we can do more work with this regex? Like using it to extract all comments and includes before the first non-blank, non-comment line?

```suggestion
const DIRECTIVE_REGEX: &str = r"(?x)
  (
    ^\s*$ # blank
    ^\#.*$ # comment
    ^(<?P<directive>!.*)$ # directive
    (\n|\r\n)
)*
";
```

This could extract all leading directives, which we could then process individually. All directives other than `!include` are an error.

casey (changes requested):
I don't think we need the `--includes` flag. An `!include` line without the flag is a syntax error, and is unstable anyways. Also, if we stabilize it, it would be cumbersome to have to include the `--includes` flag in every invocation.

See the comment about the regex, but I think we can use a regex to extract and process all directives at once.

Integration tests:

- Test that !include is an error without unstable
- Test that !include works with unstable
- Tests that elicit all error messages

casey:
Actually, thinking more about it, I no longer think using a regex is a good idea.

Instead I think we should do something like this:

```
for line in file.lines() {
  if line.starts_with('!') {
    // process directive
  } else if line.starts_with('#') {
    // comment_line
  } else if line.trim().is_empty { // or equivalent
    // blank
  } else {
    // stop processing lines
    break;
  }
}
```

src/loader.rs
neunenak: I'm reluctant to have this regex do too much work, since it involves codifying the syntax of a valid justfile in multiple places that can get out of sync. Really, I think that using a regex here to do preliminary parsing of one syntactic construct is a quick hack, and all the code that deals with parsing a justfile really ought to live in the parsing module, and at some point we should do the fairly-hefty redesign of having the parsing step take place before includes (or module imports). In the meantime, hopefully any further directives that prove necessary can happen in the parser and don't need to happen in the loader step.

src/loader.rs
neunenak: Makes sense. It's not that important, intersperse would just make this bit of code a few lines more concise. I wouldn't bother making an issue out of it.

casey (approved):

MadBomber:
congratulations I will begin the process of deprecating the Justprep stopgap reprocessor.

PR #1474, opened 2023-01-04
Fix windows support in www/install.sh

bloodearnest:
The current version of install.sh does not quite fully support installing windows binaries.

This change fixes that by:

 - handling the fact that windows releases are zips not tarballs
 - handling `uname -s` output being funky on MINGW compiled bash

Additionally, it adds a step to the CI tests to test install.sh works on multiple platforms. These tests uncovered that `need rev` was failing on windows bash.  However, `rev` is not actually used any more, so I removed that and `need basename`, as that doesn't seem to be used any more either.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -132,11 +132,7 @@ jobs:
     - name: Test
       run: cargo test --all
 
-    - name: Run install.sh
-      shell: bash
+    - name: Test install.sh
       run: |
         bash www/install.sh --to /tmp
-
-    - name: Test installed just works
-      shell: bash
-      run: /tmp/just --help
+        /tmp/just --version
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -131,3 +131,12 @@ jobs:
 
     - name: Test
       run: cargo test --all
+
+    - name: Run install.sh
+      shell: bash
+      run: |
+        bash www/install.sh --to /tmp
+
+    - name: Test installed just works
+      shell: bash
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -131,3 +131,12 @@ jobs:
...
       run: cargo test --all
+
+    - name: Run install.sh
+      shell: bash
+      run: |
+        bash www/install.sh --to /tmp
+
+    - name: Test installed just works
+      shell: bash
+      run: /tmp/just --help
```
casey:
```suggestion
    - name: Test install.sh
      run: |
        bash www/install.sh --to /tmp
        /tmp/just --version
```

casey:
This is a great PR, thank you! I think Windows support in install.sh has been broken for a long time. (Possibly forever, and nobody actually tried it / noticed / reported it 😅)

bloodearnest:
Thanks for your attention, and the merge :tada: 

For context, in case it's of interest, the reason behind this fix is we were using https://github.com/extractions/setup-just to install just in our GH actions.

However, under the hood, this action uses the Github API to get the current release. Due to Github runners sharing IPs, we would relatively frequently get rate limit errors in our CI. The documented workaround is pass our GITHUB_TOKEN as a parameter, but we're not keen to send that to a third party action.

So, we ended up moving to using our own action, based on https://just.systems/install.sh to install the release (as it uses curl rather than GH API), thus avoiding the rate limits. However, we have some projects that need Windows CI, hence this PR.

Thanks for just, its great!

PR #1475, opened 2023-01-04
Support multiple recipe search path arguments

neunenak:
Allow multiple search-path recipe arguments if the recipe search path is the same for all of them, and raise an error message if they have conflicting search paths. 

This means that if you have a subdirectory `subdir` with a Justfile containing `recipe_a` and `recipe_b`, you can now invoke just as `just subdir/recipe_a subdir/recipe_b`, and have it correctly run `recipe_a` and `recipe_b` from that Justfile (the current behavior is that it will try to parse "subdir/recipe_b" as a recipe name in and of itself and fail). 

If you try to specify multiple separate Justfiles using this syntax (e.g. `just subdir1/recipe_a subdir2/recipe_b` where subdir1 and subdir2 contain different Justfiles), `just` will fail with an error message.


src/config_error.rs
```diff
@@ -11,6 +11,8 @@ pub(crate) enum ConfigError {
     message
   ))]
   Internal { message: String },
+  #[snafu(display("Conflicting path-prefixed recipes: `{}` and `{}`", seen, conflicting))]
```
casey:
```suggestion
  #[snafu(display("Conflicting path arguments: `{}` and `{}`", seen, conflicting))]
```

src/config_error.rs
```diff
@@ -11,6 +11,8 @@ pub(crate) enum ConfigError {
...
+  MultipleSearchDir { seen: String, conflicting: String },
```
casey:
```suggestion
  ConflictingSearchDirArgs { seen: String, conflicting: String },
```

src/positional.rs
```diff
@@ -36,41 +36,68 @@ pub struct Positional {
...
-  pub fn from_values<'values>(values: Option<impl IntoIterator<Item = &'values str>>) -> Self {
+  pub(crate) fn from_values<'values>(
+    values: Option<impl IntoIterator<Item = &'values str>>,
+  ) -> Result<Self, ConfigError> {
     let mut overrides = Vec::new();
     let mut search_directory = None;
     let mut arguments = Vec::new();
 
+    let mut processing_step = ProcessingStep::Overrides;
+
```
casey, line 55: This might be cleaner with something like:

```
let values = values.into_iter().peekable();

while let Some(value) = values.peek() {
  match processing_step {
    // do step, using values.next() to advance iterator
  }
}
```

Not sure though, TIOLI.

src/positional.rs
```diff
@@ -36,41 +36,68 @@ pub struct Positional {
...
+        if let ProcessingStep::SearchDir = processing_step {
+          if value == "." || value == ".." {
             search_directory = Some(value.to_owned());
+            continue;
           } else if let Some(i) = value.rfind('/') {
             let (dir, tail) = value.split_at(i + 1);
 
-            search_directory = Some(dir.to_owned());
+            if let Some(ref sdir) = search_directory {
```
casey:
```suggestion
            if let Some(ref seen) = search_directory {
```

src/positional.rs
```diff
@@ -36,41 +36,68 @@ pub struct Positional {
...
+              if sdir != dir {
```
casey:
```suggestion
              if seen != dir {
```

src/positional.rs
```diff
@@ -36,41 +36,68 @@ pub struct Positional {
...
+                return Err(ConfigError::MultipleSearchDir {
+                  seen: sdir.clone(),
```
casey:
```suggestion
                  seen: seen.clone(),
```

src/positional.rs
```diff
@@ -225,4 +252,35 @@ mod tests {
     search_directory: None,
     arguments: ["a", "a=b"],
   }
+
+  test! {
```
casey, line 263: I think these would be better as integration tests. I'd also add one that tries other args, like `.`, `./foo`, `../foo`

casey (changes requested):
I think this is totally reasonable. See review for comments.

src/positional.rs
neunenak: Hm, any reason in particular an integration test would be better? Seems to me like a unit test can exercise the new functionality here, and that it's better to use a unit test than an integration test all else being equal because unit tests are simpler to set up and modify.

src/positional.rs
neunenak: I coded this up - I think it's more or less as clean (which is to say, slightly messy) as the previous solution. Might as well use it since it's added to this PR now unless you really hate it.

jpbochi:
@neunenak I'm not acquainted with the source code here, nor with rust, but I'm interested in this change. Would you mind writing a higher level example[s] of what this change does? 🥺

neunenak:
> @neunenak I'm not acquainted with the source code here, nor with rust, but I'm interested in this change. Would you mind writing a higher level example[s] of what this change does? pleading_face

Updated the description to hopefully make the intent of this change clearer.

jpbochi:
@neunenak Thanks! That makes sense to me. I really like the compromise you found.

Should you also update the README.md part about "Invoking justfiles in Other Directories"? Currently, it says that a `/` will only matter in the first argument. With your change, it will matter in other arguments, too.

src/positional.rs
casey: With unit tests, you have to think about whether or not the unit test captures all aspects of the feature or change. In this case, the error test doesn't check the formatting of the error message, which an integration test would. But, engaging my brain for a second, I think it's fine here.

casey (changes requested):
Looks good! I would:

- Add an integration test that checks the formatting of the error message.
- Update the readme

casey (approved):
Nice, reworded the readme a little, but otherwise LGTM!

PR #1478, opened 2023-01-05
Renable install.sh test

casey:


casey:
@bloodearnest btw, I noticed that the install.sh test was flaky, it was failing when it hits the API to get the latest tag. I disabled it, but now I'm reenabling it with `--tag 1.0.0`, which would prevent hitting the API, just downloading the binary. It might wind up that in order for install.sh to not be flaky on Github Actions, you have to give it a tag. Not sure though, since I'm just reenabling it now.

PR #1480, opened 2023-01-06
Allow recipe parameters to shadow variables

casey:
Recipe parameters cannot have the same name as bindings in the environment.

This has been striking me as an unnecessarily conservative restriction, and counter to the way that most programming languages work, so this PR removes it.

I'm interested in feedback before merging.

@neunenak Thoughts?

### DIFF
--- a/tests/shadowing_parameters.rs
+++ b/tests/shadowing_parameters.rs
@@ -1,5 +1,3 @@
-use super::*;
-
 test! {
   name:     parameter_may_shadow_variable,
   justfile: "FOO := 'hello'\na FOO:\n echo {{FOO}}\n",
--- a/tests/shadowing_parameters.rs
+++ b/tests/shadowing_parameters.rs
@@ -1,0 +1,25 @@
+use super::*;
+
+test! {
+  name:     parameter_may_shadow_variable,
+  justfile: "FOO := 'hello'\na FOO:\n echo {{FOO}}\n",
+  args:     ("a", "bar"),
+  stdout:   "bar\n",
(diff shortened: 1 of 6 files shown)


tests/shadowing_parameters.rs
```diff
@@ -0,0 +1,25 @@
...
+  args:     ("a", "bar"),
+  stdout:   "bar\n",
+  stderr:   "echo bar\n",
+}
+
+test! {
+  name:     shadowing_parameters_do_not_change_environment,
+  justfile: "export FOO := 'hello'\na FOO:\n echo $FOO\n",
+  args:     ("a", "bar"),
+  stdout:   "hello\n",
```
neunenak, line 13: It's not immediately clear to me why passing "bar" on the command line doesn't override "hello" here, but it does in the other two cases.

neunenak:
I don't see any issues. In fact, I think this might've made a justfile that I used at my previous job that was trying to provide a default env var value in the justfile but also allow it to be overridden on the command line to be slightly more concise.

tests/shadowing_parameters.rs
casey: In this case, `$FOO` refers to an environment variable, which comes from the outer export. In the first test, `FOO` refers to the just variable, not the environment variable. In the last case, the just parameter `FOO` is exported with `$FOO`, so it now refers to the exported parameter.

casey:
Thanks for the feedback! I'm going to let this marinate for a week and then probably merge it.

PR #1496, opened 2023-01-11
Specify minimum rust version

### LINKED ISSUE #1495: Cargo install fails with latest on Ubuntu 22.04
Cargo build fails on default Ubuntu 22.04

The Cargo / rustc version is 1.61.0.

Fails when running:

```
cargo install just
```

The failure is:

```
error[E0015]: cannot call non-const fn `std::sync::Mutex::<interrupt_handler::InterruptHandler>::new` in statics
  --> /home/fquinn/.cargo/registry/src/github.com-1ecc6299db9ec823/just-1.10.0/src/interrupt_handler.rs:17:48
   |
17 |     static INSTANCE: Mutex<InterruptHandler> = Mutex::new(InterruptHandler::new());
   |                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: calls in statics are limited to constant functions, tuple structs and tuple variants

For more information about this error, try `rustc --explain E0015`.
```

And it happens for both 1.10.0 and 1.11.0 (1.9.0 works fine) if specified explicitly on cargo command line so thinking it may be related to #1442

benmoss:
1.63 is required by Mutex::new added in https://github.com/casey/just/pull/1442

Sorta fixes #1495 ? Doesn't really fix it, but makes it more clear to users that they're not gonna have a good time :)

Example of the error Cargo gives:
> error: package `just v1.11.0 (/workspace)` cannot be built because it requires rustc 1.63 or newer, while the currently active rustc version is 1.62.1

It's not super clear at first glance but [rust-version](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) is just a minimum:
> If the currently selected version of the Rust compiler is older than the stated version, cargo will exit with an error, telling the user what version is required.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,14 +51,6 @@ jobs:
         sudo apt-get install ripgrep
         ./bin/forbid
 
-    - name: cargo-msrv
-      uses: baptiste0928/cargo-install@v1
-      with:
-        crate: cargo-msrv
-
-    - name: Verify MSRV
-      run: cargo-msrv verify
-
   pages:
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,6 +50,15 @@ jobs:
         sudo apt-get update
         sudo apt-get install ripgrep
         ./bin/forbid
+
+    - name: cargo-msrv
+      uses: baptiste0928/cargo-install@v1
+      with:
+        crate: cargo-msrv
+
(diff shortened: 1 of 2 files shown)


benmoss:
Maybe can do something fancy with https://crates.io/crates/cargo-msrv to do something in CI to keep this in sync

casey (changes requested):
I think that cargo-msrv is just too slow to run on CI. I think keeping the MSRV in sync manually is probably the best option.

casey (approved):

PR #1511, opened 2023-01-15
Change setting defaults to lower-case

s1ck:
Using uppercase `False` gives the impression that you need to change it to `True`, but the accepted literals are case-sensitive `true` and `false`.


casey:
Good change, thank you!

s1ck:
I just thought we could also put them in single quotes, to make it even more obvious. Similar to the names. Your call :)

casey:
You mean backticks? E.g., `false`? Yah, that would be good.

s1ck:
> You mean backticks? E.g., `false`? Yah, that would be good.

yep, I'll change it.

PR #1524, opened 2023-01-24
Add -n as a short option for dry-run

maiha:
Thank you for very interesting application. I am a heavy user of make and have one point of concern.

In `make`, the first and most frequently executed command is `-n`. It is very practical to be able to check that it works by **just adding a few letters** just before executing the command.

In `just`, a nice dry-run feature is also implemented, but unfortunately not comfortable **due to its length**. This PR introduces `-n` to make migration from `make` easier and to make just more convenient.

Best regards,

### DIFF
--- a/completions/just.bash
+++ b/completions/just.bash
@@ -20,7 +20,7 @@ _just() {
 
     case "${cmd}" in
         just)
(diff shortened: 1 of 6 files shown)


casey:
Hi @maiha, thank you for the detailed rationale, and PR with tests. That all makes sense, seems like a great addition!

PR #1527, opened 2023-01-25
Analyzer::analyze borrows rather than clones Ast

neunenak:
I was looking at the code around the Analyzer and I realized it was cloning the entire `Ast` object, which didn't seem to be necessary.


casey:
Nice! I removed the `'a` lifetime since I think it isn't necessary.

Issue #1523, opened 2023-01-23
bug: new "conflicting path arguments" error in 1.12.0


colearendt:
In a directory with a subdirectory `sub` and a `Justfile` at `sub/Justfile` with target `test`, the following just command used to work:

```
just-1.11.0 sub/test docker/image
```

As of 1.12.0, we get:
```
just-1.12.0 sub/test docker/image
# error: Conflicting path arguments: `sub/` and `docker/`
```

This error message is problematic for building docker images, which use the "slash" for repository/image distinction. 

The only workaround I am aware of in the short term is to downgrade `just` to 1.11.0, without us rewriting or escaping our build commands.

neunenak:
This is a side-effect of https://github.com/casey/just/pull/1475 which changed how command-line arguments with a `/` in them are interpreted. I didn't realize that it's possible to specify multiple recipes, each with their own parameters, on the command line. 

I think this implies that the code in `positional.rs` that parses command-line arguments can't have the behavior it currently has where there is a `SearchDir` stage and then an `Arguments` stage strictly following it. There needs to be a way of distinguishing arguments to recipes and the names of additional recipes (which may themselves take arguments) when specifying them on the command line. 

Maybe the command line parser should be able to parse a recipe and its argument(s) surrounded by parentheses, similar to the syntax for passing parameters to a dependency in a justfile? E.g. `just (sub/test docker/image)`?

casey:
@neunenak I think you're right. In that case I think that #1475 should probably be reverted. Requiring passing arguments surrounded with parenthesis is pretty annoying, and #1475 was a purely aesthetic change.

casey:
The alternative is using the parsed justfile to figure out what's a recipe and what's an argument.

neunenak:
I'm happy to revert #1475 for now, maybe there's a way to bring it back that functionality without breaking recipe arguments with slashes in the name. Trying to use the parsed justfile to figure out what's a recipe and what's an argument seems rather messy because right now the `positional.rs` code runs before the justfile has even been read from disk.

casey:
Fixed in #1528. Sorry for the breakage!

PR #1528, opened 2023-01-25
Revert "Allow matching search path arguments (#1475)"

neunenak:
https://github.com/casey/just/pull/1475 accidentally broke justfile invocations in certain cases where a recipe argument has a slash character in it (see https://github.com/casey/just/issues/1523). Since #1475 was a purely aesthetic change, it seems best to revert it for now.


casey:
Nice, merged. Also added a test that arguments with different search arguments are allowed.

PR #1534, opened 2023-01-25
Analyze and Compiler tweaks

neunenak:
- make the `Settings` type responsible for constructing itself from an iterator of `Setting` values, which allows the analyzer code to be more concise
- use `Compiler::compile` consistently


casey:
Seems reasonable to me!

PR #1539, opened 2023-01-27
modify the default chooser to use `just --show` to preview

fzdwx:
Overwrite `CHOOSER_DEFAULT`: `fzf` -> `fzf --preview 'just --show {}'`

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,14 @@ temptree = "0.2.0"
 which = "4.0.0"
 yaml-rust = "0.4.5"
 
+[lib]
+doctest = false
+
+[[bin]]
+path = "src/main.rs"
+name = "just"
+test = false
+
 [features]
 # No features are active by default.
 default = []
--- a/completions/just.elvish
+++ b/completions/just.elvish
@@ -49,7 +49,7 @@ edit:completion:arg-completer[just] = [@words]{
             cand -v 'Use verbose output'
             cand --verbose 'Use verbose output'
             cand --changelog 'Print changelog'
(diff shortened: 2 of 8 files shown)


casey (changes requested):
This looks good! It looks like the tests are failing though.

fzdwx:
Ok, test case has been fixed.

1. `invoke_error_function` update regex
2. `default` use `--chooser fzf`

fzdwx:
Cool !

casey:
Merged! Thank you for the PR!

lyderic:
It's nice. Unfortunately, it doesn't work if the justfile is called something else than 'justfile' :-(

For example:

    just --justfile foo --choose

doesn't display the recipes in the right pop-up window, but:

    error: No justfile found

**Note:** This is the same behaviour for shebang _just_ scripts (i.e. scripts starting with '#!/usr/bin/env -S just --justfile')

Is there a workaround? Maybe a way to default to the old, plain fzf behaviour without --preview, when using --justfile <name> or a just shebang script. I tried setting the environment variable CHOOSER_DEFAULT as follows:

    $ CHOOSER_DEFAULT=fzf just --justfile foo --choose

but it didn't help.

Many thanks,
L.

**Steps to reproduce:**

    $ just --version
    just 1.14.0
    $ just --init
    Wrote justfile to `/dev/shm/justfile`
    $ just --choose
    # above command works, I can see the 'default' recipe in the right hand side fzf window
    $ mv -v justfile foo
    renamed 'justfile' -> 'foo'
    $ just --justfile foo --choose
    # this latter command doesn't work (see screenshot)

Screenshot from 2023-06-03 09-47-52

Qeole:
> Unfortunately, it doesn't work if the justfile is called something else than 'justfile'

Also reported in https://github.com/casey/just/issues/1638

PR #1543, opened 2023-02-02
Produce a deb package as part of release

benmoss:
Maybe counts as part of #429 ?


mmstick:
It'd be best to adapt https://github.com/pop-os/just/tree/master/debian

PR #1551, opened 2023-02-17
Support escaping newlines

### LINKED ISSUE #1157: add new line escapes
here is Makefile
```Makefile
rule: many many \
    many many many requirements
%:
    @echo $@
```

it would be nice to have this also in Just

and some more examples
```Makefile
# this is Justfile
# for now ther is `error: Unknown start of token` at backslash

var := "a" \
       "b" \
       "c" \
       "d" \
       "e" \
       "f" \
       "..."

# or `if` inside rule
default:
   echo "hello, {{ \
     if "2" == "2" { \
       "long string so not Good!" \
     } else { \
       "it would be nice if i can split" \
       + " it to more lines 1984" \
     } \
   }}"
```

ids1024:
Fixes https://github.com/casey/just/issues/1157.

### DIFF
--- a/src/lexer.rs
+++ b/src/lexer.rs
@@ -481,6 +481,7 @@ impl<'src> Lexer<'src> {
       ',' => self.lex_single(Comma),
       '/' => self.lex_single(Slash),
       ':' => self.lex_colon(),
+      '\\' => self.lex_escape(),
       '=' => self.lex_choices('=', &[('=', EqualsEquals), ('~', EqualsTilde)], Equals),
       '@' => self.lex_single(At),
       '[' => self.lex_delimiter(BracketL),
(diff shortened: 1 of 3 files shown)


src/lexer.rs
```diff
@@ -709,6 +710,31 @@ impl<'src> Lexer<'src> {
...
+    } else if self.accepted('\r')? {
+      if !self.accepted('\n')? {
+        return Err(self.error(UnpairedCarriageReturn));
+      }
+      while self.next_is_whitespace() {
+        self.advance()?;
+      }
+      self.token(Whitespace);
+    } else if let Some(next) = self.next {
+      return Err(self.error(CompileErrorKind::InvalidEscapeSequence { character: next }));
```
casey:
```suggestion
    } else if let Some(character) = self.next {
      return Err(self.error(CompileErrorKind::InvalidEscapeSequence { character }));
```

casey (changes requested):
Looks good to me! Can you add tests that cover:

1. Escaping a `\r\n`
2. The unpaired `\r\n` error
3. The invalid escape sequence error?

ids1024:
Updated.

casey:
Nice, merged!

nk9:
@casey, @laniakea64 has noticed that this works in a bunch of other cases. See https://github.com/nk9/just_sublime/issues/23 and this gist: https://gist.github.com/laniakea64/b3381389beca11226cb8fefc4e1abf85

Just wanted to check: is this intentional/desired? If it is, there should probably be a lot more tests.

PR #1568, opened 2023-03-23
Add a num_cpus() function

schultetwin1:
This is useful for when you are calling `make` from Just and want to pass a `-j` argument


casey:
Nice, merged!

Issue #974, opened 2021-09-19
Shell completion add nushell support


hustcer:
Hi, Thanks for the great work, and `just` currently has no  completion support for the modern, powerful and elegant [NuShell](https://github.com/nushell/nushell), is there any plan for this? Thx

```
nu --version
───┬────────────────────┬────────────────────────────────────────────────────────────────────────────────────
 # │      Column0       │                                      Column1
───┼────────────────────┼────────────────────────────────────────────────────────────────────────────────────
 0 │ version            │ 0.37.0
 1 │ build_os           │ macos-x86_64
 2 │ rust_version       │ rustc 1.55.0
 3 │ cargo_version      │ cargo 1.55.0
 4 │ pkg_version        │ 0.37.0
 5 │ build_time         │ 2021-09-14 19:45:30 +00:00
 6 │ build_rust_channel │ release
 7 │ features           │ clipboard-cli, ctrlc, dataframe, default, rustyline, term, trash, uuid, which, zip
 8 │ installed_plugins  │
───┴────────────────────┴────────────────────────────────────────────────────────────────────────────────────
```

casey:
You are most welcome!

Shell completions are currently mostly generated by `clap`, the argument parsing library. So support for nushell would have to be added to `clap`, or someone would have to contribute hand-written completion scripts.

hustcer:
https://github.com/clap-rs/clap/issues/2778

hustcer:
https://github.com/clap-rs/clap/issues/2778#issuecomment-992733874  Will this help? Thanks

casey:
Yup, if clap automatically generated nushell completions, that would definitely help.

hustcer:
> Yup, if clap automatically generated nushell completions, that would definitely help.

I mean the last comment they replied: "If nushell works like bash where you have to write your own parser, #3166 could be a big help towards this."  It doesn't look like they are going to provide “automatically generated nushell completions” support in clap

casey:
I think that might make it easier on the clap side to implement nushell completion support, but we would still have to wait for it to be implemented there before Just could take advantage of it.

starthal:
It looks like Clap has adopted this: https://github.com/clap-rs/clap/tree/master/clap_complete_nushell 

Then I imagine this depends on clap v4: https://github.com/casey/just/issues/1084

FrancescElies:
In the meanwhile you can put the following somewhere in your nushell config

```
def "nu-complete just" [] {
[...]

PR #1571, opened 2023-03-29
Nushell completion: add a simple script for recipes

presidento:
Add a basic completion script for Nushell. The best solution would be to use generated completion script from `clap` (see #974), yet this simple script helps me a lot, so I just wanted to share. At this implementation only recipes are completed, but only they changes between projects.

image

image


casey:
Sorry for letting this sit for so long. Merging!

PR #1577, opened 2023-04-06
Update link to github search for justfiles in README

schpet:
the [current link in the README][1] seems to be broken by github's new search

[1]: https://github.com/search?o=desc&q=filename%3Ajustfile&s=indexed&type=Code


nk9:
The current link is working for me. Is it still broken to you? Do you have any documentation on a change in the GitHub search syntax?

schpet:
@nk9 ah! OK i think it might only be broken for me because i'm in github's [code search beta](https://github.com/features/code-search). i thought it was out of beta but apparently not.

i'm going to close this for now, and hope when they launch this they'll support legacy url params. thanks for catching that!

Issue #1575, opened 2023-04-02
Allow `--unstable` to be passed through via environment variables


p5:
It would be great if users can globally opt-in to unstable features by setting an environment variable such as `JUST_ALLOW_UNSTABLE`.  

If that variable is set to any value, it acts the same as if the `--unstable` flag is passed through to the commands.
This still allows the safety of feature-flagging the newest commands, but those who want the absolute latest-and-greatest can set this option system-wide and not have to think about passing through the `--unstable` parameter

Taywee:
This would also be useful for calling `just` inside of a recipe and being sure the unstable option propagates through.

ealtintas:
I think using a directive like `set unstable` in a parent justfile (which may include other justfiles) would also a be good solution for this issue. 
```
error: The !include directive is currently unstable. Invoke `just` with the `--unstable` flag to enable unstable features.
```
How long do you think the '!include' directive should stay as 'unstable' ? @casey

ealtintas:
Related: https://github.com/casey/just/issues/1430 , https://github.com/casey/just/issues/1564

casey:
Fixed in #1588!

PR #1588, opened 2023-04-21
Allow --unstable to be set with an environment variable

neunenak:
Allow the environment variable JUST_ALLOW_UNSTABLE=true to act as equivalent to the --unstable flag. Any other value of JUST_ALLOW_UNSTABLE is treated as false. 

This addresses https://github.com/casey/just/issues/1575

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -570,9 +570,9 @@ impl Config {
     };
 
     let unstable = matches.is_present(arg::UNSTABLE)
-      || std::env::var_os("JUST_UNSTABLE").map_or(false, |val| {
-        !["false", "0", ""].contains(&val.to_string_lossy().as_ref())
-      });
+      || std::env::var_os("JUST_UNSTABLE")
+        .map(|val| !["false", "0", ""].contains(&val.to_string_lossy().as_ref()))
+        .unwrap_or_default();
 
     Ok(Self {
       check: matches.is_present(arg::CHECK),
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,8 +3,6 @@ use {
   clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, ArgSettings},
 };
 
-pub(crate) const UNSTABLE_KEY: &str = "JUST_ALLOW_UNSTABLE";
-
 // These three strings should be kept in sync:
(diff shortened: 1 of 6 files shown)


src/config.rs
```diff
@@ -3,6 +3,8 @@ use {
   clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, ArgSettings},
 };
 
+pub(crate) const UNSTABLE_KEY: &str = "JUST_ALLOW_UNSTABLE";
```
casey: What about `JUST_UNSTABLE` instead? `JUST_UNSTABLE=true` makes sense, and is shorter.

src/config.rs
```diff
@@ -569,6 +571,9 @@ impl Config {
       None
     };
 
+    let unstable = matches.is_present(arg::UNSTABLE)
+      || std::env::var_os(UNSTABLE_KEY).map_or(false, |val| val.to_string_lossy() == "true");
```
casey: I'd be inclined to inline this:
```suggestion
      || std::env::var_os("JUST_UNSTABLE").map_or(false, |val| val.to_string_lossy() == "true");
```

casey (changes requested):
Just a few random nitpicks, see comments. Also: Is there any sort of convention regarding how other tools convert environment variable values are converted to bools? For example, Rust has the `RUST_BACKTRACE` variable, which if unset or set to `0` will not capture a backtrace, but if set to any other value will. I wonder if there's any particular reason to go with one standard vs another.

neunenak:
> Just a few random nitpicks, see comments. Also: Is there any sort of convention regarding how other tools convert environment variable values are converted to bools? For example, Rust has the `RUST_BACKTRACE` variable, which if unset or set to `0` will not capture a backtrace, but if set to any other value will. I wonder if there's any particular reason to go with one standard vs another.

I'm not aware of any convention other than what seems reasonable from the perspective of a developer working with standard programming languages. I'd probably allow both "0" and "false" to be treated as a false boolean, the convention that setting the env var to anything else makes it true, seems reasonable to me.

casey:
Tweaked a little bit and merged. Sorry for letting this sit for so long. In the mean time, `--dump-format json` was stabilized, so I changed the test to use `--fmt` instead.

PR #1589, opened 2023-04-23
add zsh completions for subdirectory recipes

joe-dipilato:


casey:
Thank you for the PR! I tested this out, and `**/*/justfile` is very slow if the current directory contains a lot of subdirectories. Since slow completion scripts are kind of a pet-peeve of mine, I'm going to close this.

Issue #1597, opened 2023-05-11
Idea: Provide an additional `go_arch()` function


nachtjasmin:
First of all, thanks for providing `just`! It's been extremely helpful to me as a dev, really love it to integrate in all my projects.

However, there's one minor inconvenience: When I'm trying to download Go-based tools from external resources, like GitHub, the built-in `arch()` function is not as useful as it could be.

For example, if I wanna download [air](https://github.com/cosmtrek/air/releases/tag/v1.43.0) from GitHub, I need to build by own translation of `x86_64` to `amd64`, just because Go uses a different naming scheme for architectures. It would be helpful to have a `go_arch()` function that does this for me under the hood.

Go [has a lot of reserved architectures](https://github.com/golang/go/blob/3e9876cd3a5a83be9bb0f5cbc600aadf9b599558/src/go/build/syslist.go#L56). Not all of them make sense for just, e.g. `wasm`. A simple mapping from `arch()` to `go_arch()` should be sufficient for most cases, I guess.

casey:
Thanks for opening the issue and the PR! There are a lot of different ways of spelling architectures and targets, so supporting the long tail seems like a something i"m not sure I want to do.

nachtjasmin:
Yeah, that's understandable. If someone in the future is going to need it as well, I [wrote a blog post](https://jasminchen.dev/2023/determine-the-go-architecture-in-a-justfile/) about that.

For simplicity, I've copied the relevant part of the justfile:


```justfile
# Maps arch() to one of the known values inside the syslist.go
# https://github.com/golang/go/blob/3e9876cd3a5a83be9bb0f5cbc600aadf9b599558/src/go/build/syslist.go#L56
go_arch_complete := if arch() =="aarch64" {
    "arm64"
} else if arch() == "arm" {
    "arm"
} else if arch() == "mips" {
    "mips64"
} else if arch() == "powerpc" {
    "ppc"
} else if arch() == "powerpc64" {
    "ppc64"
} else if arch() == "s390x" {
    "s390x"
} else if arch() == "sparc" {
    "sparc64"
} else if arch() == "wasm32" {
    "wasm"
} else if arch() == "x86" {
    "386"
} else if arch() == "x86_64" {
    "amd64"
} else { error("unknown architecture {{ os() }}") }
```

runeimp:
@chloe-the-catgirl Nice! I'm saving that!

PR #1598, opened 2023-05-11
Add a basic `go_arch()` function

nachtjasmin:
Unfortunately, the Go team decided to use a different naming scheme for architectures. This commit marks the beginning of a simple mapping from `target::arch()` to the Go architecture strings which are defined in syscall.go.

Unfortunately, I'm not a Rust developer, this is my first Rust code that I've written. :sweat_smile:
There are open questions and commented out code in the current version, mainly because I'm not sure about the error handling for this method.


casey:
Thank you for the PR! See #1597 for comment.

Issue #1278, opened 2022-07-24
Add shorter alias for `env_var` functions


casey:
These are nice and legible, but a little too long, since they're very common.

Some ideas:

Add a `$` operator which is a prefix operator that dereferences its argument:

```
export PATH := "/foo/bin:" + $"PATH"
```

Add a shorter function:

```
export PATH := "/foo/bin:" + env("PATH")
```

I like the `$` operator, since why not, but I know that I have an extremely high tolerance for line-noise-like syntax.

runeimp:
I like both options. The only downside is the `$"VARNAME"` is very close to the standard POSIX style `$VARNAME` or `${VARNAME}` that everyone using such systems will flub by typing `"$VARNAME"` instead like 98.643% of the time. Yes, that _exact_ amount of the time.  😉 

So I think just adopting standard Bourne shell / POSIX syntax for variables would be _much_ better. Especially if their to be specifically used to access actual environment variables and not internal Justfile variables or some such.

The other downside is that Windows users will either prefer `$Env:VARNAME` or `%VARNAME%` style syntax. And those that are semi-familiar with Bourne shell / POSIX syntax will 99.994% 👼 of the time use the wrong syntax as well.

Oh and don't come back with 'My version of Zsh accepts $"VARNAME" syntax.' because I don't believe it. 😉

Also `env(VARNAME)` is more neutral and stands out as part Just rather than potentially being somehow possibly part of the shell Just is running in. 😁

casey:
I feel you. `$"FOO"` is pretty ugly, and you're right it's not familiar so people will mess it up. `$FOO` is, if you think about it though, pretty weird. `FOO` is usually an identifier that points to a variable, but in this case it's a special way of writing the string `"FOO"`, i.e. the identifier is just interpreted as a bare word string. `$"FOO"` is slightly more flexible. But maybe if the point is to have short syntax, it should be `$FOO`. I think `env` is good too. I feel like `env_var` should be deprecated in favor of `env` whatever we decide about `$`.

nk9:
FWIW, I prefer `env()` and also really don't like `$"FOO"`.

casey:
I think we should add `env` as an alias, since it's obviously better than `env_var`. We could also `env(NAME, DEFAULT)` to replace `env_or_default(NAME, DEFAULT)`. `$"FOO"` is probably just too weird.

nk9:
Yeah, just having "env" do both seems like the right move.

kykyi:
[...]

PR #1613, opened 2023-06-07
New function alias: env()

kykyi:
In relation to https://github.com/casey/just/issues/1278, this PR adds a new `env` function which acts as an alias to both `env_var` and `env_var_or_default`:
```just
env_var('x') => env('x')
env_var_or_default('x', 42) => env('x', 42)
```

To make such a function signature possible, a new `UnaryOpt` enum variant was added which accounts for the majority of the touched files!

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2335,7 +2335,7 @@ mod tests {
   }
 
   error! {
-    name: function_argument_count_unary_plus,
+    name: function_argument_count_unary_opt,
     input: "x := env()",
     offset: 5,
     line: 0,
@@ -2344,7 +2344,7 @@ mod tests {
     kind: FunctionArgumentCountMismatch {
       function: "env",
       found: 0,
-      expected: 1..usize::MAX,
+      expected: 1..2,
     },
   }
 
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2335,7 +2335,21 @@ mod tests {
   }
 
   error! {
-    name: function_argument_count_unary_opt,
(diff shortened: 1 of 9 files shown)


casey:
Thanks for the PR! It looks like the `env` function will silently ignore excess arguments, which isn't great. How about something like:

```
  UnaryOpt(fn(&FunctionContext, &str, Option<&str>) -> Result<String, String>),
```

Which accepts only one or two arguments?

kykyi:
Good call @casey! I have updated to introduce UnaryOpt 👌

src/parser.rs
```diff
@@ -2334,6 +2334,34 @@ mod tests {
     },
   }
 
+  error! {
```
kykyi, line 2336: These tests break the `function_argument_<variant>` convention by adding a too-many and a too-few test, unsure if this is what you want or should just have the single test.

src/thunk.rs
```diff
@@ -60,6 +67,18 @@ impl<'src> Thunk<'src> {
           arg: Box::new(arguments.pop().unwrap()),
           name,
         }),
+        (Function::UnaryOpt(function), 1..=2) => {
+          let a = Box::new(arguments.remove(0));
+          let opt_b = match arguments.pop() {
+            Some(value) => Box::new(Some(value)),
+            None => Box::new(None),
+          };
```
kykyi: `remove(0)` to grab from the front of the list, then a `pop` on the last element, happy to hear a prettier way 😅

casey:
Nice, merged!

Issue #1574, opened 2023-04-02
Does recipe with attribute `[unix]` trigger in MacOS?


whinee:
Hello. Just a stupid question: does a recipe with the attribute `[unix]` trigger in a machine running MacOS?

According to Wikipedia ― which, I daresay is a reliable source of information (yes, I'm looking at you, grade school teachers grrrr) ― and other reputable sources, MacOS is a unix OS. And as such, I believe that this should trigger the recipe with an attribute `[unix]`.

Being the poor, Asian girl that I am, I do not have a MacOS to test this on, nor do I know how to read rust code. And so, here I am, asking.

In any case, if this is the case, I think that mentioning it in the documentation would be super helpful.

Thank you so much!

Taywee:
I just ran a test on a MacBook, and `[unix]` recipes indeed do run on MacOS (and Linux as well).

In the future, if you ever need to run something in MacOS as a quick hack or a quick check, [GitHub Actions have MacOS runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners).  It's a pain, but it's not too unreasonable, given how hard it is to test or run things on MacOS without buying Apple hardware, which is prohibitively expensive for very many people.

whinee:
> I just ran a test on a MacBook, and `[unix]` recipes indeed do run on MacOS (and Linux as well).

waaaaa, thank you for testing!!!!

> In the future, if you ever need to run something in MacOS as a quick hack or a quick check, [GitHub Actions have MacOS runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners)

I actually do use that to use that. I don't know why I didn't even tried it in the first place tho...

> It's a pain, but it's not too unreasonable, given how hard it is to test or run things on MacOS without buying Apple hardware, which is prohibitively expensive for very many people.

Oh, absolutely. Running any kind of virtual machine on my 11 year old puppy (thinkpad) is impossible too. :c

Anyways, thank you very much!!!

PR #1619, opened 2023-06-15
Clarify [unix] attribute in documentation

neunenak:
Clarify that the `[unix]` attribute enables a recipe on MacOS as well as other unixes. See https://github.com/casey/just/issues/1574


casey:
Looks good, merged.

PR #1620, opened 2023-06-15
Update README.md

avi-cenna:
Correct small typo in "String" example, and provide additional clarification of how multi-line strings are stripped.


casey:
Nice, good catch! I reworded the note slightly. Merging!

PR #1622, opened 2023-06-17
Use the grep command instead of an alias in bash completions

BojanStipic:
**Problem**:

If a user has a grep alias, such as this:

```
  alias grep='grep --perl-regexp'
```

Running the bash completions script will break with the following output:

```
  grep: conflicting matchers specified
```

This happens because the script tries to use the `-F/--fixed-strings` flag, which conflicts with `--perl-regexp` flag.

**Solution**:

Prefix grep with `\` so that the grep command is always used instead of an alias.


casey:
Nice, merged!

PR #1640, opened 2023-07-11
Working on specifying a `.env` path with a string

groovenectar:
As someone not well-versed in Rust yet, only a fan, this update in this project turns out to be non-trivial

But I did update some `README` files, so I'd like to draft a pull request as I've been pondering this feature for quite some time

I plan to revisit this, but would definitely appreciate any help or pointers

This addresses some posted issues such as https://github.com/casey/just/issues/945

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -60,7 +60,7 @@ assignment    : NAME ':=' expression eol
 export        : 'export' assignment
 
 setting       : 'set' 'allow-duplicate-recipes' boolean?
-              | 'set' 'dotenv-load' boolean?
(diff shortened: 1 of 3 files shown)


casey (changes requested):
Thanks for the PR! `just` is definitely non-trivial to work on, there are a lot of layers (lexer, parser, error handling, documentation, etc) to modify.

I think it would probably be better to add a new setting, `dotenv-path`, which takes a string.

As far as pointers, check out the `tempdir` setting, since takes a string, and `dotenv-path` will probably have a similar implementation.

casey:
This wound up being implemented in #1692.

PR #1663, opened 2023-08-28
Add rust-toolchain file

neunenak:
Adds a `rust-toolchain.toml` file to the repo, pinned to the current rust stable version as of today, which is 1.72.0. 

This codifies exactly what version of rust will definitely build `just`, and prevents the problem that when the current stable version of rust changes, the next time CI runs, the new version of rust might pull in new clippy lints that the codebase won't pass, causing CI failures unrelated to an actual build problem. This does mean that it will be necessary to periodically manually update `rust-toolchain.toml` to be up-to-date with the latest rust stable version, but I think that's a virtue rather than a vice.


casey:
I actually kind of like not pinning the latest stable version of rust in CI. On the one hand CI does break periodically when new lints are introduced, on the other hand, that reminds me to fix the lints and update to the latest version of Rust. I think overall I'm inclined not to do this, unless there's some way to automate a reminder to update Rust.

PR #1692, opened 2023-10-11
Add `dotenv-filename` and `dotenv-path` to in-`justfile` settings

ltfourrier:
## Context

I was trying to load a `.env` file located in another directory from multiple other `justfile`s for a project with a lot of repositories earlier and couldn't find any way to do that in the documentation.

After checking the code, I found the `--dotenv-filename` and `--dotenv-path` arguments that did exactly what I needed to do (introduced in #941), but it was impossible to set those options in the `justfile` directly.

## Changes

This pull request pretty much makes the `--dotenv-filename` and `--dotenv-path` available as settings that can be set using the `set` keyword in a `justfile`.

The implementation of the `load_dotenv` function was modified so that it considers these new settings, on top of the already existing program arguments. The program arguments take precedence over the settings in the `justfile`.

The documentation in `README.md` was updated to reflect those changes.

## Caveats

I found out right before creating this PR that there was already a pull request (#1640) and a few open issues about this topic (#945), and did not take them into account during implementation. There's some very good ideas floating in these, but my PR just exposes the existing program arguments as settings.

I did not update the documentation in `README.中文.md` since I cannot read or write Chinese.


casey:
Looks great! I collapsed the various sections about the dotenv integration into one, since otherwise there would be a bunch of repetition and cross-linking.

PR #1694, opened 2023-10-11
README.md: version number for each attribute (both en & zh)

JoeyTeng:
Also added the missing "attribute" `[private]` in the attribute list, as well as a link to the later section.


casey:
Looks good, thank you!

PR #1699, opened 2023-10-14
Cleanup error reporting code, minor nits

nyurik:
I gave in into my OCD a bit, tried to use `thiserror` but that seems like not a good fit, so instead made error enum a bit more readable and far shorter.  Also, a few tiny nits with trailing commas. Lastly, consolidated command formater.

Note that in a few places I removed `*` in `if let Some(v) = *v`  (v is a `&Option<T>`) -- seems like it is not needed here


casey:
Nice, looks good to me!

PR #1700, opened 2023-10-14
unneeded name qualifiers, spelling, CSS

nyurik:
Just some nits


casey:
I kept the types in the IntoIterator implementations sorted, but otherwise looks good. Merged!

By the way, how did you find these?

nyurik:
Intelij does that automatically

PR #1704, opened 2023-10-17
Feat/issue 1698 global quiet

dharrigan:
Hi,

Here's a first pass at introducing `quiet recipes`. I'm happy to take unboard any thoughts you may have. Not quite sure about the name `set quiet-recipes`, but was struggling to come up with something better (maybe `set globally-quiet`?) :-)

Tests pass (with my additional ones added).

Thank you for your consideration.

#Fixes 1698 

-=david=-


casey:
Thanks for the PR!

Some thoughts:

- Let's call the setting `quiet`, so you can do `set quiet`.
- Let's call the attribute `[noquiet]`
- We could also not have an attribute, and invert the meaning of `@` if `set quiet` is used. So if you `set quiet`, you can do `@recipe:`, which would make that recipe not quiet`. `@` is a bit confusing though, so I'm not convinced either way on this.

dharrigan:
Hi,

Thank you for your helpful feedback. I'll make the changes within the next few days :-)

With regards the inversion of the `@`. I thought about this and I decided on the principle of least surprise (for the user). They are used to `@` to signify that the recipe (or line) should be quiet, so I didn't want to break that by introducing an inverted case. I would rather accrete changes than break expectations. I then thought that by introducing the attribute of `[no-quiet-recipe]` (now `[noquiet]`), they could then choose to make it explicit if they want an already demarked recipe of `@` to become non-quiet when the global attribute is used.

So, no surprises was my thoughts behind that one :-)

-=david=-

dharrigan:
A question...

You've recommended the attribute to be renamed to `noquiet`. However, there is the precedent of `no-exit-message`, with a hyphen between the `no` and the action. Would it be better to be consisent and call the attribute `no-quiet`?

Thank you.

casey:
Ah, yeah, good catch. In that case it should be `no-quiet`.

dharrigan:
Hi!

I was wondering if you have had a chance to consider to merge in this PR? Anything else I can do?

Thank you.

-=david=-

README.md
```diff
@@ -188,10 +188,10 @@ You can also set the shell using command-line arguments. For example, to use Pow
     <tr>
       <td><a href="https://www.gentoo.org">Gentoo Linux</a></td>
       <td><a href="https://wiki.gentoo.org/wiki/Portage">Portage</a></td>
-      <td><a href="https://github.com/gentoo-mirror/guru/tree/master/sys-devel/just">guru/sys-devel/just</a></td>
+      <td><a href="https://github.com/gentoo-mirror/dm9pZCAq/tree/master/sys-devel/just">dm9pZCAq/sys-devel/just</a></td>
```
casey: It looks like there are some unrelated changes, and merge conflicts in README.md.

README.md
```diff
@@ -1243,21 +1244,20 @@ These functions can fail, for example if a path does not have an extension, whic
...
-| `[no-cd]`<sup>1.9.0</sup>           | Don't change directory before executing recipe. |
-| `[no-exit-message]`<sup>1.7.0</sup> | Don't print an error message if recipe fails.   |
-| `[private]`<sup>1.10.0</sup>        | See [Private Recipes](#private-recipes).        |
-| `[unix]`<sup>1.8.0</sup>            | Enable recipe on Unixes. (Includes MacOS).      |
-| `[windows]`<sup>1.8.0</sup>         | Enable recipe on Windows.                       |
+| Name                                | Description                                                    |
+| ----------------------------------- | -----------------------------------------------                |
+| `[no-cd]`<sup>1.9.0</sup>           | Don't change directory before executing recipe.                |
+| `[no-exit-message]`<sup>1.7.0</sup> | Don't print an error message if recipe fails.                  |
+| `[no-quiet]`<sup>?</sup>            | Override globally quiet recipes and always echo out the recipe |
```
casey: We should also add `[quiet]`.

README.md
```diff
@@ -2221,6 +2220,33 @@ goodbye
 # all done!
 ```
 
+In addition, an entire Justfile can be marked as being globally quiet by
+setting `set quiet` in the Justfile. For example:
```
```suggestion
All recipes in a Justfile can be made quiet with `set quiet`:
```

README.md
```diff
@@ -2221,6 +2220,33 @@ goodbye
...
+
+foo:
+  echo "This is quiet"
+
+@foo2:
+  echo "This is also quiet"
+```
+
+If the recipe has the attribute of `[no-quiet]`, then the setting is
+ignored: For example:
```
```suggestion
If a recipe has`[no-quiet]`, then the setting is ignored:
```

README.md
```diff
@@ -2338,6 +2364,7 @@ import 'foo/bar.just'
 
 a: b
   @echo A
+
```
```suggestion
```

src/attribute.rs
```diff
@@ -12,6 +12,7 @@ pub(crate) enum Attribute {
   NoCd,
   NoExitMessage,
   Private,
+  NoQuiet,
```
casey, line 15: These enum variants should be sorted.

src/attribute.rs
```diff
@@ -31,7 +32,12 @@ mod tests {
   use super::*;
 
   #[test]
-  fn to_str() {
+  fn no_exit_message_to_str() {
```
casey:
```suggestion
  fn to_str() {
```

src/attribute.rs
```diff
@@ -31,7 +32,12 @@ mod tests {
...
     assert_eq!(Attribute::NoExitMessage.to_str(), "no-exit-message");
   }
+
+  #[test]
```
casey: We can remove this test. The other test is there just to make sure that we're serializing attribute names in kebab-case, but we don't need to test every variant.

src/recipe.rs
```diff
@@ -103,6 +103,10 @@ impl<'src, D> Recipe<'src, D> {
     !self.attributes.contains(&Attribute::NoExitMessage)
   }
 
+  fn no_quiet(&self) -> bool {
+    !self.attributes.contains(&Attribute::NoQuiet)
```
casey: Should this be inverted, i.e.:

```suggestion
    self.attributes.contains(&Attribute::NoQuiet)
```

src/recipe.rs
```diff
@@ -150,7 +154,10 @@ impl<'src, D> Recipe<'src, D> {
       }
       let mut evaluated = String::new();
       let mut continued = false;
-      let quiet_command = lines.peek().map_or(false, |line| line.is_quiet());
+
+      let quiet_command = self.no_quiet() && context.settings.quiet
```
casey: This is used to determine if we should strip a leading `@` from the line, so we should only do so if it starts with an `@`:

```suggestion
      let quiet_line = lines.peek().map_or(false, |line| line.is_quiet());
```

src/recipe.rs
```diff
@@ -193,7 +200,9 @@ impl<'src, D> Recipe<'src, D> {
 
       if config.dry_run
         || config.verbosity.loquacious()
-        || !((quiet_command ^ self.quiet) || config.verbosity.quiet())
+        || !((quiet_command ^ self.quiet)
+          || (context.settings.quiet && self.no_quiet())
+          || config.verbosity.quiet())
```
casey: I think this should be something like:

```suggestion
        || !((quiet_line ^ self.quiet) || (context.settings.quiet && !self.no_quiet()) || config.verbosity.quiet())
```

casey (changes requested):
Sorry I was slow getting to this! Check out my review comments.

README.md
dharrigan: Fixed

README.md
dharrigan: There isn't a `[quiet]` attribute, just `[no-quiet]` which overrides the `set quiet` if that's been set.

README.md
dharrigan: done

README.md
dharrigan: done

README.md
dharrigan: done

src/attribute.rs
dharrigan: done

src/attribute.rs
dharrigan: done

src/recipe.rs
dharrigan: done

src/recipe.rs
dharrigan: done

src/recipe.rs
dharrigan: done

dharrigan:
Hi!

All changes asked for applied. When you can spare the time, have a looksee and see what you think :-)

Thank you.

-=david=-

casey (changes requested):
Sorry for letting this sit! This looks good, it just needs integration tests, which can go in `tests/quiet.rs`. The tests in that file are using the `test!` macro, which I try to avoid these days, so for examples of how to write tests, check out `tests/modules.rs`. In particular, all the interactions with the quiet setting, the noquiet attribute, quiet lines, and the `@` prefix to both recipes in lines, should be tested.

dharrigan:
Hi,

Additional integration tests done. I hope this is looking better.

-=david=-

casey (approved):
Looks good to me!

casey:
Thanks for the PR! I tweaked it a little bit, simplifying the tests and removing an `if` statement that would never be executed.

This is easily the worst conditional I've ever seen, amazingly using `||`, `!`, `^`, and `&&`:

```rust
      if config.dry_run
        || config.verbosity.loquacious()
        || !((quiet_line ^ self.quiet)
          || (context.settings.quiet && !self.no_quiet())
          || config.verbosity.quiet())
      {
```

But so be it! Functionality for making things quiet or not quiet has just evolved organically over time, so I guess that's just how it is 😂

PR #1708, opened 2023-10-20
Use XDG_CACHE_HOME if /tmp is mounted with noexec

### LINKED ISSUE #1611: Error: Recipe `test` with shebang `#!/usr/bin/env bash` execution error: Permission denied (os error 13)
Hello :)

I'm using just to run command on my server and I created a recipe with bash commands like this : 

Justfile
```
test:
  #!/usr/bin/env bash
  echo "This is a test"
```

Error : 
`error: Recipe `test` with shebang `#!/usr/bin/env bash` execution error: Permission denied (os error 13)`

Do I missed something ? Is someone already had this issue ?

I's a fresh new machine in Ubutu 22.04.

Regards,

aidanharris:
Fixes #1611

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,6 +157,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "dirs"
+version = "5.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


ben31121987 (approved):

casey:
I just merged #2067, which always uses the cache dir if available. We'll see if it causes problems for anyone, if so, we can explorer checking for `noexec`.

Issue #1703, opened 2023-10-17
auto completion was broken in 1.15.0


azone:
Please see the video below:

macOS version: Sonoma 14.0
Shell: fish 3.6.1

https://github.com/casey/just/assets/164407/f36daa52-b528-41d4-a597-7cf351457e25

casey:
Ahh, that sucks. I don't really know anything about fish, and the completion scripts are auto-generated by clap, so someone else will have to take the lead in debugging this.

sebdvd:
Hi, I have the same issue, but it's not really because of fish, but because of the `sed` command used in the fish autocomplete file:

```shell
just --list 2> /dev/null | sed -e '1d; s/^\s*\([^[:space:]]*\)[^#]*$/\1/' -e 's/^\s*\([^[:space:]]*\)[^#]*# \(.*\)$/\1\t\2/'
```

The second sed arg isn't working properly with the native `macOS` sed command, for example consider this justfile:

```justfile
# Commands list
help:
  @just --list
```

The previous command will output this:

```
	Commands list
```

You can see that the recipe is missing.

But now the same command with `gsed` will output the expected result (and fix the autocompletion):

```
help	Commands list
```

l4zygreed:
@azone can you try this PR? 😺

azone:
@l4zygreed Thanks a lot,  but it seems still not working for me...

sebdvd:
@l4zygreed it works for me 🤷 (currently tested on macOS 14.1 and fish 3.6.1)

l4zygreed:
@azone have you replaced autocompletion file with new one? 🤔 
I placed mine in `.config/fish/completions/just.fish`
did you test it from PR branch? it's not merged yet 🤡 
We have 3 confirmations that it's working, so probably you missed something P:
(also you may need to reload the shell after changing complition file)

casey:
Fixed by @l4zygreed in #1710. Thank you for the PR!

azone:
> @azone have you replaced autocompletion file with new one? 🤔 I placed mine in `.config/fish/completions/just.fish` did you test it from PR branch? it's not merged yet 🤡 We have 3 confirmations that it's working, so probably you missed something P: (also you may need to reload the shell after changing complition file)

Oh sorry~, I forgot to switch the branch. It works now!

PR #1710, opened 2023-10-25
fish_completions: fix completion for macos

l4zygreed:
tried to fix broken completion for macos #1703
edited `src/completions.rs` file and regenerated completions via `bin/generate-completions` script

tested on macos sonoma and archlinux machine, works the same
replaced `sed` with `awk` to have more flexible formatting on arguments (and because macos `sed` have differences)
example image


casey:
This looks good to me! I'm not a fish user, so if anyone who uses fish wants to try this out before I merge, please do! Otherwise I'll just let it sit for a little bit and merge it.

sebdvd:
I can confirm it works on macos 14.1 with fish 3.6.1

sneeu:
This change is working for me on macOS 13.5 & fish 3.6.1.

PR #1713, opened 2023-10-26
feat: add `semver_matches` utility function

### LINKED ISSUE #1712: Add `semver_lt`/`semver_gt`/`semver_eq`/... functions
Thanks for the awesome tool -- `just` has been a huge part of my daily workflow and I take it to "just" about every project I start now. 

One thing I've often run into is the need to parse & check versions of dependencies cross-language/toolchain against each other, and checking/ensuring a certain version of semver often comes up. 

For exact matches this is relatively simple, but it might be nice to do be able to do this a bit more semantically as a utility within `just`.

A somewhat contrived example of what I'd like to be able to do is:

```just
dep := env_var_or_default("DEP", "dep")
dep_version := `dep --version` # let's pretend that this prints the semver version
dep_version_supported := if semver_gte(dep_version, "0.2.0") { "supported" } else { "not supported" }

check:
  echo "your dep version of dep ({{dep_version}}) is {{dep_version_supported}}"
```
Does this seem like something that makes sense as a part of `just`?

t3hmrman:
While coordinating different programs using `just`, it's often necessary to compare the *versions* of different binaries or utilitiies. Assuming that the utility in question providers a `--version` switch and can reasonably print out a version, being able to match semver of the utility to an expectation is very useful.

This commit adds a `semver_matches` utility which utilizes the `semver` crate (https://crates.io/crates/semver) to provide matching functionality to `Justfile`s.

Resolves #1712

### DIFF
--- a/README.md
+++ b/README.md
@@ -1241,6 +1241,10 @@ These functions can fail, for example if a path does not have an extension, whic
 - `sha256_file(path)` - Return the SHA-256 hash of the file at `path` as a hexadecimal string.
 - `uuid()` - Return a randomly generated UUID.
 
+#### Semantic Versions
+
+- `semver_matches(version, requirement)`<sup>master</sup> - Check whether a [semantic version](https://semver.org) `version`, e.g., `"0.1.0"` matches requirement `requirement`, e.g., `">=0.1.0"`, returning `"true"` if so and `"false"` otherwise. 
+
 ### Recipe Attributes
 
 Recipes may be annotated with attributes that change their behavior.
--- a/README.md
+++ b/README.md
@@ -1243,7 +1243,7 @@ These functions can fail, for example if a path does not have an extension, whic
 
 #### Semantic Versions
 
-- `semver_matches(version, requirement)`<sup>master</sup> - Check whether a [semantic version](https://semver.org) `version`, e.g., `"0.1.0"` matches requirement `requirement`, e.g., `">=0.1.0"`, returning `"true"` if so and `"false"` otherwise. 
+- `semver_matches(version, requirement)`<sup>master</sup> - Check whether a [semantic `version`](https://semver.org), e.g., `"0.1.0"` matches a `requirement`, e.g., `">=0.1.0"`, returning `"true"` if so and `"false"` otherwise.
 
 ### Recipe Attributes
 
--- a/src/function.rs
+++ b/src/function.rs
@@ -424,7 +424,7 @@ fn semver_matches(
   Ok(
     requirement
       .parse::<VersionReq>()
-      .map_err(|err| format!("invalid semver version requirement: {err}"))?
+      .map_err(|err| format!("invalid semver requirement: {err}"))?
       .matches(
         &version
           .parse::<Version>()
--- a/src/function.rs
+++ b/src/function.rs
@@ -4,6 +4,7 @@ use {
     ToKebabCase, ToLowerCamelCase, ToShoutyKebabCase, ToShoutySnakeCase, ToSnakeCase, ToTitleCase,
     ToUpperCamelCase,
   },
+  semver::{Version, VersionReq},
   Function::*,
 };
 
@@ -46,6 +47,7 @@ pub(crate) fn get(name: &str) -> Option<Function> {
     "quote" => Unary(quote),
     "replace" => Ternary(replace),
     "replace_regex" => Ternary(replace_regex),
+    "semver_matches" => Binary(semver_matches),
     "sha256" => Unary(sha256),
     "sha256_file" => Unary(sha256_file),
     "shoutykebabcase" => Unary(shoutykebabcase),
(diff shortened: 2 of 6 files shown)


t3hmrman:
Hey @casey got some code up to implement the new utility -- please let me know if I'm missing anything, or if you'd like more test cases!

src/function.rs
```diff
@@ -1,9 +1,12 @@
+use semver::VersionReq;
+
 use {
   super::*,
   heck::{
     ToKebabCase, ToLowerCamelCase, ToShoutyKebabCase, ToShoutySnakeCase, ToSnakeCase, ToTitleCase,
     ToUpperCamelCase,
   },
+  semver::Version,
```
casey: You can put everything into the same use block:

```suggestion
use {
  super::*,
  heck::{
    ToKebabCase, ToLowerCamelCase, ToShoutyKebabCase, ToShoutySnakeCase, ToSnakeCase, ToTitleCase,
    ToUpperCamelCase,
  },
  semver::{Version, VersionReq},
```

src/function.rs
```diff
@@ -411,3 +415,14 @@ fn without_extension(_context: &FunctionContext, path: &str) -> Result<String, S
 
   Ok(parent.join(file_stem).to_string())
 }
+
+/// Check whether a string processes properly as semver (e.x. "0.1.0")
+/// and matches a given semver requirement (e.x. ">=0.1.0")
+fn semver_matches(_context: &FunctionContext, v: &str, req: &str) -> Result<String, String> {
+  Ok(
+    VersionReq::parse(req.as_ref())
+      .map_err(|err| err.to_string())?
```
casey: I'm curious what the error message looks like. Sometimes libraries produce errors which aren't super helpful when they're bubbled up to the end user. If it's ambiguous or doesn't give enough information, add context with something like:

```suggestion
      .map_err(|err| format!{"invalid semver: {err}")?
```

src/function.rs
```diff
@@ -411,3 +415,14 @@ fn without_extension(_context: &FunctionContext, path: &str) -> Result<String, S
 
   Ok(parent.join(file_stem).to_string())
 }
+
+/// Check whether a string processes properly as semver (e.x. "0.1.0")
+/// and matches a given semver requirement (e.x. ">=0.1.0")
+fn semver_matches(_context: &FunctionContext, v: &str, req: &str) -> Result<String, String> {
+  Ok(
+    VersionReq::parse(req.as_ref())
```
casey: I like using parse with a turbofish:

```suggestion
    req.parse::<VersionReq>()
```

src/function.rs
```diff
@@ -411,3 +415,14 @@ fn without_extension(_context: &FunctionContext, path: &str) -> Result<String, S
...
+      .map_err(|err| err.to_string())?
+      .matches(&Version::parse(v.as_ref()).map_err(|err| err.to_string())?)
```
casey: Not sure if it works with the type inference here, but you might be able to do:

```suggestion
      .matches(&v.parse().map_err(|err| err.to_string())?)
```

The same comment about the error applies here. If it's ambiguous on it's own, give it some context:

tests/functions.rs
```diff
@@ -412,6 +412,98 @@ test! {
     stderr: "echo Bar\n",
 }
 
+test! {
```
casey: These tests are great, but I think we don't need all of them. In general, when using a library, I try to avoid writing tests which test the library itself, instead of testing how it's being used. I just kind of assume the library works, and only test the way it's being used. Since the semver library is handling all the comparisons, I think we can get away with just one test with a matching semver, one with a non-matching semver, and one with an error when the version fails to parse, and one where the req fails to parse.

src/function.rs
```diff
@@ -411,3 +415,14 @@ fn without_extension(_context: &FunctionContext, path: &str) -> Result<String, S
 
   Ok(parent.join(file_stem).to_string())
 }
+
+/// Check whether a string processes properly as semver (e.x. "0.1.0")
+/// and matches a given semver requirement (e.x. ">=0.1.0")
+fn semver_matches(_context: &FunctionContext, v: &str, req: &str) -> Result<String, String> {
```
casey: Maybe a little more explicit:

```suggestion
fn semver_matches(_context: &FunctionContext, version: &str, requirement: &str) -> Result<String, String> {
```

README.md
```diff
@@ -1241,6 +1241,12 @@ These functions can fail, for example if a path does not have an extension, whic
 - `sha256_file(path)` - Return the SHA-256 hash of the file at `path` as a hexadecimal string.
 - `uuid()` - Return a randomly generated UUID.
 
+#### Semantic Versions
+
+- `semver_matches(v, req)` - Check whether a [semantic version][semver] `v` (e.x. `"0.1.0"`) matches a version requirement `req` (e.x. `">=0.1.0"`), returning `"true"` if so and `"false"` otherwise.
```
casey: Rewording a little bit here:

```suggestion
- `semver_matches(version, requirement)`<sup>master</sup> - Check whether a [semantic version](https://semver.org) `version`, e.g., `"0.1.0"` matches requirement `requirement`, e.g., `">=0.1.0"`, returning `"true"` if so and `"false"` otherwise. 
```

I often forget, but I try to put `<sup>VERSION</sup>` in the docs, so people know what version of Just a feature was added, and `<sup>master</sup>` if it's a feature which hasn't made it into a release yet.

casey (approved):
Looks good! Check out the review comments.

src/function.rs
t3hmrman: thanks for this suggestion -- combined all the imports!

tests/functions.rs
t3hmrman: Yeah this is pretty reasonable -- we can simply trust most of what `semver` has here, I'll do a matching `gte` which I think I think is probably the most common

t3hmrman: OK so for this I left in the gte test!

t3hmrman:
Thanks for the thorough review @casey , I made the changes and squashed to get one commit (I'm a fan of single non-merge commit history).

casey:
Nice, merged! Thank you for the PR!

t3hmrman:
Thanks for the help and thorough reviews! :)

PR #1717, opened 2023-11-06
fix typo in quiet recipe docs.

barraponto:


casey:
Nice catch, thank you!

Issue #1711, opened 2023-10-25
Feature request – confirmation option


lud-wj:
It would be nice if we could flag a command to ask for confirmation. 

For instance we use a lot of commands like `reset-db-test` and `reset-db-dev`, but the latter would mean losing all data manually entered by the developer on the project. We can do the confirmation manually but it could be a nice feature.

What do you think?

casey:
You can use `--dry-run` to preview the commands that would be executed, without actually running them, and then run just again without the `--dry-run` flag to actually run them. Does that work?

lud-wj:
Not really because I know what the command does. It's just that I run the wrong one by mistake. But I dont want to dry-run all the time.

For now I have this recipe:

```just
_confirm:
  #!/usr/bin/env bash
  read -p "Are you sure? " -n 1 -r
  if [[ ! $REPLY =~ ^[Yy]$ ]]
  then
      echo # newline
      [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
  else
      echo # newline
  fi
  ```
  
  And for instance I use it on the "reset dev" but not on "reset-test":
  
```just
# Rollbacks and re run all migrations for the test database
reset-db-test:
  MIX_ENV=test mix ecto.drop
  MIX_ENV=test mix ecto.create --quiet
  MIX_ENV=test mix ecto.migrate --log-migrations-sql

# Rollbacks and re run all migrations for the development database
reset-db-dev: _confirm
  mix ecto.drop
  mix ecto.create --quiet
  mix ecto.migrate --log-migrations-sql
  ```
  
It works, but it could be nice if the recipe was built-in, with nicer propmpt, and the guarantee that answering "no" will terminate everything (I don't know bash, I am not sure those `return` and `exit` will work 100% of the time).

But that's just a suggestion, for now what I have works!

casey:
Gotcha! I'd be okay with this, especially if the implementation is simple.

Hwatwasthat:
I've had a look at this, I wondered about adding an attribute that would require confirmation for just the annotated recipe. I've got a running example but I don't know if it's handled in a way you'd want in the project (it reads from stdio right now, which I can't see being used directly elsewhere in the project). I can make a pull request and we could go from there?

casey:
I wouldn't be opposed to an attribute which required confirmation. I think this covers @lud-wj's original feature request, @lud-wj, what do you think?

lud-wj:
Sure! I did not know about attributes, but that's great :)

[...]

PR #1723, opened 2023-11-12
Add a confirm attribute for recipes

Hwatwasthat:
This would add a "Confirm" attribute that requests confirmation before running a recipe. Runs prior to evaluation, and defaults to not running. Potential solution to issue [#1711](https://github.com/casey/just/issues/1711).


.gitignore
```diff
@@ -13,3 +13,4 @@
 /test-utilities/Cargo.lock
 /test-utilities/target
 /tmp
+.vscode/
```
casey: This should be:

```suggestion
/.vscode
```

Also, the lines in this file should be sorted.

src/attribute.rs
```diff
@@ -13,6 +13,7 @@ pub(crate) enum Attribute {
   Private,
   Unix,
   Windows,
+  Confirm
```
casey: These enum variants should be sorted.

src/recipe.rs
```diff
@@ -63,6 +63,16 @@ impl<'src, D> Recipe<'src, D> {
     self.name.line
   }
 
+  pub(crate) fn confirm(&self) -> bool {
```
casey: This function should be fallible, since `read_line` can fail, so it should return `RunResult<bool>`

src/recipe.rs
```diff
@@ -159,6 +175,8 @@ impl<'src, D> Recipe<'src, D> {
         }
       }
 
+      
+
       if comment_line {
```
casey: Extra whitespace:

```suggestion
      if comment_line {
```

src/recipe.rs
```diff
@@ -63,6 +63,16 @@ impl<'src, D> Recipe<'src, D> {
     self.name.line
   }
 
+  pub(crate) fn confirm(&self) -> bool {
+    if self.attributes.contains(&Attribute::Confirm) {
+      eprint!("Confirm running {} (y/N): ", self.name);
+      let mut buf = String::new();
+      std::io::stdin().read_line(&mut buf).is_ok() && buf.trim_end().eq_ignore_ascii_case("y")
```
```suggestion
      let mut line = String::new();
      std::io::stdin().read_line(&mut line)?;
      let line = line.trim().to_lowercase();
      line == "y" || line == "yes"
```

casey (changes requested):
I just added some comments, and I thought of something else: If a recipe doesn't run, any recipe in the current invocation which depends on that recipe should also not run.

Hwatwasthat:
> I just added some comments, and I thought of something else: If a recipe doesn't run, any recipe in the current invocation which depends on that recipe should also not run.

I've added in the confirmation during the running of dependencies. Now, if a recipe that requires confirmation is run as a dependency and is not confirmed then it will return an error of type `DependencyNotConfirmed`, which I could only find a way to insert the name of the dependency being run, I can't find an easy way to inject the parents name without adding in a lot more convolution.

The confirmation is carried out at the start of the `Justfile::run_recipe` function, as I couldn't see another place to put it that would prevent anything running and would allow the returning of an error rather than the ok of a single recipe that is being run (as that would stop any later recipes that were from running as I understand it, which is not the intended use of confirmation (again, as I understand it).

Hwatwasthat:
Apologies, on testing a bit I realised that I've added in a double confirm. I'll figure out the best way to resolve this and then update the pull.

casey:
Aside from the double confirm, that all sounds good. I think returning an error if a recipe isn't confirmed is sensible. I think that if we get to a recipe that needs confirmation, and it isn't confirmed, we can abort the whole run, and not run any other recipes. Although it is true that we could run any recipes which aren't dependencies of the unconfirmed recipe, not running them doesn't seem like a big deal.

Also, this will require a couple of tests: One that a recipe that requires and receives confirmation runs, and one that a recipe which requires confirmation but doesn't receive it does not run.

casey:
Also, if you feel like adding a `--yes` flag that confirms all recipes (per this comment https://github.com/casey/just/issues/1711#issuecomment-1807732753) which seems useful, go for it, otherwise it can be left to another PR.

Hwatwasthat:
> Aside from the double confirm, that all sounds good. I think returning an error if a recipe isn't confirmed is sensible. I think that if we get to a recipe that needs confirmation, and it isn't confirmed, we can abort the whole run, and not run any other recipes. Although it is true that we could run any recipes which aren't dependencies of the unconfirmed recipe, not running them doesn't seem like a big deal.
> 
> Also, this will require a couple of tests: One that a recipe that requires and receives confirmation runs, and one that a recipe which requires confirmation but doesn't receive it does not run.

Ok, that makes things easier then, I've got a version working that does that so I'll just plow on with it. Tests to come with that in the next commit.

Hwatwasthat:
> Also, if you feel like adding a `--yes` flag that confirms all recipes (per this comment [#1711 (comment)](https://github.com/casey/just/issues/1711#issuecomment-1807732753)) which seems useful, go for it, otherwise it can be left to another PR.

I was just wondering on the name. yes is a bit ambiguous, maybe `--force-confirm` (just to make it clear), I feel like `--force` or `--no-confirm` doesn't work, as force is ambiguous and no confirm reads to me as the inverse (automatically denying confirm stuff). I have a flag working, right now it's `--yes`.

casey (approved):
I tweaked a few error messages, as well as printed the error confirmation message to stderr, which is also line-buffered, so no need to flush.

Thanks for the PR!

PR #1726, opened 2023-11-14
Clarify aliases documentation

### LINKED ISSUE #1725: Specifying dependency by alias causes error
Using `just` 1.16.0 Linux x86_64 build downloaded from https://github.com/casey/just/releases/tag/1.16.0

Minimal test justfile:
```
foo: a
	echo 'Foo'

actual-recipe:
	echo 'this recipe has an alias'
alias a := actual-recipe
```

Running `just` just errors with -
```
$ just
error: Recipe `foo` has unknown dependency `a`
  |
1 | foo: a
  |      ^
```

Same error still occurs when putting the aliased recipe and/or its alias before the recipe that depends on it.

Workaround: instead of using an alias, define it as another recipe:
```
foo: a
	echo 'Foo'

actual-recipe:
	echo 'this recipe has an alias'
a: actual-recipe
```

Is it possible to make this work with alias?

laniakea64:
Resolves https://github.com/casey/just/issues/1725 by documenting the scope of the alternative names defined with `alias`.  Also updated the output of the example `just b` command to match what current `just` would do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -634,7 +634,7 @@ $ just --list --list-heading ''
 
 ### Aliases
 
-Aliases allow using alternative names for recipes when invoking them on the command line:
+Aliases allow recipes to be invoked on the command line with alternative names:
 
 ```just
 alias b := build
--- a/README.md
+++ b/README.md
@@ -634,7 +634,7 @@ $ just --list --list-heading ''
 
 ### Aliases
 
-Aliases allow recipes to be invoked with alternative names:
+Aliases allow using alternative names for recipes when invoking them on the command line:
 
 ```just
 alias b := build
@@ -645,7 +645,6 @@ build:
 
 ```sh
 $ just b
-build
 echo 'Building!'
 Building!
 ```


README.md
```diff
@@ -634,7 +634,7 @@ $ just --list --list-heading ''
 
 ### Aliases
 
-Aliases allow recipes to be invoked with alternative names:
+Aliases allow using alternative names for recipes when invoking them at the command line:
```
casey: How about this:

```suggestion
Aliases allow recipes to be invoked on the command line with alternative names:
```

casey:
Added a comment, also, the branch is out of date. You might want to set the option that allows maintainers to edit pull requests, so I can do things like update the branch.

README.md
laniakea64: I tried several variations on that wording before submitting this PR (including replacing "with" with "by" or "under", and replacing "recipes to be invoked" with "invoking recipes"), none of them read as clearly to me.  That maybe problem on my end, so do you think most `just` users would find this wording clearer?

laniakea64:
> You might want to set the option that allows maintainers to edit pull requests, so I can do things like update the branch.

If that's enabled and you decide to make more significant changes than simply updating the branch, would the squash merge commit give you the co-author credit you'd deserve?

casey:
> If that's enabled and you decide to make more significant changes than simply updating the branch, would the squash merge commit give you the co-author credit you'd deserve?

It would, since it adds a `Co-authors: Casey Rodarmor <casey@rodarmor.com>` line to the commit message, but I usually just delete that, since it adds noise to the commit history, and I don't mind if I don't get credit. Plus I usually only make minor changes to PRs like that.

README.md
casey: I'm not sure! For this change, I think the important thing is to change `be invoked` to `be invokeed on the command line`, and I don't mind the previous wording.

laniakea64: Thanks for the answer.  Let's try this, if people ask for clarification it can be changed later, if they don't then this way is better because it's more concise.

casey:
Nice, merged!

PR #1607, opened 2023-05-23
Update github search link to github's suggestion

jbaber:
The suggested change seems to find justfiles.

PR #1730, opened 2023-11-18
Fix unrecognized qualifier in zh doc

oluceps:
Corresponding change: https://github.com/casey/just/pull/1607

```
Unrecognized qualifier. Looking for a filename? Try using the path qualifier
filename:justfile
Suggestion: path:**/justfile
```


casey:
Nice, merging!

Issue #1736, opened 2023-11-21
Make default recipe selection sane when using `!include`


casey:
Currently, we select the recipe in all source files with the lowest line number as the default recipe. This is clearly crazy. We should probably select the first recipe in the root source file, and not select a default recipe if there are no recipes in the root source file.

If people want to have a default recipe in an `!include`, we should add a `[default]` attribute.

casey:
Fixed in #1740.

PR #1739, opened 2023-11-22
Use correct recipe as default when running a justfile with includes

neunenak:
Addresses https://github.com/casey/just/issues/1736


casey:
Tokens now include a relative path to the file the token is from, so I did this in #1740 by checking if the name of the recipe is from the root file.

PR #1752, opened 2023-12-03
Fix typo in README.md

laniakea64:
https://just.systems/man/en/ table of contents on the left currently has an entry "1.6.11. Requiring Confirmation for Recipesmaster".  This looks like a typo, looks like it was intended to be a sub-section of the "Recipe Attributes" page?


casey:
Nice, looks good!

PR #1758, opened 2023-12-11
Pass `--unstable` and `--color always` to fzf's default preview command

Qeole:
Pass `--unstable` to the preview command for the chooser, so that recipe previews can work even when the justfile relies on unstable features, such as `!include` directives.

Note that we're not taking the user by surprise by enabling unstable features in the preview: when the justfile includes unstable features, the user will need to pass `--unstable` explicitely to call `just --choose --unstable` already, or `just` will exit before launching the chooser.

Also tell `just`, in a second commit, to use colours for recipe previews.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -4,7 +4,7 @@ use {
 };
 
 // These three strings should be kept in sync:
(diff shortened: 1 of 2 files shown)


casey:
Nice, good change! Merging now.

Qeole:
Thanks!

PR #1759, opened 2023-12-12
Chooser: Pass justfile path to preview command for default chooser

### LINKED ISSUE #1638: --choose side pane not using correct just file
I have a non-default just file named eg `justfoo`, with a shebang line like

```bash
#!/usr/bin/env just -f
```

When I run `./justfoo --choose`, the left pane lists justfoo's recipes, and they run correctly when selected. But the right pane seems to look for the default `justfile` instead, and shows error messages like "no justfile found" or "Justfile does not contain recipe ...". I expected it to show the recipes from `justfoo`.

Is this a bug ?

Qeole:
The default chooser is fzf, and we pass it a command to generate the preview with `just --show` from the recipe name. This has been working well as long as we rely on the default justfile available in the directory, given that the preview command can find it too. However, passing `--chose` alongside `--justfile` to select a specific file results in the preview command not being able to find the right justfile to process.

To address this issue, we turn the const string defining the preview command into a function that takes the internal representation of the justfile as an argument, and updates the preview command with the right file.

Fixes: https://github.com/casey/just/issues/1638

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,13 +3,21 @@ use {
   clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, ArgSettings},
 };
 
-// These three strings should be kept in sync:
-pub(crate) const CHOOSER_DEFAULT: &str =
-  "fzf --multi --preview 'just --unstable --color always --show {}'";
+// These two strings, and the two in function chooser_default below should be kept in sync:
 pub(crate) const CHOOSER_ENVIRONMENT_KEY: &str = "JUST_CHOOSER";
 pub(crate) const CHOOSE_HELP: &str = "Select one or more recipes to run using a binary. If \
                                       `--chooser` is not passed the chooser defaults to the value \
                                       of $JUST_CHOOSER, falling back to `fzf`";
+
+// Return the string for the default chooser. This is a function and not a const, because we want
+// to edit the preview command and pass it the right justfile path for the target.
+pub(crate) fn chooser_default(justfile: &Path) -> OsString {
+  let mut chooser = OsString::new();
+  chooser.push("fzf --multi --preview 'just --unstable --color always --justfile ");
+  chooser.push(justfile);
+  chooser.push(" --show {}'");
+  chooser
+}
 
 #[derive(Debug, PartialEq)]
 #[allow(clippy::struct_excessive_bools)]
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,19 +3,17 @@ use {
   clap::{App, AppSettings, Arg, ArgGroup, ArgMatches, ArgSettings},
 };
 
-// These two strings, and the two in function chooser_default below should be kept in sync:
 pub(crate) const CHOOSER_ENVIRONMENT_KEY: &str = "JUST_CHOOSER";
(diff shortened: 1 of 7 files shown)


src/subcommand.rs
```diff
@@ -216,7 +216,7 @@ impl Subcommand {
     let chooser = chooser
       .map(OsString::from)
       .or_else(|| env::var_os(config::CHOOSER_ENVIRONMENT_KEY))
-      .unwrap_or_else(|| OsString::from(config::CHOOSER_DEFAULT));
+      .unwrap_or_else(|| config::chooser_default(justfile));
```
casey: I think it's more natural to pass in `Search::justfile` here:

```suggestion
      .unwrap_or_else(|| config::chooser_default(&search.justfile));
```

src/config.rs
```diff
@@ -3,14 +3,27 @@ use {
...
-  "fzf --multi --preview 'just --unstable --color always --show {}'";
+// These two strings, and the two in function chooser_default below should be kept in sync:
 pub(crate) const CHOOSER_ENVIRONMENT_KEY: &str = "JUST_CHOOSER";
 pub(crate) const CHOOSE_HELP: &str = "Select one or more recipes to run using a binary. If \
                                       `--chooser` is not passed the chooser defaults to the value \
                                       of $JUST_CHOOSER, falling back to `fzf`";
 
+// Return the string for the default chooser. This is a function and not a const, because we want
+// to edit the preview command and pass it the right justfile path for the target.
+pub(crate) fn chooser_default(justfile: &Justfile) -> OsString {
```
```suggestion
pub(crate) fn chooser_default(justfile: &Path) -> OsString {
```

casey (changes requested):
Left a couple small comments, otherwise looks good.

src/config.rs
qmonnet: Right, it also makes the function simpler, no need to check the Option value.

qmonnet: sorry, wrong GH account - I'm deleting this

Qeole: Right, it also makes the function simpler, no need to check the Option value.

src/subcommand.rs
Qeole: Thanks, updated.

Qeole:
I'm getting a failure on the CI test this time, `justfile` has been replaced with `/tmp/just-test-tempdirWEynpr/justfile`. I can probably use a wildcard to address this in the test.

```
---- choose::invoke_error_function stdout ----
thread 'choose::invoke_error_function' panicked at tests/test.rs:244:9:
Stderr regex mismatch:
"error: Chooser `/ -cu fzf --multi --preview 'just --unstable --color always --justfile /tmp/just-test-tempdirWEynpr/justfile --show {}'` invocation failed: Permission denied (os error 13)\n"
!~=
/Regex("^error: Chooser `/ -cu fzf --multi --preview 'just --unstable --color always --justfile justfile --show \\{\\}'` invocation failed: .*\\n$")/
```

casey (approved):
Nice, LGTM! I added quotes around the path in the chooser string, since it might wind up having spaces.

PR #1782, opened 2023-12-27
Add modules

### LINKED ISSUE #383: Modules and subcommands

casey:
Fixes #383.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2409,6 +2409,8 @@ Currently, recipes in submodules run with the same working directory as the
 root `justfile`, and the `justfile()` and `justfile_directory()` functions
 return the path to the root `justfile` and its parent directory.
 
+See the [module stabilization tracking issue](https://github.com/casey/just/issues/929) for more information.
+
 ### Hiding `justfile`s
 
 `just` looks for `justfile`s named `justfile` and `.justfile`, which can be used to keep a `justfile` hidden.
--- a/README.md
+++ b/README.md
@@ -2397,6 +2397,9 @@ $ just bar b
 B
 ```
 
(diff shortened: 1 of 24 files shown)


vtvz:
God how I love this feature. Now I wait for `justfile()` to work with modules

casey:
Glad you like it! I opened #1850 to track adding functions which are like `justfile()` but return the path to the current module.

Issue #1781, opened 2023-12-24
SIGHUP seems to not be propagated to inner child processes


davoclavo:
Howdy! First of all, I am a recent `just` convert and I am loving it. Thanks for all the work into this amazing tool.

I have been using a mix of `just` and [`zellij`](https://github.com/zellij-org/zellij) to handle some of my workflows, and noticed that when I close a `zellij` pane running a `just` recipe it leaves the process alive.

Summary: It seems that a SIGHUP signal is being sent to `just`, but that signal is not being propagated to the inner child process, which causes the whole process tree to be left alive.

Here is a basic way to reproduce this issue using a simple bash script:

`just --version`
```
just 1.16.0
```

`sighup.sh`
```bash
#!/bin/bash

# Define a function to handle SIGHUP
handle_sighup() {
 echo "Received SIGHUP"
 exit 1
}

# Register the function as a signal handler
trap handle_sighup SIGHUP

# Main program loop
while true; do
 sleep 1
done
```

`justfile`
```just
sighup:
  bash ./sighup.sh
```

Problematic behaviour:

```bash
# Run recipe and send it to the background
just sighup &

# Attempt to SIGHUP the `just` process
kill -1 $(pgrep -f 'just sighup')
# Problem here: Nothing happens!

# Attempt to SIGHUP the child process directly
kill -1 $(pgrep -f 'bash ./sighup.sh')
# Got the following expected stdout output
Received SIGHUP
[1]  + 21612 exit 130   just sighup
```

Expected behaviour:

```bash
# Run recipe and send it to the background
just sighup &

# Attempt to SIGHUP the just process
kill -1 $(pgrep -f 'just sighup')

# Get the following expected stdout output and process termination
Received SIGHUP
[1]  + 21612 exit 130   just sighup
```

Happy to contribute a fix for this, but wanted to check in first if this is something worth fixing, or is it an expected behaviour? Also, any pointers on where to look for would be appreciated.

davoclavo:
This issue also might be related to these other issues:

https://github.com/casey/just/issues/1558
https://github.com/casey/just/issues/1560

damccull:
I'm also experiencing this when using just to launch tasks inside of zellij. I would love to see a fix for this. For now I'm going to have to adjust my zellij layout to launch the commands directly instead of using just. Hopefully can migrate back soon.

kuglee:
I've also ran into this and it's not ideal. I'd be great to have a fix/option for this.

casey:
[...]

Issue #1560, opened 2023-03-02
Trap signals?


OJFord:
I'm trying to do something like:
```justfile
debug: fmt
    #!/usr/bin/env bash
    function build() {
        echo "Building for debug"
        go build -gcflags='all=-N -l' -o debug.out
    }
    build
    trap build SIGUSR1
    gdlv -r stdout:/dev/stdout exec ./debug.out -debug
```

But then `just debug` doesn't respond to <kbd>Ctrl</kbd><kbd>T</kbd> in the same shell (nor the gdlv window), `kill -USR1` on the PID of the `bash /tmp/just<random>/debug` process does nothing, and the same on the `just debug` process terminates it (and also seems to propagate the signal to the parent shell).

Is there a way of doing this?

stefanchrobot:
Similar story here: `just` doesn't seem to forward signals to the command it's running. My app terminates nicely when it receives SIGTERM, but nothing happens if the same command is invoked via `just`.
just 1.5.0

LLBlumire:
I believe the same issues causes https://github.com/casey/just/issues/1558

vadixidav:
This also creates an issue where I need to still cache some data even if a command fails, but currently ctrl+c isn't counted as a command failure and simply terminates just entirely.

casey:
Closing in favor of #2473, where I tried to collect everything related to signals in one place. `just` should not touch `SIGUSR1` or `SIGINFO` (generated by CTRL-T) at all, so if that's happening, please reopen this issue. @stefanchrobot just ignores `SIGTERM`, but should not change whether or not children of `just` receive `SIGTERM`. So I'm not sure what's going on there.

Issue #1558, opened 2023-02-28
Running Firestore Emulator From Just Prevents Control C


LLBlumire:
I have the following in a Justfile.

```
start:
    gcloud emulators firestore start
```

If I run the command on it's own, I am able to CTRL+C to stop the emulator. If I run the command through just with `just start`, CTRL+C does nothing.

---

For a reproducible example of this here's a `shell.nix` which will set up an environment that demonstrates this.

```
{ pkgs ? import <nixpkgs> {} }:
let
    gcloud = pkgs.google-cloud-sdk.withExtraComponents([
        pkgs.google-cloud-sdk.components.cloud-firestore-emulator
    ]);
    shell = pkgs.mkShell {
        buildInputs = [
            pkgs.just
            pkgs.jdk11
            gcloud
        ];
    };
in
shell
```

If you run

```
nix-shell --pure
gcloud emulators firestore start
<C-c>
just start
<C-c>
```

You'll see the first control C does stop the program, but the second one does nothing (and the only way to exit is to start a new terminal and kill the java process).

Hopefully the nix environment is enough to reproduce the error (that's the whole point of nix after all)!

LLBlumire:
On further experimentation, I believe this may be connected to the fact that just doesn't forward SIGTERM into it's child processes. I'm uncertain however

runeimp:
Checkout issue #302

LLBlumire:
Is there any configuration required to enable that feature? As it does not appear the signal is being forwarded?

runeimp:
To my knowledge that feature was never added. Sorry, I thought that was obvious from the other discussion. Maybe not as cut-and-dry as I thought though.

LLBlumire:
The issue was closed as completed and it looked to me as https://github.com/casey/just/pull/345 provided this functionality somehow? I may have misunderstood! So there is no way to have signals be forwarded to the program Just is running? and there's no scope planned for this to be implemented?

runeimp:
Ah, you're right. Looks like that was my mistake. Not sure how to activate it or anything. @casey ?

casey:
Closing in favor of #2473, where I've tried to collect everything related to signal handling in one place. Take a look and see if I've missed anything! A reproduction of this which doesn't depend on `gcloud` would be much appreciated.

Issue #302, opened 2018-02-14
Wait for child process to exit on SIGINT


bheisler:
Thanks for working on Just, I've found it quite useful since I started using it.

I've noticed that if the user presses Ctrl-C to kill a command while Just is running it, Just terminates immediately - without waiting for the child process to end. That leaves the child process orphaned, but still attached to the terminal. This is especially annoying if the child process doesn't handle Ctrl-C correctly and continues writing output - you have to use `ps` to find and `kill` it.

It would be nice if Just caught Ctrl-C's and waited for the current child process to terminate before exiting itself.

casey:
Glad you like it!

Hmm, it's not clear to me why that behavior would be better. If `just` waits for the child process to exit, and the child process doesn't handle ctrl-C correctly, then you still have to `ps` and `kill` it. Or am I missing something?

bheisler:
Well, I admit that my use-case is a little unusual here. The process in question is a unit test runner that interprets Ctrl-C as a failed test. If I run the tests from the terminal directly, I can hold Ctrl-C until it runs out of tests and exits on its own, but I can't do that if `just` terminates immediately, because that stops the Ctrl-C signal from reaching the test runner.

More generally, I think users would expect this behavior and it's generally best to conform to user expectations where reasonable. Waiting for the child process to finish makes `just` behave the same way as if I had simply run the commands myself by pasting them into the terminal. It also has the benefit of making it very clear to the user when the process finally terminates, because they don't see the terminal prompt again until it does.

casey:
Gotcha, that sounds reasonable.

A simple implementation would be to install a signal handler that ignores SIGINT before running a recipe, and then uninstall it once that recipe is done. It looks like [Detegr/rust-ctrlc](https://github.com/Detegr/rust-ctrlc) would be good for this.

Would you be interested in submitting a PR? This would be a pretty good first issue, for you or for anyone interested in contributing to `just`.

I've actually been meaning to make a tag for good first issue, so I think I'll do that now, and also tag this issue.

PR #1803, opened 2023-12-30
Improve signal/interrupt handling

davoclavo:
Howdy! I took a stab to fix multiple issues related to signal handling  - and address some improvements on the current signal/interruption handler

https://github.com/casey/just/issues/1558
https://github.com/casey/just/issues/1560
https://github.com/casey/just/issues/1781

A summary of the changes:

- Renamed the `interrupt_handler` to `signal_handler` to make it more explicit that any arbitrary signal can be handled
- Replace [ctrlc](https://github.com/Detegr/rust-ctrlc) with [signal-hook](https://github.com/vorner/signal-hook) to be able to handle specific signals (such as SIGHUP) and be able to propagate them independently, instead of handling them all via a single handler
  - A substantial change was done to handle continuously polling the process via `child.try_wait()`, instead of handler that was used previously with `ctrlc`
- Fix error codes when recipes/commands/shebangs/inline are terminated via a signal by [adding 128 to the signal termination code](https://tldp.org/LDP/abs/html/exitcodes.html)
- Add [command-group](https://github.com/watchexec/command-group) to be able to handle shebang command executions which may cause sub-processes to be spawned
- Re-enable signal/interrupt tests and add a test that validates that multiple signals can be sent if the program is able to trap them

As a side note, I noticed that the current interruption handling mechanism was incrementing the block counter via the `block()` method, but was never calling `unblock()` thus decreasing the block counter - I assume this implementation was done with concurrent processes in mind, however it seemed to be incomplete. As far as I am aware there are no concurrent processes/recipes that would need this behaviour, but could definitely take it in mind if required.

Bonus:
Possibly re-fix this other reported issue a while back, which from my shallow understanding, currently multiple Ctrl-C's were not able to be forwarded to the child process due to the aforementioned block counter reason
https://github.com/casey/just/issues/302

Please let me know if I should adjust any part of the implementation, more than happy to get some feedback and make appropriate adjustments :)


src/platform.rs
```diff
@@ -38,6 +38,10 @@ impl PlatformInterface for Platform {
     exit_status.signal()
   }
 
+  fn exit_code_from_signal(signal: i32) -> i32 {
```
casey, line 42: Does this need to be implemented on Windows as well?

src/platform_interface.rs
```diff
@@ -16,6 +16,11 @@ pub(crate) trait PlatformInterface {
   /// signal
   fn signal_from_exit_status(exit_status: ExitStatus) -> Option<i32>;
 
+  /// Convert a signal into an exit code, for the case where a process was
+  /// terminated by a signal
+  /// See https://tldp.org/LDP/abs/html/exitcodes.html
```
casey:
```suggestion
  /// Convert a signal number into an exit code
```

src/platform.rs
casey, line 42: I think the link to tldp.org only applies on unix, so let's move it here:

```suggestion
  /// See https://tldp.org/LDP/abs/html/exitcodes.html
  fn exit_code_from_signal(signal: i32) -> i32 {
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+
+use command_group::{CommandGroup, GroupChild};
+use signal_hook::consts::signal::*;
+use signal_hook::consts::TERM_SIGNALS;
+use signal_hook::flag;
+use signal_hook::iterator::exfiltrator::WithOrigin;
+use signal_hook::iterator::SignalsInfo;
+use std::io::Read;
+use std::process::Command;
+use std::sync::{atomic::AtomicBool, Arc};
```
casey:
```suggestion
use {
  super::*,
  command_group::{CommandGroup, GroupChild},
  signal_hook::{flag, consts::{TERM_SIGNALS, signal::*},iterator::{exfiltrator::WithOrigin, SignalsInfo}},
  std::io::{Read, process::Command, sync::{atomic::AtomicBool, Arc}},
}
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+  pub(crate) fn instance() -> MutexGuard<'static, Self> {
+    static INSTANCE: Mutex<SignalHandler> = Mutex::new(SignalHandler::new());
+
+    match INSTANCE.lock() {
+      Ok(guard) => guard,
+      Err(poison_error) => {
+        eprintln!(
+          "{}",
+          Error::Internal {
+            message: format!("interrupt handler mutex poisoned: {poison_error}"),
```
casey:
```suggestion
            message: format!("signel handler mutex poisoned: {poison_error}"),
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+    instance.verbosity = verbosity;
+
+    let signal_flag = Arc::new(AtomicBool::new(false));
+
+    const MISC_SIGNALS: &[i32] = &[
+      #[cfg(not(windows))]
+      SIGHUP,
+      // SIGABRT,
+      // SIGPIPE,
+      // SIGALRM,
```
casey: Should these be uncommented or removed?

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+    ];
+
+    let mut sigs = MISC_SIGNALS.to_vec();
```
casey:
```suggestion
    let mut signals = MISC_SIGNALS.to_vec();
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+    sigs.extend(TERM_SIGNALS);
```
casey:
```suggestion
    signals.extend(TERM_SIGNALS);
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    let signals = SignalsInfo::<WithOrigin>::new(&sigs)?;
+    instance.signals = Some(signals);
```
casey:
```suggestion
    instance.signal_info = Some(SignalsInfo::<WithOrigin>::new(&signals)?);
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    for sig in TERM_SIGNALS {
```
casey:
```suggestion
    for signal in TERM_SIGNALS {
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+      // NOTE This could be set before the register handler, in order to terminate in a double ctrl-c
+      // flag::register_conditional_shutdown(*sig, 1, Arc::clone(&signal_flag))?;
```
casey: I don't understand this comment, could you elaborate?

src/signal_handler.rs
casey: Are we unconditionally setting this on the instance every time? Could we instead set it once in the `instance` method?

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+      flag::register(*sig, Arc::clone(&signal_flag))?;
+    }
+
+    for sig in MISC_SIGNALS {
```
casey:
```suggestion
    for signal in MISC_SIGNALS {
```

This loop looks like it's doing the same thing as the one above, can we combine them?

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+      flag::register(*sig, Arc::clone(&signal_flag))?;
+    }
+    Ok(())
+  }
+
+  pub(crate) fn instance() -> MutexGuard<'static, Self> {
+    static INSTANCE: Mutex<SignalHandler> = Mutex::new(SignalHandler::new());
```
casey:
```suggestion
    static INSTANCE: Mutex<SignalHandler> = Mutex::new(Self {
      verbosity: Verbosity::default(),
      signal_info: None,
    });
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+      }
+    }
+  }
+
+  const fn new() -> Self {
+    Self {
+      verbosity: Verbosity::default(),
+      signals: None,
+    }
+  }
```
casey:
```suggestion
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+use signal_hook::flag;
+use signal_hook::iterator::exfiltrator::WithOrigin;
+use signal_hook::iterator::SignalsInfo;
+use std::io::Read;
+use std::process::Command;
+use std::sync::{atomic::AtomicBool, Arc};
+
+pub(crate) struct SignalHandler {
+  verbosity: Verbosity,
+  signals: Option<SignalsInfo<WithOrigin>>,
```
casey:
```suggestion
  signal_info: Option<SignalsInfo<WithOrigin>>,
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+    Self {
+      verbosity: Verbosity::default(),
+      signals: None,
+    }
+  }
+
+  fn wait_child(mut child: GroupChild) -> std::io::Result<(GroupChild, ExitStatus)> {
+    let mut instance = Self::instance();
+    loop {
+      if let Some(signals) = instance.signals.as_mut() {
```
casey:
```suggestion
      if let Some(signal_info) = signal_info.signals.as_mut() {
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+        for signal_info in signals.pending() {
```
casey:
```suggestion
        for info in signals.pending() {
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+          unsafe {
+            // Send signal to the child process group, thus the negative pid
+            libc::kill(-(child.id() as i32), signal_info.signal);
```
casey:
```suggestion
            libc::kill(-(child.id() as i32), info.signal);
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  fn wait_child(mut child: GroupChild) -> std::io::Result<(GroupChild, ExitStatus)> {
+    let mut instance = Self::instance();
+    loop {
+      if let Some(signals) = instance.signals.as_mut() {
+        for signal_info in signals.pending() {
+          unsafe {
+            // Send signal to the child process group, thus the negative pid
```
casey:
```suggestion
            // Send signal to the process group using negative pid
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+            libc::kill(-(child.id() as i32), signal_info.signal);
+          }
+        }
+      }
+      match child.try_wait() {
+        Ok(None) => {
+          std::thread::sleep(std::time::Duration::from_millis(10));
```
casey: 10 milliseconds seems aggressive, let's do 50 instead, since I think a user is unlikely to notice a difference:

```suggestion
          std::thread::sleep(std::time::Duration::from_millis(50));
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+        }
+      }
+    }
+  }
+
+  pub(crate) fn guard_output(mut command: Command) -> Result<String, OutputError> {
+    let child = command
+      .stdout(std::process::Stdio::piped())
+      .group_spawn()
+      .unwrap();
```
casey: Can we avoid the unwrap here?

```suggestion
?;
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+          }
+          None => Err(OutputError::Unknown),
+        }
+      }
+      Err(error) => Err(OutputError::Io(error)),
+    }
+  }
+
+  pub(crate) fn guard(mut command: Command) -> Result<std::process::ExitStatus, std::io::Error> {
+    let child = command.group_spawn().unwrap();
```
casey: Can we avoid the unwrap here?

```suggestion
    let child = command.group_spawn()?;
```

src/signal_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+    match Self::wait_child(child) {
+      Ok((_, status)) => Ok(status),
```
casey, line 158: Do we also need to get the status and convert the error into a `Signal` error, like in the above `guard_output`?

casey (changes requested):
Dope! Lots of comments, mostly because I don't fully understand what's going on here 😅

casey:
BTW, are you the Davo that I know IRL?

davoclavo:
Thanks a lot for the feedback/comments, I will work on fixing them! I will also address the issues clippy surfaced.

And yes, I am that davo from rc 😄 - stoked to see just and ords becoming big creatures!

src/signal_handler.rs
davoclavo: I left them commented because potentially we could intercept them, but not sure if it would be desired, as I am unsure if it could break other things. For instance, trying to add SIGKILL causes a panic, because that signal is unblockable, or unhandleable.

 I will remove them for now, and if the need arises in the future we can just add them if people request use for them.

src/signal_handler.rs
davoclavo: Seems like this particular change is not possible due to the following error:


```rust
41 |     static INSTANCE: Mutex<SignalHandler> = Mutex::new(Self {
   |                                                        ^^^^
   |                                                        |
   |                                                        use of generic parameter from outer item
   |                                                        refer to the type directly here instead
   ```
   
   So will do this instead, and also remove `fn new`
   
   ```rust
    static INSTANCE: Mutex<SignalHandler> = Mutex::new(SignalHandler {
      verbosity: Verbosity::default(),
      signals_info: None,
    });
```

src/signal_handler.rs
davoclavo: Good question, this is already happening via `error_from_signal` [here for linewise](https://github.com/casey/just/blob/master/src/recipe.rs#L258-L262) and [here for shebangs](https://github.com/casey/just/blob/master/src/recipe.rs#L389), but it is not currently being converted for subcommands. [source](https://github.com/casey/just/blob/master/src/justfile.rs#L162-L176)

I could try to refactor `fn guard` so it returns a `RunResult<'src, ()>` - Happy to try to refactor some things, although I am afraid it might involve quite a bit more changes all over the place

src/signal_handler.rs
davoclavo: This is a mechanism provided by `signal_hook` to allow the program to exit if you hit ctrl-c twice, in case the handler is stuck for some reason.

More info here: https://docs.rs/signal-hook/latest/signal_hook/#a-complex-signal-handling-with-a-background-thread 

I am inclined to delete this commented code and also unify both `for` statements (from your other comment)

src/platform.rs
davoclavo: Good point, I forgot to add that to the windows platform bit, will do that.

I am unsure how the exit codes map tho, so I currently will just propagate the original one without any transformations if that's ok.

```rust
 fn exit_code_from_signal(signal: i32) -> i32 {
   signal
 }
 ```

src/signal_handler.rs
davoclavo: Not sure if we need to add the signal handler registration to the `instance` method, as we probably want a single handler for signals that we can then decide to forward to the current single child process that is running, and potentially N-child processes that could be running concurrently

This is currently only executed once during `SignalHandler::install` happening [here](https://github.com/casey/just/blob/master/src/config.rs#L676-L678)

I am very likely misunderstanding your question, or could be missing something obvious tho. I would appreciate if you could elaborate a bit further if that's the case.

davoclavo:
Alright, comments adjusted, and made some additional improvements I noticed.

I did not mark any of your comments as "resolved" myself, as I am not sure what the right etiquette is for that process, but here is a list of the changes. I think I fixed all or most of your suggestions, and the ones I needed more clarification I left you a response in each thread. Happy to take more feedback if needed.

- Remove `unwrap()` calls
- Implement `exit_code_from_signal` for windows `platform.rs` part
- Refactored `use` statements
- Cleaned up commented code
- Increased sleep time in between polling to 50ms
- Simplified Signal registering logic
- Changed `SignalHandler::instance` method from public to private
- Deleted `SignalHandler::new` method
- Fixed clippy comments, especially the one regarding conversions from u32 to i32

src/signal_handler.rs
tgross35: Typo in the suggestion

davoclavo: Thanks! I did notice and fixed it 😄

davoclavo:
Rebased and took the liberty to resolve all trivial comments to make it less of a burden to review

src/signal_handler.rs
davoclavo: Deleted all commented code and made this function quite a bit simpler

casey (changes requested):
I tweaked a few things and pushed them to the branch:
- Removed `SignalHandler::verbosity` since it looks like it we don't need it anymore
- Made `SignalHandler::install` set the whole struct and inlined some stuff
- Added `#[ignore]` back to the test, since for some reason they don't run when I run the tests inside `cargo watch`
- Made `#[ignore]`d tests run on CI
- Run ignored tests when running `just test`

It looks like this isn't working on Windows CI, there appear to be some undefined symbols.

davoclavo:
Removed `extended-siginfo` feature from `signal-hook` crate which _might_ be causing this issue, and we didn't really need it in the first place

Could you try running the workflow once again to see if tests pass for the Windows CI workflow?

if it fails, I'll do more research and try to get an easy way to replicate the problem and find a solution 🪟

casey:
Just re-ran the checks, it looks like the missing symbol errors are gone, but there are a few more issues.

davoclavo:
Pushed a couple more changes to address these new issues, hopefully all the CI checks pass this time 🤞  - thanks for the patience!

Also... I found what seems to be a `libc::kill` replacement for [windows](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/Console/fn.GenerateConsoleCtrlEvent.html), so we may be able to enable the `signal` test for windows as well (modulo the test that uses SIGHUP, which apparently is not available in windows). Sadly I don't have a windows box handy to iterate and make sure things work, and have no clue how to set up a windows-latest github runner. I will try to set up one in my fork.

davoclavo:
I got my windows runner working and there are some other issues, let me fix them and I will let you know when everything is green

casey:
This looks a little stale, so I'm going to convert it to a draft.

casey:
This is pretty stale, so closing. See #2473, where I try to get to the bottom, once and for all, of `just`'s current signal handling behavior.

PR #1805, opened 2023-12-30
chore(README): Remove toc link

michidk:
For logged-in users (might be only for the ones in the feature preview, not sure), the table of contents are on the other side:
image

Since this is not consistent for every user and might change in future, I would be in favor of removing this notice from the readme. This is something GitHub has to fix, not us!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,3 @@
-↖️ Table of Contents
-
 <h1 align="center"><code>just</code></h1>
 
 <div align="center">
--- a/README.中文.md
+++ b/README.中文.md
@@ -1,5 +1,3 @@
-↖️ 目录
-
 <h1 align="center"><code>just</code></h1>
 
 <div align="center">


casey:
I definitely agree this should be GitHub's problem, but I got a few requests for a table of contents, when people didn't see GitHub's, so it is what it is 😂 I changed the arrow to point up and to the right in #1806. I don't have any feature previews enabled that would affect this, so I think it's now on the upper right for everybody.

michidk:
@casey Now it's on the wrong side if you are not logged in 😅

image

casey:
fml 🤦

casey:
I think there was recently a redesign, and logged in users see it, and logged out users don't. I assume the redesign will eventually get to logged out users, so I'm leaning towards keeping it how it is.

michidk:
Normally such things are behind feature previews:
image

If this one isn't (confirmed with a second account that has no feature previews enabled), then I think it's by intention.

Btw, even when logged out, in this view the TOC is on the right as well:
https://github.com/casey/just/blob/master/README.md

So I guess its on the right in more views than it is on the left 😅 What a mess.

I personally think people using Just, should be able to properly use GitHub. But the root cause is that the readme is just very long and you have to scroll a lot to actually arrive at the quick start. Looking it other big projects, they mostly use their readme differently. E.g. they have the list of supported platforms and editor plugins and some other pages.

You probably discussed this before, but what do you think about simplifying the readme and putting all the instructions into a https://github.com/rust-lang/mdBook or GitHub Wiki? I think this would increase readbility by a lot.

casey:
The readme is actually also available as an mdbook, which is automatically generated from the readme: https://just.systems/man/en/

The long-ass readme is definitely a bit insane, but I think it's also useful. You can easily `ctrl-f` to find whatever you need from the repo front page, and it's much nicer to edit and search a single long file in the terminal / an editor, instead of having them in a bunch of different files.

So I think if there's any improvement to be made, it's probably improving the beginning of the readme, so that people are more likely to find what they need easily. For example, that they want to read it in book form, or quick installation inscriptions, or a quickstart type section.

PR #1816, opened 2024-01-03
Fix recipe doc alignment in just --list

crdx:
`recipe.private` doesn't take into account recipes with the `[private]` attribute, so private recipes affect the alignment calculation of the the recipe docs. This causes them to align too far to the right.

Use `recipe.public()` instead which appears to be the canonical way of differentiating between public and private.


casey:
Nice, good catch! Can you add a test for this?

crdx:
Yep. Should it go in `tests/modules.rs` or a new file `tests/subcommand.rs`?

crdx:
I went ahead and added it to `tests/modules.rs` as that felt like the right place. Verified it fails before the change and passes after.

casey:
Whoops, sorry for forgetting about this. Looks good to me!

PR #1819, opened 2024-01-05
Add cli arg to skip dependencies

ngharrington:
New PR to address https://github.com/casey/just/issues/1314


casey:
Nice! I refactored it a little:

- Changed the name to `--no-deps`, since that seems more natural
- `no_depsendencies` is on the `RecipeContext`, so I got it from there, instead of passing it separatelly

PR #1822, opened 2024-01-06
Add functions to return standard directories

### LINKED ISSUE #1820: Feature request: functions to get XDG base directories (home, config, etc)
Install-like or system configuration steps usually need to know some variant of these, but they aren't always easy to discover natively Would it be possible to provide an easy way to get this information? 

My suggested implementation is to wrap the [directories](https://docs.rs/directories/5.0.1/directories/index.html) crate (which obeys XDG) and expose something like the following:

```just
# BaseDirs::home_dir
home_directory() 

# BaseDirs::cache_dir
cache_directory() 

# BaseDirs::config_dir
config_directory() 

# BaseDirs::config_dir_local
local_config_directory() 

# directories::ProjectDirs::from("", "", "MyProject").map_or("", ProjectDirs::project_path)
proj_path := project_path("", "", "MyProject")

# ProjectDirs::x calls
project_cache_directory(proj_path)
project_config_directory(proj_path)
project_data_directory(proj_path)

# std::env::temp_dir
temp_dir()

# ...
```

Not every option is needed of course, but some of these would make some system setup a lot easier!

tgross35:
Add the following builtin functions:

- `cache_directory()`
- `config_directory()`
- `config_local_directory()`
- `data_directory()`
- `data_local_directory()`
- `executable_directory()`
- `home_directory()`

These use the `dirs` crate to return XDG-compliant paths, or the system defaults.

Fixes: https://github.com/casey/just/issues/1820

### DIFF
--- a/README.md
+++ b/README.md
@@ -1421,7 +1421,7 @@ These functions return paths to user-specific directories for things like
 configuration, data, caches, executables, and the user's home directory. These
 functions follow the
 [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html),
-and use implementation provided by the
+and are implemented with the
 [`dirs`](https://docs.rs/dirs/latest/dirs/index.html) crate.
 
 - `cache_directory()` - The user-specific cache directory.
--- a/README.md
+++ b/README.md
@@ -1424,20 +1424,13 @@ functions follow the
 and use implementation provided by the
 [`dirs`](https://docs.rs/dirs/latest/dirs/index.html) crate.
 
-- `cache_directory()` - Locate the user-specific cache directory. Returns an
-   empty string if it could not be found.
-- `config_directory()` - Locate the user-specific config directory. Returns an
-   empty string if it could not be found.
-- `config_local_directory()` - Locate the user-specific config_local directory.
-   Returns an empty string if it could not be found.
-- `data_directory()` - Locate the user-specific data directory. Returns an empty
-   string if it could not be found.
-- `data_local_directory()` - Locate the user-specific data_local directory.
-   Returns an empty string if it could not be found.
-- `executable_directory()` - Locate the user-specific executable directory.
-   Returns an empty string if it could not be found.
-- `home_directory()` - Locate the user-specific home directory. Returns an empty
-   string if it could not be found.
+- `cache_directory()` - The user-specific cache directory.
+- `config_directory()` - The user-specific configuration directory.
+- `config_local_directory()` - The local user-specific configuration directory.
+- `data_directory()` - The user-specific data directory.
+- `data_local_directory()` - The local user-specific data directory.
+- `executable_directory()` - The user-specific executable directory.
+- `home_directory()` - The user's home directory.
 
 ### Recipe Attributes
 
--- a/README.md
+++ b/README.md
@@ -1415,6 +1415,30 @@ which will halt execution.
   `requirement`, e.g., `">=0.1.0"`, returning `"true"` if so and `"false"`
   otherwise.
 
+##### Directory Locaters
+
+- `cache_directory()` - Locate the user-specific cache directory. Returns an
+   empty string if it could not be found.
+- `config_directory()` - Locate the user-specific config directory. Returns an
+   empty string if it could not be found.
+- `config_local_directory()` - Locate the user-specific config_local directory.
+   Returns an empty string if it could not be found.
+- `data_directory()` - Locate the user-specific data directory. Returns an empty
+   string if it could not be found.
+- `data_local_directory()` - Locate the user-specific data_local directory.
+   Returns an empty string if it could not be found.
+- `executable_directory()` - Locate the user-specific executable directory.
+   Returns an empty string if it could not be found.
+- `home_directory()` - Locate the user-specific home directory. Returns an empty
+   string if it could not be found.
+
+The output for these commands is produced by the
+[`dirs`](https://docs.rs/dirs/latest/dirs/index.html) crate, which first checks
+for `XDG_*` environment variables then falls back to platform defaults.
+
+The XDG specification is available here:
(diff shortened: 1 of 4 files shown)


laniakea64:
Would it be better to use [`dirs::preference_dir`](https://docs.rs/dirs/latest/dirs/fn.preference_dir.html) for `config_directory()`?

tgross35:
> Would it be better to use [`dirs::preference_dir`](https://docs.rs/dirs/latest/dirs/fn.preference_dir.html) for `config_directory()`?

I think it would be better to keep names consistent with what they are on `dirs`, which could mean adding `preference_directory`. But the only difference between the two is on mac - `/Users/Alice/Library/Application Support`(`config_dir`) contains a directory with the application name and then whatever contents is needed, pretty in line with other platforms.

`/Users/Alice/Library/Preferences` (`preference_dir`) only contains `com.*.*.plist` files, which are a something like a binary JSON. I think less likely to be used than the config directory where you can just place files.

src/function.rs
```diff
@@ -114,6 +121,10 @@ fn clean(_context: &FunctionContext, path: &str) -> Result<String, String> {
   Ok(Path::new(path).lexiclean().to_str().unwrap().to_owned())
 }
 
+fn dirs_map(f: fn() -> Option<PathBuf>) -> Result<String, String> {
+  Ok(f().map_or(String::new(), |p| p.to_string_lossy().into_owned()))
```
casey: We should use String::from_utf8, and return an error if conversion fails. `to_string_lossy` will replace invalid UTF-8 sequences with replacement characters, if the user subsequently tries to access the returned path, it will fail.

We should also add a `&'static str` argument to `dirs_map`, which is the name of the directory "cache directory", etc, we're looking up, and include it in the error message.

casey: Also, let's call `dirs_map` `dir`.

README.md
```diff
@@ -1415,6 +1415,28 @@ which will halt execution.
   `requirement`, e.g., `">=0.1.0"`, returning `"true"` if so and `"false"`
   otherwise.
 
+##### Directory Locaters
+
+- `cache_directory()` - Locate the user-specific cache directory. Returns an
```
casey: Let's call these `cache_dir` `config_dir`, etc.

README.md
```diff
@@ -1415,6 +1415,28 @@ which will halt execution.
...
+   string if it could not be found.
+- `data_local_directory()` - Locate the user-specific data_local directory.
+   Returns an empty string if it could not be found.
+- `executable_directory()` - Locate the user-specific executable directory.
+   Returns an empty string if it could not be found.
+- `home_directory()` - Locate the user-specific home directory. Returns an empty
+   string if it could not be found.
+
+The output for these commands is produced by the [`dirs`] crate, which first
+checks for `XDG_*` environment variables then falls back to platform defaults.
```
casey: Let's also link to the XDG spec.

README.md
```diff
@@ -1415,6 +1415,28 @@ which will halt execution.
...
+
+[`dirs`]: https://docs.rs/dirs/latest/dirs/index.html
```
casey: I generally don't bother with out-of-line links`, so let's just put this one inline.

casey (changes requested):
Awesome! Thanks so much for the docs and tests, a lot of PRs miss those. See review for comments.

src/function.rs
tgross35: Thanks for the review! I made this change

README.md
tgross35: Added

README.md
tgross35: Changed

README.md
tgross35: I think it is good to have consistency with `invocation_directory` `justfile_directory` and `parent_directory`, which is why I did it this way. If you prefer `_dir` here, would you be okay adding aliases to the existing functions as e.g. `justfile_dir`?

 I get confused with APIs that alternate naming :)

casey: It is a bit confusing, but I think most people won't wind up looking at the `dirs` docs, so being out of sync won't be a problem.

tgross35: I think I'm misunderstanding - are you agreeing with my statement above?

My suggestion was either to leave the PR as-is so Just's functions are `justfile_directory`, `config_directory`, `cache_directory` etc (consistent within Just but inconsistent with `dirs` is fine), or to use `config_dir` but also add aliases so `justfile_dir`, `invocation_dir`, `parent_dir` all work. As opposed to having `justfile_directory` but `cache_dir`, and maybe needing to look up which is which.

Do either of these sound better to you?

casey: Ah yeah, I see what you're saying. In that case, let's us `_directory`, and later I might open a PR which adds `_dir` aliases for everything.

tgross35: That works for me 👍

casey (approved):

tgross35:
I see you changed this to return an error if the directory is not found (thanks for the updates!). Would it be better to make these accept either 0 or 1 argument, so a fallback can be provided if not available?

tgross35:
Or looks like it's ready to merge. If the above sounds reasonable, I can submit a followup PR :)

casey:
Nice! I tweaked the readme a little, and changed the functions to return an error if the directory isn't found. Also, for some reason, on my mac the executable directory wasn't found, so I modified the test to handle this error.

casey:
> I see you changed this to return an error if the directory is not found (thanks for the updates!). Would it be better to make these accept either 0 or 1 argument, so a fallback can be provided if not available?

Whoops, I just noticed this comment. If people want it, I think it would be fine to modify these functions to accept a fallback which is returned if the directory isn't found. We could either wait and see if anyone complains, or if you feel like it, and do it in a follow up PR.

tgross35:
Sounds great. Thanks!

PR #1831, opened 2024-01-09
Fix unnecessary abbreviation of just in README

thled:


casey:
Merged, thanks for the PR!

PR #1833, opened 2024-01-10
Add function returning PID of `just` process

### LINKED ISSUE #1019: Add function that returns PID of the just process
Mentioned by @page-down in #1011.

Swordelf2:
Fixes #1019.

I had to split `Test::run()` into `spawn()` and `check_output` to allow inspection of `process::Child` (for its PID) before setting expected `stdout`. I'm open to any other solutions.


tests/test.rs
```diff
@@ -1,3 +1,5 @@
+use std::process::Child;
```
casey: I would put this import in `lib.rs`, since it's common to import `std::process::Child` as just `Child`.

tests/test.rs
```diff
@@ -177,19 +179,17 @@ impl Test {
 }
 
 impl Test {
-  pub(crate) fn run(self) -> Output {
+  pub(crate) fn run(mut self) -> Output {
```
casey: Instead of breaking this function up, how about adding the pid to the output struct? That way, callers can all still use run, and if they care about the pid, they can get it from the output.

tests/test.rs
```diff
@@ -200,15 +200,24 @@ impl Test {
       command.args(["--shell", "bash"]);
     }
 
-    let mut child = command
-      .args(self.args)
+    command
+      .args(self.args.drain(..))
```
casey: What's this drain for?

casey (changes requested):
Nice addition! Check out my review comments.

tests/test.rs
Swordelf2: Done.

tests/test.rs
Swordelf2: Yeah, this makes much more sense.

Done.

tests/test.rs
Swordelf2: It was needed because now that `self` was behind a reference, we could not do a partial move out of `self` by passing `self.args`.

This is gone with the revert of `Test::run()` split.

casey (approved):
Nice, LGTM!

PR #1834, opened 2024-01-10
Add confirm prompt

CramBL:
Add functionality to allow specifying a prompt for the `confirm` recipe attribute.

e.g. `[confirm("this recipe is dangerous because blabla")]`

It's a quick and dirty implementation, feel free to nitpick. And thanks a lot for your great work!


src/attribute.rs
```diff
@@ -1,12 +1,10 @@
...
 
-#[derive(
-  EnumString, PartialEq, Debug, Copy, Clone, Serialize, Ord, PartialOrd, Eq, IntoStaticStr,
-)]
+#[derive(EnumString, PartialEq, Debug, Clone, Serialize, Ord, PartialOrd, Eq, IntoStaticStr)]
 #[strum(serialize_all = "kebab-case")]
 #[serde(rename_all = "kebab-case")]
 pub(crate) enum Attribute {
-  Confirm,
+  Confirm(Option<Box<str>>),
```
casey: Can this just be `Option<String>`?

src/compile_error.rs
```diff
@@ -35,7 +35,7 @@ impl Display for CompileError<'_> {
       AliasInvalidAttribute { alias, attr } => write!(
         f,
         "Alias {alias} has an invalid attribute `{}`",
-        attr.to_str(),
+        attr.clone().to_str(),
```
```suggestion
        attr.to_str(),
```

tests/confirm.rs
```diff
@@ -103,3 +103,35 @@ fn do_not_confirm_recipe_with_confirm_recipe_dependency() {
...
+        ",
+    )
+    .stderr("This is dangerous - Run recipe `requires_confirmation`? echo confirmed\n")
+    .stdout("confirmed\n")
+    .stdin("y")
+    .run();
+}
+
+#[test]
+fn confirm_recipe_arg_with_prompt() {
```
casey: I think it's okay to remove this test, since this PR doesn't change the logic of whether or not a recipe with a confirm attribute is run.

src/recipe.rs
```diff
@@ -69,8 +69,15 @@ impl<'src, D> Recipe<'src, D> {
   }
 
   pub(crate) fn confirm(&self) -> RunResult<'src, bool> {
-    if self.attributes.contains(&Attribute::Confirm) {
-      eprint!("Run recipe `{}`? ", self.name);
+    if let Some(Attribute::Confirm(prompt_text)) = self
```
casey: I think we should probably completely replace the prompt, so that users can customize it however they want. I.e., we either print out the default prompt if there's none given, or we print the user-specified prompt, and nothing else. This allows users fully customize the prompt.

casey: Let's do something like:

```
for attribute in self.attributes {
  if let Attribute::Confirm(prompt) = attribute {
    // print prompt
    break;
  }
}
``

**Review by casey — APPROVED** [review]
Looks great! Check out the review comments. If you feel like it, also take a crack at documenting this in the readme.

File: src/attribute.rs (hunk 1 above)
Comment (CramBL) [reply src/attribute.rs]: Yes, I use `Box<str>` to not include the capacity field in the variant. Since this is now by far the largest enum variant, I want to reduce its size. I'll make it an `Option<String>` instead for simplicity.

File: src/recipe.rs (hunk 4 above)
Comment (CramBL) [reply src/recipe.rs]: That seems better yes.

Comment (CramBL) [reply src/recipe.rs]: I simplified the logic with a loop like you suggested

**Comment by CramBL** [comment]
> Looks great! Check out the review comments. If you feel like it, also take a crack at documenting this in the readme.

I will document it in the README.md later today (I expect)

EDIT: Done.

File: README.md [hunk 5]
```diff
@@ -755,20 +755,20 @@ foo:
...
-| `dotenv-path` | string | - | Load a `.env` file from a custom path, if present. Overrides `dotenv-filename`. |
-| `export` | boolean | `false` | Export all variables as environment variables. |
-| `fallback` | boolean | `false` | Search `justfile` in parent directory if the first recipe on the command line is not found. |
-| `ignore-comments` | boolean | `false` | Ignore recipe lines beginning with `#`. |
-| `positional-arguments` | boolean | `false` | Pass positional arguments. |
-| `shell` | `[COMMAND, ARGS…]` | - | Set the command used to invoke recipes and evaluate backticks. |
-| `tempdir` | string | - | Create temporary directories in `tempdir` instead of the system default temporary directory. |
-| `windows-powershell` | boolean | `false` | Use PowerShell on Windows as default shell. (Deprecated. Use `windows-shell` instead. |
-| `windows-shell` | `[COMMAND, ARGS…]` | - | Set the command used to invoke recipes and evaluate backticks. |
+| Name                      | Value              | Default | Description                                                                                   |
```
casey: Can you undo adding alignment to these tables? I like the way alignment looks, but it produces large diffs when new items are added that change the alignment.

README.md
```diff
@@ -1509,6 +1510,16 @@ delete all:
   rm -rf *
 ```
 
+#### Customize recipe confirmation prompt <sup>unreleased</sup>
+
+The default prompt: ``Run recipe `foo`?`` can be overwritten with the `[confirm("custom prompt")]` syntax.
```
```suggestion
The default prompt: ``Run recipe `foo`?`` can be overwritten with `[confirm(PROMPT)]`:
```

README.md
```diff
@@ -1509,6 +1510,16 @@ delete all:
...
+
+```just
+[confirm("Are you sure you want to delete all? It is not reversible")]
+delete all:
```
```suggestion
[confirm("Are you sure you want to delete everything?")]
delete-everything:
```

src/parser.rs
```diff
@@ -917,7 +917,26 @@ impl<'run, 'src> Parser<'run, 'src> {
             first: *line,
           }));
         }
-        attributes.insert(attribute, name.line);
+        match attribute {
```
casey: How about something like this:

```rust
let arguments = if self.accepted(ParenL)? {
  let mut arguments = Vec::new();
  while self.accepted(StringToken)? {
    arguments.push(self.parse_string_literal()?);
  }
  self.expect(ParenR)?;
  Some(arguments)
} else {
  None
};

let attribute = if let Some(arguments) = arguments {
  attribute.with_arguments(arguments)?
} else {
  attribute
};
```

`Attribute::with_arguments` then returns an error if the attribute doesn't take the given number of arguments.

There are some places in the code which take the approach of parsing different depending on a keyword, but this leads to confusing error messages.

casey (changes requested):
Nice, looks good, check out my review comments.

src/parser.rs
CramBL: I revised the parsing but one test fails.

```shell
failures:

---- attributes::multiple_attributes_one_line_error_message stdout ----
Bad stderr: Diff < left / right > :
<"error: Expected ']', ',', or '(', but found identifier\n ——▶ justfile:1:17\n  │\n1 │ [macos, windows linux]\n  │                 ^^^^^\n"
>"error: Expected ']' or ',', but found identifier\n ——▶ justfile:1:17\n  │\n1 │ [macos, windows linux]\n  │                 ^^^^^\n"
```
It fails because recipe attributes generally accept arguments now. So maybe the error message should just be revised for the new kind? And then if you write e.g. `[linux("arg")]` it should say something like "expected 0 arguments got 1"?

Not sure what the preferred way of fixing it is.

src/attribute.rs
```diff
@@ -22,9 +20,41 @@ impl Attribute {
...
+  pub(crate) fn to_str(&self) -> &'static str {
     self.into()
   }
+
+  /// Returns a range from the min to max expected arguments of a given attribute
+  pub(crate) fn expect_args(&self) -> Range<usize> {
+    use Attribute::*;
+
+    match self {
+      Confirm(_) => 1..2,
```
casey: Should this be `0..=1`, since confirm takes 0 or 1 argument?

src/parser.rs
```diff
@@ -922,8 +922,33 @@ impl<'run, 'src> Parser<'run, 'src> {
             first: *line,
           }));
         }
+
+        let arguments = if self.accepted(ParenL)? {
+          let mut arguments = Vec::new();
+
+          while self.next_is(StringToken) {
```
casey: Actually, we only have one attribute that accepts a string literal, and if we're accepting multiple arguments, we would want to accept commas between them, so we can make `arguments` be an `Option`.

casey (changes requested):
Added some comments.

CramBL:
I made some semi-related changes to `DisplayRange` because it was bugged, and I needed to use ranges to make it more generic. I prefer using `RangeInclusive` because `Range` just looks ugly when you have e.g. `1..1` which at first glance appears to contain `1` but it is a range of length 0 that does not contain 1.

Now the `Display` impl for `DisplayRange` looks pretty messy but it is actually correct now.

src/parser.rs
casey: These tests can just be updated. When the parser encounters an unexpected token, in prints out the tokens that would have been acceptable. Since now, `(` is accepted after an attribute name, the `(` is part of the acceptable tokens.

src/attribute.rs
CramBL: Well is `[confirm()]` valid syntax?

casey: Ah, yeah, I see what you're saying. You're right, I think `confirm()` shouldn't be valid syntax.

src/parser.rs
CramBL: Comma-separated arguments are now accepted, do should comma+space also be accepted?

casey: Ah, no sorry, what I was getting at is that we don't have any multi-argument attributes, so we shouldn't bother either with parsing multiple arguments, with or without commas. We can just parse `( STRING )`.

CramBL: Alright, but now it is implemented, do you want it removed? It's like 2 lines and all the scaffolding is there now if there's a need/want for features that take advantage of recipe attributes with arguments. Off the top of my head it could be stuff like `[linux("arm64", "amd64")]` that enables recipes based on OS + Architecture, or adding kernel version for requirements etc.

But I respect if you invoke YAGNI, then I'll revert it.

casey: I agree we'll probably add it later, but that could be a while, so I'd assume just remove it. It also makes `Attribute::with_arguments` simpler, since it can take an `Option` instead of a `Vec`.

casey:
Merged! Thank you very much! This release is turning out to have a lot of great contributions ^_^

Issue #1776, opened 2023-12-24
Feature Request: Recipe grouping


nefrob:
Consider allowing recipes to be aded to groups. For example, using the existing recipe attributes format:

```
[group:lint]
format:
 ...

[group:lint]
check:
  ...

[group:build]
build:
  ...
```

and listed by group when calling `just --list`, ex.

```
Available recipes:
    [lint]
    format
    check
    
    [build]
    build

    [default]
    ...
```

List groups via `just --groups`, ex.

```
Groups:
  lint
  build
  default (undecorated recipes could fall in this group or this can be dropped)
```

There might be desire to allow recipes to be part of multiple groups, so this could be considered in the implementation but should not be an initial requirement.

This might also be a way to solve: https://github.com/casey/just/issues/1652. All recipes from a group could be executed at once rather than needing to match on recipe name.

neunenak:
This is a feature I'd like to have myself and I've started a PR working on it https://github.com/casey/just/pull/1842

kranurag7:
Hey @neunenak 
As a user, I would like to have all import in one group. 
```bash
$ tree .
.
├── justfile
└── justfiles
    ├── docker.justfile
    ├── gh.justfile
    ├── git.justfile
```
So, as per the format above, I'll have to add attributes in each of the recipe inside `docker.justfile` to have it included inside `docker` group. But it would be easier we can group based on imports as well. 

I think this can be a separate issue since this issue is about having attribute set on an individual recipe. 

Thanks for working on it.

casey:
I think it'll be easy to add make `[group: foo]` work on imports in a follow-up PR.

laniakea64:
Will these recipe groups have additional functionality beyond cosmetically improving list output?

For example, if https://github.com/casey/just/issues/626 is implemented, will recipe groups be used to define which recipes could be run in parallel and which recipes have to stay sequential?

marcbria:
It's cosmetic (helping a lot with the usability when you have long list) but not only.
[#1614](https://github.com/casey/just/issues/1614) is kind of related.

neunenak:
https://github.com/casey/just/pull/1842 is merged now. I think the suggestion in https://github.com/casey/just/issues/1776#issuecomment-1890386390  of having there be a way to tag imports with a group would be useful and I want to try to get that in as a follow-up PR, so I think it's worth keeping this issue open.

casey:
@neunenak I wanted to make a more specific issue, so I opened #2087 and #2086.

[...]

Issue #1648, opened 2023-07-18
Support for global justfiles


spartas:
I'm a recovering [Task][1] / [Taskfile][2] refugee, who's recently got into `just`. One of the features of Task that I use extensively is the `-g` ([global][3]) option, which will only search the user's `$HOME` directory (and no subdirectories) for a Taskfile.

This is very handy for running global tasks or recipes that do not need the current directory context, but it is also useful for running tasks that are somewhat generic, but do need the current directory as context (e.g. `docker compose` commands that use the services from the current working directory). Obviously this can be replicated in just with something like `just $HOME/`, but when used with flags and parameters, this quickly becomes unwieldy.

[1]: https://github.com/go-task/task
[2]: https://taskfile.dev
[3]: https://taskfile.dev/usage/#running-a-global-taskfile

fordneild:
if you try a command that isnt in your justfile, iirc just will searching up the file tree looking for one. I have a "global" justfile with general commands at ~/justfile and then project specific commands i put in a justfile at the root of the repository.

casey:
This seems reasonable to me, and pretty easy to implement. If you pass `--global` or `-g`, it looks for a justfile in `~/.justfile` (or maybe `~/justfile` or `~/.config/just/justfile`)

spartas:
> fordneild:
> if you try a command that isnt in your justfile, iirc just will searching up the file tree looking for one. I have a "global" justfile with general commands at ~/justfile and then project specific commands i put in a justfile at the root of the repository.

I didn't consider that `just` would continue searching up the tree until it found a match. There are a few things that I do that wouldn't work though.

In my Docker projects, I typically have a `default` recipe (it's the first recipe in my `justfile`) that runs `docker compose ps`. It doesn't make sense to put this (context-specific recipe) in a global `justfile`, so I also have a `default` recipe in the global file that runs `docker compose ls` to list running compose projects across the whole system.

Secondly, I'm often working under mounted filesystems or subtrees that aren't descendants of `$HOME`.

> casey:
> This seems reasonable to me, and pretty easy to implement. If you pass `--global` or `-g`, it looks for a justfile in `~/.justfile` (or maybe `~/justfile` or `~/.config/just/justfile`)

[...]

Issue #1344, opened 2022-09-16
Feature request: categories/sections


leviska:
I think it would be good to have some sections in `just -l`
For example, output of my file is
```
Available recipes:
    default                     # list all recipes
    build cmd                   # build go binary locally
    docker-build cmd            # build single cmd docker image and tag is as cmd
    docker-run cmd *flags       # run image with tag based on cmd
    docker-build-run cmd *flags # docker-build and then docker-run
    compose-build cmd           # build and push a compose file
    deploy cmd context_name="master" # deploy compose file
    compose-build-deploy cmd context_name="master" # compose-build and then deploy
```
And I would like to be able to make something like
```
Available recipes:
    just:
        default                     # list all recipes
    build:
        build cmd                   # build go binary locally
    docker:
        docker-build cmd            # build single cmd docker image and tag is as cmd
        docker-run cmd *flags       # run image with tag based on cmd
        docker-build-run cmd *flags # docker-build and then docker-run
    deploy:
        compose-build cmd           # build and push a compose file
        deploy cmd context_name="master" # deploy compose file
        compose-build-deploy cmd context_name="master" # compose-build and then deploy
```

In file it could look something like

```make
# just-category: build
...build recipes
# just-category: docker
...docker recipes
```

or 

```make
# build go binary locally #category:build
build cmd:
...
# build single cmd docker image and tag is as cmd #category:docker
docker cmd:
...
```

Or whatever you like :) 

Or may be even make nested recipes, so you could call something like
```bash
just build # same stuff
just docker build # instead of `just docker-build`
just docker run # instead of `just docker-run`
just deploy # same
just deploy build # instead of `just compose-build`
```

runeimp:
I like the idea but it smacks of namespacing and once Just adds module support the category/section stuff would likely further be altered in the list output. May not be a big deal, no idea, but definitely seems like something to consider in that context. Don't mind me. Just thinking out-loud.  👼

casey:
I'm inclined to agree with @runeimp, although since module support is far away, I wouldn't mind low-impact things that made large justfiles easier to navigate.

How about some kind of section marker that you can put in a justfile that if used gets added to `--list` output and disables alphabetic sorting of `--list`.

So you can do:

```
!! A

foo:

!! B

bar:

!! C

baz:
```

And then `just --list` looks like:

```
$ just --list
Available recipes:
  # A
  foo
  # B
  bar
  # C
  baz
```

[...]

PR #1842, opened 2024-01-13
Recipe grouping

neunenak:
Implements recipe grouping feature, see https://github.com/casey/just/issues/1776


neunenak:
I'm currently using the `-g` flag (for `--groups`) in this draft PR. This issue: https://github.com/casey/just/issues/1648 asks for a global feature, for which it would also be convenient to use the `-g` flag. I'm not sure which item should get the `-g` flag, and if so what would be a better alternate flag letter for either.

casey:
Nice!

You might want to rebase, I just merged a PR which allows overriding the confirm prompt using `[confirm("prompt")]`, so there's some churn in attribute stuff.

I think `[group: foo]`, as you have it, is definitely the nicest syntax. If we have it though, we should also have `[confirm: "prompt"]` too otherwise users will likely get confused.

casey:
I think probably the global flag should get `-g`, since I think it might be more commonly used. The easiest thing to do is probably to not assign short flags to either and wait for someone to complain that one should have a short flag.

neunenak:
> Nice!
> 
> You might want to rebase, I just merged a PR which allows overriding the confirm prompt using `[confirm("prompt")]`, so there's some churn in attribute stuff.
> 
> I think `[group: foo]`, as you have it, is definitely the nicest syntax. If we have it though, we should also have `[confirm: "prompt"]` too otherwise users will likely get confused.

Solid, I was debating whether `[attr: argument]` or `[attr(argument)]` was a better syntax - if there's another attr now that uses `()` function-like syntax, that seems like the one to go with then.

neunenak:
> I think probably the global flag should get `-g`, since I think it might be more commonly used. The easiest thing to do is probably to not assign short flags to either and wait for someone to complain that one should have a short flag.

Makes sense. This is a feature that I personally want and expect to use reasonably frequently, so I'd want to add a short flag. Maybe `-r`?

neunenak:
> Nice!
> 
> You might want to rebase, I just merged a PR which allows overriding the confirm prompt using `[confirm("prompt")]`, so there's some churn in attribute stuff.
> 
> I think `[group: foo]`, as you have it, is definitely the nicest syntax. If we have it though, we should also have `[confirm: "prompt"]` too otherwise users will likely get confused.

Do we want to keep the `[confirm(prompt)]` syntax around and allow both, or get rid of it? The version you just pushed has the `()` style, but it would probably be okay to get in a quick change that changes `[confirm]` to use the `:`-syntax, push a new 1.23.1 version that changes this before too many people start writing justfiles with the `()` syntax, then get the rest of groups in later.

casey:
> Makes sense. This is a feature that I personally want and expect to use reasonably frequently, so I'd want to add a short flag. Maybe `-r`?

Oh, hmmm, I see. `--groups` will list groups and organize by group when used with `--list`. So the two uses are `just --groups` and `just --list --groups`. What if we made `--list` organize by group by default? If there are groups in the justfile, I assume that people will want to see them.

Then you wouldn't have to do `just --list --groups`, which is a super long and annoying invocation. And maybe `just --groups` is less common, so it doesn't need a short flag?

> Do we want to keep the `[confirm(prompt)]` syntax around and allow both, or get rid of it?

It made it into a release, so we should keep it. But I think the other reason for keeping it is that we might want to have attributes that take multiple arguments. So `[attr: arg]` is a convenient shorthand when there's a single argument, and `[attr(arg)]` can be use with one or more arguments.

neunenak:
> > Makes sense. This is a feature that I personally want and expect to use reasonably frequently, so I'd want to add a short flag. Maybe `-r`?
> 
> Oh, hmmm, I see. `--groups` will list groups and organize by group when used with `--list`. So the two uses are `just --groups` and `just --list --groups`. What if we made `--list` organize by group by default? If there are groups in the justfile, I assume that people will want to see them.
> 
> Then you wouldn't have to do `just --list --groups`, which is a super long and annoying invocation. And maybe `just --groups` is less common, so it doesn't need a short flag?
> 
> > Do we want to keep the `[confirm(prompt)]` syntax around and allow both, or get rid of it?
> 
> It made it into a release, so we should keep it. But I think the other reason for keeping it is that we might want to have attributes that take multiple arguments. So `[attr: arg]` is a convenient shorthand when there's a single argument, and `[attr(arg)]` can be use with one or more arguments.

I like the idea of listing by groups by default. A justfile without any recipes with a `[group]` will have the same listing behavior as now so it won't break existing behavior. I agree that listing the group names themselves is uncommon and doesn't need a short option. And yeah I think the convention where `[attribute: single-arg]` is shorthand for `[attribute(single-arg)]` that doesn't apply to hypothetical multi-argument attributes is fine.

neunenak:
Alright, got this working like so:

justfile:
```
# Comment
[group: alpha]
a:

[group(alpha)]
[group(beta)]
b:

c:

[group:"multi word group"]
d:

[group("alpha")]
e:

[group: "beta"]
[group(alpha)]
f:
```

`just --list` output:
```
Available recipes:
(no group)
    c
[group: alpha]
    a # Comment
    b
    e
    f
[group: beta]
    b
    f
[group: multi word group]
    d
```

This code deliberately allows you to put a recipe in as many groups as you want, and prints a recipe listing once per group it's in. I think these are good defaults, and if they're not what people want in all cases I think it'd be fine to add additional flags to `--list` that control this.

neunenak:
https://github.com/casey/just/issues/1344 is also probably taken care of by this PR

neunenak:

jermatic:
when merge?

casey:
I'm super busy with another project, so just is sadly being neglected right now. Honestly it'll probably take at least a month before I can start grinding down my GitHub notification backlog, which is insanely large at this point.

epicserve:
@neunenak,

In reading the docs that were added, I'm wondering if you would consider the following output. It's a little more readable to me and is based off of how Django does it.

```
$ just --list

Available recipes:

[no group]
    email-everyone

[lint]
    cpp-lint
    js-lint
    rust-lint

[rust jobs]
    rust-lint

```

Instead of:
```
$ just --list
Available recipes:
(no group)
    email-everyone # Not in a group
[group: lint]
    cpp-lint
    js-lint
    rust-lint
[group: rust jobs]
    rust-lint
```

epicserve:
@neunenak,

Thank you for the update! I'm really looking forward to this new feature getting added!

neunenak:
> @neunenak,
> 
> Thank you for the update! I'm really looking forward to this new feature getting added!

Thanks! I'm gonna ping @casey about this PR again, hopefully this can get merged pretty quickly.

src/attribute.rs
```diff
@@ -16,43 +17,49 @@ pub(crate) enum Attribute<'src> {
...
-
-    if let Some(argument) = self.argument() {
-      write!(f, "({argument})")?;
+    let attr_name = self.name();
+    match self {
+      Self::Confirm(Some(prompt)) => write!(f, "{attr_name}('{prompt}')"),
+      Self::Group { name } => {
+        let use_quotes = name.contains(char::is_whitespace);
+        let mq = if use_quotes { "\"" } else { "" };
+        write!(f, "{attr_name}({mq}{name}{mq})")
```
casey: This will not work in general. Single quoted strings, and triple-delimited strings might have content that can't be contained in a single `"`. I think the simplest thing is to punt on accepting bare words, and only accept string literals for the time being.

GRAMMAR.md
```diff
@@ -111,6 +111,7 @@ recipe        : attributes* '@'? NAME parameter* variadic? ':' dependency* body?
 attributes    : '[' attribute* ']' eol
 
 attribute     : NAME ( '(' string ')' )?
+              | NAME ':' string
```
casey: I'd prefer to leave this syntax for a follow up PR, to keep this one simple.

src/recipe.rs
```diff
@@ -1,4 +1,6 @@
 use super::*;
+use std::collections::HashSet;
+use std::process::{ExitStatus, Stdio};
```
casey: These imports can be combined.

casey (changes requested):
Did a little review, see comments.

GRAMMAR.md
neunenak: Removed the code that did this parsing from the PR

src/attribute.rs
neunenak: Makes sense. I modified `Group` to take a `StringLiteral`.

casey:
I started looking at this, and to be honest I don't really like the refactor into multiple functions. I don't really mind long functions, and generally don't like functions which are defined and only used once. I think I prefer being able to read a long linear function, and not indirect through other functions. Also, it makes the review harder, because it's hard to tell what changed and what didn't. Can you inline it back into a single function in `Subcommand::list`?

casey (approved):
LGTM!

Some notes:

- On entries of vecs, you can do `.or_default().push(x)` which saves some characters.
- I had to indent the group names because of a conflict with the way modules are printed. Without indenting the group names, it isn't clear that the modules aren't part of the groups.

PR #1845, opened 2024-01-13
Add `assert` expression

### LINKED ISSUE #1657: Make the `else` branch of a conditional optional
I have the following recipe:

```just
@bump package:
    {{ if path_exists("Formula" / package + ".rb") == "false" { error("no formula for package " + package) } else { "" } }}
```

Which works great! But, it gives an error if I don't include a dud `else` block. 

```
error: Expected keyword `else` but found `'}}'`
  |
7 |     {{ if path_exists("Formula" / package + ".rb") == "false" { error("no formula for package " + package) } }}
```

It would be nice if it were optional

de1iza:
This PR adds an `assert (CONDITION, ERROR)` expression, which takes a boolean `CONDITION`, and errors out with the value of `ERROR` if `CONDITION` is false. 

I extracted `lhs`,  `rhs` and `operator` out of `Conditional` into a separate struct `Condition`, as it is now shared with `Assert` expression.
 
Fixes #1657.


de1iza:
Hi @casey, I was wondering if you could review my PR.

casey:
Nice work! Made some minor changes, but otherwise looks good to me!

yakimant:
Didnt find any documents on `assert`at https://just.systems/man

PR #1846, opened 2024-01-14
Global justfile

### LINKED ISSUE #1648: Support for global justfiles

neunenak:
Allow the `-g` or `--global` flag to reference a global `justfile` in the user's home directory.

Resolves https://github.com/casey/just/issues/1648


neunenak:
One thing that's probably worth discussing is whether the default justfile path should be `$HOME/justfile` , `$HOME/.justfile`, or `$XDG_CONFIG_DIR/just/justfile`. Using bare dotfiles in the homedir seems like it's falling out of fashion these days, and for good reason, so really it's just a choice between `$HOME/justfile` and `$XDG_CONFIG_DIR/just/justfile`.  Or if it should search for both (and in what order) and just use the first one it finds.

casey:
I personally prefer dotfiles in my homedir, since it's easier to edit than the XDG config dir, which on a mac is in `~/Library/Application Support` which is very verbose. Also, the cost of checking an additional location is minimal. So I'd do `~/.justfile`, `~/justfile`, and then `$XDG_CONFIG_DIR/just/global.just`, although I'm not sure in which order.

neunenak:

casey:
Some notes:

- I removed the ability to initialize the global justfile. I think it has a bug, where if the config directory doesn't exist, it will throw an I/O error.

- This could use a test, which sets XDG_CONFIG_HOME to a tempdir with a justfile

- What do you think about calling the global justfile in the config dir `justfile` instead of `global.justfile`?

- Why not use `dirs::config_dir()` instead of looking up `XDG_CONFIG_HOME` manually?

neunenak:
> Some notes:
> 
>     * I removed the ability to initialize the global justfile. I think it has a bug, where if the config directory doesn't exist, it will throw an I/O error.

Might be nice to have the ability to initialize a global justfile, but that's additional functionality that can come in a subsequent PR. 
> 
>     * This could use a test, which sets XDG_CONFIG_HOME to a tempdir with a justfile

Makes sense, I'll try to add this shortly (probably not tonight). 

>     * What do you think about calling the global justfile in the config dir `justfile` instead of `global.justfile`?

Fine by me.

> 
>     * Why not use `dirs::config_dir()` instead of looking up `XDG_CONFIG_HOME` manually?

Ah I didn't realize there was already code that this.

neunenak:
@casey looks like the macos CI job is failing with no error message. I'm not sure if this is a problem with these changes or CI.

neunenak:
> @casey looks like the macos CI job is failing with no error message. I'm not sure if this is a problem with these changes or CI.

I think what I figured out was that setting `XDG_CONFIG_DIR` in the test for this functionality only works on linux, so I disabled the test for other OSs. Not sure if there's a way to fake the configuration directory for those OSs.

casey:
> I think what I figured out was that setting `XDG_CONFIG_DIR` in the test for this functionality only works on linux, so I disabled the test for other OSs. Not sure if there's a way to fake the configuration directory for those OSs.

I think it's fine not to test other OSs, since the only way something could go wrong is if there was a bug in dirs.

casey:
LGTM! I made the flag `--global-justfile`, since there's already `-g` for short, and `--global` seemed ambiguous. I also made it conflict with `--justfile` and `--working-directory`. Support for `--working-directory` can be added in a future PR, since they could work together.

PR #1847, opened 2024-01-15
Update clap to version 4 (not current)

neunenak:
Update clap to version 4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,6 +21,54 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.7"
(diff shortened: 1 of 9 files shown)


casey:
The big lift here is going to be completion scripts. We add a bunch of hacky patches to the clap 2 generated completion scripts, and if they've changed a lot, those patches will need to be ported to the new completion scripts.

casey:
I just took a look at the diff, and I think it would be better to first land a PR which upgrades clap to 4.0, but doesn't use the declarative, derive-based API, and then in a follow-up PR, switch to using the derive-based API.

Issue #1799, opened 2023-12-29
Fetch module source files using Git or HTTP


casey:
A few people would like to be able to load module source files from git or using HTTP.

I think that this probably isn't a good idea, but I wanted to make this issue for discussion.

The feature would work something like this, with some made up syntax:

```
mod foo http('https://foo.com/foo.just')
```

Just would then fetch the file `https://foo.com/foo.just` and use it as the source for module `foo`.

The main reason that I think this would be a bad idea is the complexity of the feature, the difficulty of getting it right, and the ease of workarounds.

Some hard questions regarding how the feature would work:

- Would `just` automatically fetch files, or would you have to tell it to explicitly? The latter is probably better, since you might not want `just` to fetch and execute files remotely.
- Which protocols would `just` support? HTTP is pretty easy, but a common ask is for git support too.
- How frequently would `just` update module source files? We'd probably need a subcommand for re-fetching new source files.

And the workaround looks something like this:

```
mod? foo

update:
  curl -O https://foo.com/foo.just
```

Then, you could run `just update` to fetch the source file, which would then be loaded the next time you ran `just`. Using git would looks something like:

```
mod? foo

update:
  #!/usr/bin/env bash

  if [[ -d foo ]]; then
    cd foo
    git pull
  else
    git clone git@foo.com/foo.git
  fi
```

Then, you could run `just update` to fetch the git repo `foo`, or update it if it already exists.

This allows the user to get whatever behavior they want, so I'm not sure a dedicated feature would be an improvement.

laniakea64:
I agree with @casey - the workaround is simpler implementation, more versatile for justfile authors, and easier for end-users to read and understand.

Also, adding any network-using feature to `just` would impact `just`'s existing compatibility with offline dev environments.  It would mean offline dev environments start needing special consideration.  Usually, when a networking-enabled tool accommodates offline dev environments, the environments need to make a change to use the accommodation (e.g. setting an env var, passing additional command-line switch, and/or finding alternative ways to run the online part separately and transfer/save the result).  And in the case of `just`, that could potentially get hairier when considering recipes that invoke another `just` instance.  An individual justfile doesn't need to think about any of this.

vtmocanu:
Ok, agree with your points, one question comes to mind.
Let's say multiple users use this system. It's not clear to me, i have not played with modules just yet tbh, how they would import the main just modules, because they have different PATHs.
eg:
- one user may keep the modules in ~/just_modules
[...]

PR #1849, opened 2024-01-15
Support defining external justfile locations through env var

pprkut:
This extends just so you can define multiple locations from which it tries to import other justfiles. This allows, for example, to put shared justfiles in a user defined global location once, rather than having to manage duplicates in every repo they are used in or deal with symlinks.

Example use: `JUSTFILE_PATH="/path/to/external/justfiles/:./" just --list`

This is my preferred solution for #1580 and while currently only implemented for imports, should work just as well for modules too, which might provide additional input or discussion points for #1799 
I'm not set on the environment variable name, open for other suggestions.

Taking things slow as I'm fairly new to rust, but I think the idea should be clear :slightly_smiling_face:


casey:
Thanks for the PR! I actually think that expanding environment variables in paths, e.g., `mod foo '$DIR/foo.just'` might be a better approach here. A few thoughts:

- If environment variables are in paths, then we can give a good error if the environment variable isn't set. Whereas if `JUSTFILE_PATH` isn't set, and that's where the module is, we can only give the generic module not found error.
- There appears to be a very widely used crate for performing both `~` expansion and variable expansion called [shellexpand](https://docs.rs/shellexpand/latest/shellexpand/). So adding this would just require adding that crate, and passing module paths into it.
- `shellexpand` supports fallbacks, in the form of `${VAR:-/other/path}`.
- We should add this for `mod` statements first. My thinking here is that modules are still unstable, so breaking changes are acceptable, whereas imports are stable, so we can't add breaking changes. Technically, expanding variables in paths would be a breaking change, however, using `$` or `~` in paths is so uncommon that I think it would be okay to do this, even for imports. However, doing `mod` first gives us a little more room to experiment.
- Users can use different variables for different imports, instead of having a single variable for all imports.
- Users can use variables in the middle of a path, i.e., `/var/lib/${SERVICE}/mod.just`.

What do you think?

pprkut:
Expanding the variables in the path was one of my suggestions in #1580, so that's fine with me too :slightly_smiling_face: 
The reason I went with an external path variable is that is was quicker for me to implement and get going, and provided the added benefit of being backwards compatible.

I get your argument for starting with `mod` first. I'm just not using modules so for my testing I'd have to do it for imports anyway. However, I don't mind if, once I have a good grip on the changes, the `mod` version of it gets merged first (maybe through a separate PR). 

I'll have a look at `shellexpand` and see where I get. Thanks for the input! :slightly_smiling_face:

tests/modules.rs
```diff
@@ -685,6 +685,47 @@ fn module_paths_beginning_with_tilde_are_expanded_to_homdir() {
...
+    .test_round_trip(false)
+    .arg("--unstable")
+    .status(EXIT_FAILURE)
+    .stderr(
+      "
+      error: Source path for foo contains unresolved variable FOO.
+       ——▶ justfile:1:5
+        │
+      1 │ mod foo '$FOO/mod.just'
+        │     ^^^
```
pprkut, line 733: The pointers here would be nicer on the path, but I couldn't figure out how to get to a token from `relative`

src/compiler.rs
```diff
@@ -78,11 +86,16 @@ impl Compiler {
...
+              Ok(value) => value,
+              Err(error) => return Err(Error::UnresolvedVariableInImport { variable: error.var_name, path: *path }),
+            };
+
             let import = current
               .path
               .parent()
               .unwrap()
-              .join(Self::expand_tilde(&relative.cooked)?)
+              .join(&*expanded_path)
```
pprkut, line 92: I somehow doubt `&*` is the "good" thing to do here, even if rust was suggesting it like that :sweat_smile: 
Hints appreciated :slightly_smiling_face:

src/compiler.rs
```diff
@@ -48,7 +51,12 @@ impl Compiler {
             let parent = current.path.parent().unwrap();
 
             let import = if let Some(relative) = relative {
-              let path = parent.join(Self::expand_tilde(&relative.cooked)?);
+              let expanded_path = match full(&relative.cooked) {
+                Ok(value) => value,
+                Err(error) => return Err(Error::UnresolvedVariableInModule { variable: error.var_name, module: *name }),
+              };
```
casey: You can do something like this:
```suggestion
              let expanded_path = full(&relative.cooked).map_err(|error| Error::UnresolvedVariableInModule { variable: error.var_name, module: *name }))?
```

src/compiler.rs
```diff
@@ -78,11 +86,16 @@ impl Compiler {
             optional,
             path,
           } => {
+            let expanded_path = match full(&relative.cooked) {
+              Ok(value) => value,
+              Err(error) => return Err(Error::UnresolvedVariableInImport { variable: error.var_name, path: *path }),
```
casey: See my other comment on using `map_err`.

src/compiler.rs
casey: What's the error you're getting otherwise?

tests/modules.rs
casey: Relative is an Option<StringLiteral>, but in the case of an error, you know there's a path, so you can unwrap it.

casey (changes requested):
See review comments!

src/compiler.rs
pprkut: With just `expanded_path` I get
```
error[E0277]: the trait bound `Cow<'_, str>: AsRef<std::path::Path>` is not satisfied
  --> src/compiler.rs:98:21
   |
98 |               .join(expanded_path)
   |                ---- ^^^^^^^^^^^^^ the trait `AsRef<std::path::Path>` is not implemented for `Cow<'_, str>`
   |                |
   |                required by a bound introduced by this call
   |
   = help: the following other types implement trait `AsRef<T>`:
             <Cow<'_, T> as AsRef<T>>
             <Cow<'_, std::ffi::OsStr> as AsRef<std::path::Path>>
note: required by a bound in `std::path::Path::join`
```
With `&expanded_path` I get the almost identical
```
error[E0277]: the trait bound `Cow<'_, str>: AsRef<std::path::Path>` is not satisfied
  --> src/compiler.rs:98:22
   |
98 |               .join(&expanded_path)
   |                ----  ^^^^^^^^^^^^^ the trait `AsRef<std::path::Path>` is not implemented for `Cow<'_, str>`
   |                |
   |                required by a bound introduced by this call
   |
   = note: required for `&Cow<'_, str>` to implement `AsRef<std::path::Path>`
note: required by a bound in `std::path::Path::join`
  --> /tmp/rustc-1.74.0-src/library/std/src/path.rs:2550:5
help: consider dereferencing here
   |
98 |               .join(&*expanded_path)
   |                      +

For more information about this error, try `rustc --explain E0277`.
```

With `*expanded_path` I get
```
error[E0277]: the size for values of type `str` cannot be known at compilation time
  --> src/compiler.rs:98:21
   |
98 |               .join(*expanded_path)
   |                ---- ^^^^^^^^^^^^^^ doesn't have a size known at compile-time
   |                |
   |                required by a bound introduced by this call
   |
   = help: the trait `Sized` is not implemented for `str`
note: required by a bound in `std::path::Path::join
```

tests/modules.rs
pprkut: It's already unwrapped. The difficulty I face is that `relative` is a `StringLiteral`, and I need a `Token` in the error handling to position the pointers. I don't know how to get back from the literal string to the token.
I thought about perhaps using the module token and iterating to the next token after (something like `module.token.next`), but I don't see an easy way to do it this way either :confused:

src/compiler.rs
pprkut: I think https://stackoverflow.com/questions/41273041/what-does-combined-together-do-in-rust cleared this up, and this makes sense now

tests/modules.rs
pprkut: @casey Any remaining ideas on this? Otherwise I keep it as is :slightly_frowning_face:

casey: The `mod` parser uses `parse_string_literal`, but there's also `parse_string_literal` token which returns both the string literal and the token, if you store the token in `mod` in addition to the string literal, you can use it in the error message.

casey:
I see some new commits. Is this ready for review?

casey:
I just landed #2055, which adds a a new kind of string literal. Strings prefixed with `x`, i.e., `x'$FOO'`, are now shell expanded, supporting `$VAR`, `${VAR}`, `~`, and `~USER` . This shell expansion is performed at compile time, which means that it can be used anywhere a string literal is accepted, including import paths, module paths, and settings. I think this is a more general solution to the problem here, since it works everywhere, and since it requires using `x`, it's opt-in and thus backwards compatible. Sorry to waffle on this! If people find it un-ergonomic to put `x` before import paths, we can consider doing it by default in module and import paths, as explored in this PR.

pprkut:
Cool! No worries, I'm just happy the feature works :slightly_smiling_face:

fabiomoratti:
The shell expansion feature is very handy for managing different sets of variable for different environments, thanks for implementing it!

Il might be worth mentioning in the documentation that the expansion supports fallbacks: I was not familiar with the fallback feature and it took me a while (well, a lot of time, until I stumbled on this thread and the 17 Jan post about the merits of `shellexpand`) to figure out that it's possible to use the extension also if the env variable is not set.

Something like:

>| Value | Replacement |
|------|-------------|
| `$VAR` | value of environment variable `VAR` |
| `${VAR}` | value of environment variable `VAR` |
**| `${VAR:-DEFAULT_VALUE}` | value of environment variable `VAR` or `DEFAULT_VALUE` if the VAR variable is not set |**
| Leading `~` | path to current user's home directory |
| Leading `~USER` | path to `USER`'s home directory |

In case someone is interested here is my strategy to manage multiple envs:

```justfile

foo_default := "FOO_0"
bar_default := "BAR_0"

# The "envs" folder contains one <env-name>.env for each environment with the appropriate variables:
# env-1.env:
# ENV_FOO=FOO1
# ENV_BAR=BAR1
# 
# env-2.env:
# ENV_FOO=FOO2
# ENV_BAR=BAR2
# 
# If the JUST_ENV variable is not set load the default.env file, if it's missing use the hard coded values above
set dotenv-filename := x'envs/${JUST_ENV:-default}.env'

foo := env_var_or_default('ENV_FOO', foo_default)
bar := env_var_or_default('ENV_BAR', bar_default)

# Use foo and bar variables in the recipes
```

casey:
@fabiomoratti Nice, I didn't know that `${VAR:-DEFAULT}` was supported syntax. I opened #2153 to mention it in the readme.

PR #1860, opened 2024-01-21
Add `blake3` and `blake3_file` hash functions

tgross35:
Blake3 is a hash algorithm that is pretty popular when working with larger files, and is also (in theory) closer to SHA-3 in security than SHA-2. I believe it is used by sccache and some other larger build systems.

This adds `blake3` and `blake3_file` functions. `blake3_file` is especially of note because it uses mmap and parallelization rather than reading the file into memory, so it should outperform `sha256_file` quite a bit (this is the same function used by the `b3sum` utility).

I need to rebase since I did this a while ago, but I may as well get your thoughts here.


casey:
Love it, this is super cool.

tgross35:
Awesome! I rebased so this should be all set for a look.

casey:
Nice, LGTM! I used `blake3::hash` instead of a hasher in in `blake3()`, since it's one and done and doesn't need incremental updates. I also changed uses of `path.to_str` to `path.display()`. Since paths are arbitrary bytes, they can't necessarily be converted to a string, and path.display returns something that represents a path but has a display implementation, so it can be printed. It looks like we made the same mistake in the `sha256` functions, so I fixed those too.

casey:
I also added `<sup>master</sup>` annotations, so people know this hasn't made it into a release yet.

tgross35:
Awesome, thanks!

PR #1863, opened 2024-01-21
Support "::" in Bash completions

crdx:
Modules use `::` as the separator, which is considered a wordbreak character in Bash. This means they can't be completed by default, as colon is considered the end of a completion.

The (very) bad way to solve this is by globally modifying `COMP_WORDBREAKS` to remove the colon character. Not only is this bad because it can have unintended side-effects, but git's completion script actually adds the colon back if it detects it's not present.

So Bash made available the `_get_comp_words_by_ref` and `__ltrim_colon_completions` functions for this purpose. They're not present on every platform, in which case we fall back to the original method.


crdx:
... oh. 

That completion file is generated by clap. Right.

crdx:
Gonna be fun making those changes using only search/replacements in `BASH_COMPLETION_REPLACEMENTS`.

crdx:
Should be good now.

casey:
Nice, LGTM!

PR #1867, opened 2024-01-24
GitHub Action to `just test` on macOS, Ubuntu, and Windows

cclauss:
Demonstrate how to install and use just in GitHub Actions on the three major operating systems without needing third-party GitHub Actions such as https://github.com/casey/just#github-actions

Developers may be interested to run the same `just` commands that they use locally on continuous integration platforms such GitHub Actions.  Every time that a contributor creates a pull request, `just test` could be run on the three major operating systems to provide feedback to both the contributor and reviewers that tests are passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,7 +305,7 @@ export PATH="$PATH:$HOME/bin"
 just --help
 ```
 
-Note that `install.sh` may fail on GitHub actions, or in other environments
+Note that `install.sh` may fail on GitHub Actions, or in other environments
 where many machines share IP addresses. `install.sh` calls GitHub APIs in order
 to determine the latest version of `just` to install, and those API calls are
 rate-limited on a per-IP basis. To make `install.sh` more reliable in such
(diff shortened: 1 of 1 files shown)


casey:
This is dope! I'm not sure we actually want to run it though as part of our own actions runs, since it will add another job when you make a PR, etc, which is already a little noisy with five jobs. How about moving this to a contrib directory? It's pretty common to have a directory called `contrib`, which contains extras like these. I just merged #1870, which creates a contrib directory. It can be called something like `contrib/github-actions-install-just.yaml`.

cclauss:
Can we please run it ___once___ (to make sure it passes) before we move it into the docs somewhere?

If you merge one of the other two PRs then I `update branch` (below) then this action will run and we can see if all the polyglot languages are present on all three OSes.

casey:
Yup, for sure. I think it just ran and it looks like it was successful.

cclauss:
OK...  Should we move this code into https://github.com/casey/just/blob/master/README.md#github-actions or should we put it into a separate markdown file?  Other ideas?

casey:
I think we should move the full yaml file into `contrib/github-actions-install-just.yaml`, and link to it in the `#github-actions` section of the readme. Since it's probably faster to install `just` using a package manager that's already present on the actions runners, I think we should link to it at the beginning of the `#github-actions` section, before the two alternative methods.

cclauss:
The file that `CI/pages` is complaining about will exist only after this pull request is merged.

cclauss:
We can add the link from README.md in a separate PR to placate `CI/pages`.

casey:
Okay, sorry to waffle, but looking at the workflow, I think it can be made short enough that we should just stick it in the readme, which also means we don't have to worry about a broken link.

By avoiding using a matrix, we can boil it down to:

```yaml
name: install-just
on: [pull_request, push]
jobs:
  windows:
    runs-on: windows-latest
    steps:
    - run: choco install just
  macos:
    runs-on: macos-latest
    steps:
    - run: brew install just
  ubuntu:
    runs-on: ubuntu-latest
    steps:
    - run: sudo snap install --edge --classic just
```

Which will be pretty easy for users to insert into their existing workflow. What do you think?

cclauss:
The working script is 3 lines longer than the matrix script and does not set fast-fail to false but it is easier to understand and most visitors will only be interested to use the Ubuntu bit.

casey:
Nice, looks good to me! Thank you!

PR #1868, opened 2024-01-24
Keep GitHub Actions up to date with GitHub's Dependabot

cclauss:
This will automate the creation of pull requests like
* pennersr/django-allauth#3615

And avoid the warnings like at the bottom of https://github.com/casey/just/actions/runs/7603050227

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
```
casey: This doesn't need quotes, since it doesn't contain any special characters.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+  directory: "/"
```
casey: This doesn't need quotes, since it doesn't contain any special characters.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+  groups:
+    GitHub_Actions:
+      patterns:
+        - "*"  # Group all Actions updates into a single larger pull request
```
casey:
```suggestion
        - "*"  # open a single pull request to update all actions
```

casey (changes requested):

casey:
Thanks for these PRs! They're awesome. I added a review with some very small nits, take a look.

.github/dependabot.yml
cclauss:
```suggestion
- package-ecosystem: github-actions
```

.github/dependabot.yml
cclauss:
```suggestion
  directory: /
```

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: /
+  groups:
+    GitHub_Actions:
+      patterns:
+        - "*"  # open a single pull request to update all actions
```
cclauss, line 8: Are the quotes required here?

casey: They are, `*` is a special character in YAML. You can use `&` and `*` to declare and reference properties to de-duplicate repeated values.

This:

```yaml
myprop:
  uid: &id XXX
myprop1:
  id: *id
```

Is equivalent to this:

```yaml
myprop:
  uid: XXX
myprop1:
  id: XXX
```

casey (approved):
LGTM!

cclauss:
Generated:
* #1871

PR #1869, opened 2024-01-24
GitHub Actions: Remove deprecated set-output command

cclauss:
The GitHub `set-output` command is now deprecated which results in numerous warnings like at the bottom of https://github.com/casey/just/actions/runs/7603050227

This pull request replaces the old syntax by using the environment file `$GITHUB_OUTPUT` as recommended in the deprecation notice.
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands

Related to #1868 

These tests are modified:
* junk_is_other()
* valid_version_is_release()
* valid_version_with_trailing_characters_is_other()
* valid_version_with_lots_of_digits_is_release()


casey:
Awesome, great PR! I've been procrastinating dealing with these errors. The output of `ref-type` is used in [release.yaml](https://github.com/casey/just/blob/dcca9681e2993f6662a9774e7cffd10397bfa4a4/.github/workflows/release.yaml#L73), so I think that needs to be updated too.

.github/workflows/release.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
 
     - name: Ref Type
       id: ref-type
-      run: cargo run --package ref-type -- --reference ${{ github.ref }}
+      run: cargo run --package ref-type -- --reference ${{ github.ref }}  # Please provide a suggestion
```
cclauss: I am unsure what changes need to be made.  Can you make a ___\`\`\`suggestion___ or point me to some docs?

casey: I think you just need to do:

```
    - name: Ref Type
      id: ref-type
      run: cargo run --package ref-type -- --reference ${{ github.ref }} >> $GITHUB_OUTPUT
```

Once you change it, I can run the release workflow on a tag, to make sure it works.

cclauss:
```suggestion
      run: cargo run --package ref-type -- --reference ${{ github.ref }} >> $GITHUB_OUTPUT
```

casey:
Don't worry about the lint failure, I'll fix that.

cclauss:
[Linter: Shorter lines are not wrapped](https://github.com/casey/just/pull/1869/commits/c68b9210f3f751af8d81a9009cdfd63268660353)

casey:
Nice, thank you!

cclauss:
My pleasure.  Just is a cool tool!!

PR #1874, opened 2024-01-25
GitHub Actions: Remove actions-rs/toolchain which is end-of-life

### LINKED ISSUE #1873: GitHub Actions: actions-rs/toolchain is end-of-life and should be replaced
https://github.com/casey/just/blob/ebead95ac990db1257c7a46198e43c5bb0def593/.github/workflows/ci.yaml#L26
* https://github.com/actions-rs/toolchain is archived.
* https://docs.github.com/en/actions/automating-builds-and-tests currently has no Rust solution.
* https://github.com/actions-rust-lang/setup-rust-toolchain seems to be a full-featured solution.
* GitHub Actions base images have rustup v1.26.0,  cargo v1.75.0, rustc v1.75.0, and rustfmt 1.7.0 pre-installed.

cclauss:
Fixes: #1873 
* #1873 

Use the version of `cargo` that is pre-installed in the GHA base image.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,12 +22,6 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
-    - name: Install Rust Toolchain Components
-      uses: actions-rs/toolchain@v1
-      with:
-        components: clippy, rustfmt
-        toolchain: stable
-
     - uses: Swatinem/rust-cache@v2
 
     - name: Clippy
@@ -58,12 +52,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-
-    - name: Install Rust Toolchain Components
-      uses: actions-rs/toolchain@v1
-      with:
-        profile: minimal
-        toolchain: stable
 
     - uses: Swatinem/rust-cache@v2
 
@@ -117,13 +105,6 @@ jobs:
         takeown /F C:\Windows\System32\bash.exe
         icacls C:\Windows\System32\bash.exe /grant administrators:F
         del C:\Windows\System32\bash.exe
-
-    - name: Install Rust Toolchain Components
-      uses: actions-rs/toolchain@v1
-      with:
-        components: clippy, rustfmt
-        override: true
-        toolchain: stable
 
     - uses: Swatinem/rust-cache@v2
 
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -49,13 +49,6 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
-    - name: Install Rust Toolchain Components
-      uses: actions-rs/toolchain@v1
-      with:
-        profile: minimal
-        target: ${{ matrix.target }}
```
cclauss, line 56: This line bugs me...  How can we communicate the profile and target to the Rust Toolchain?

.github/workflows/ci.yaml
```diff
@@ -118,13 +106,6 @@ jobs:
         icacls C:\Windows\System32\bash.exe /grant administrators:F
         del C:\Windows\System32\bash.exe
 
-    - name: Install Rust Toolchain Components
-      uses: actions-rs/toolchain@v1
-      with:
-        components: clippy, rustfmt
-        override: true
```
cclauss, line 125: How can we communicate override to the Rust Toolchain?

.github/workflows/release.yaml
casey: Is `rustup` installed on the base image? If so, can we use `rustup` to install the correct toolchain?

.github/workflows/ci.yaml
casey: The `override` part? I'm not 100% sure it's being used, so we could just see if it works without it.

.github/workflows/release.yaml
cclauss: Yes, it is.  See the last line of #1873.  I know little about Rust tooling so you probably need to take over this pull request to get the `release` job finished.

It is cool to see that the `ci` job just works.

.github/workflows/ci.yaml
cclauss: The `ci` job is green ✅

casey:
Installing the right toolchain in `bin/package` turned out to be super easy. Great PR! Removing a dependency is always awesome, and this should make our CI job faster.

PR #1875, opened 2024-01-25
Fix link to package in Gentoo repo

vozbu:
Some packages were recently moved from sys-devel to dev-build category. The Just too.


casey:
Thank you!

Issue #1861, opened 2024-01-21
Cached recipes


tgross35:
For quite a few CMake projects, I have a separate recipes for configuration and build, with `configure` being a dependency of `build`. Normally the configuration does not need to be rerun unless variables change, and I would prefer it runs as little as possible because it can sometimes take longer than the build.

Currently I have something a bit messy that stores a hash of all captured variables to check if anything changed:

```just
# Configure Cmake
configure build-type="Debug" projects="clang":
	#!/bin/sh
	# Hash all configurable parts 
	hash="{{ sha256(source_dir + build_dir + build-type + install_dir + projects + linker_arg) }}"
	if [ "$hash" = "$(cat '{{config_hash_file}}')" ]; then
		echo "configuration up to date, exiting"
		exit
	else
		echo "config outdated, rerunning"
	fi

	printf "$hash" > "{{config_hash_file}}"

	cmake "-S{{source_dir}}/llvm" "-B{{build_dir}}" \
		-G Ninja \
		-DCMAKE_C_COMPILER_LAUNCHER=sccache \
		-DCMAKE_CXX_COMPILER_LAUNCHER=sccache \
		-DCMAKE_EXPORT_COMPILE_COMMANDS=true \
		"-DCMAKE_BUILD_TYPE={{build-type}}" \
		"-DCMAKE_INSTALL_PREFIX={{install_dir}}" \
		"-DLLVM_ENABLE_PROJECTS={{projects}}" \
		"{{linker_arg}}"
```

My suggestion is to add a way to do this by default. The above would become:

```just
[cached]
configure build-type="Debug" projects="clang":
	cmake "-S{{source_dir}}/llvm" "-B{{build_dir}}" \
	# ...
```

And Just would need to do the following:

- Evaluate all captures in the recipe
- Create a deterministic hash of the captures (and maybe also environment variables?)
- If never run before, store this information in the cache directory. Roughly
  ```json5
  {
    "cached_recipes": [
      { "path": "/home/user/project/justfile", "recipe": "configure", "hash": "09ca7e4e...", "last_run": "2024-01-21T08:40:52Z" },
      // ...
    ]
  }
  ```
- If an entry for that file and recipe already exists, compare the hash. Skip if it is the same

A more flexible alternative is to have the user specify what gets set as a cache key. This would be easier for Just to implement too, but is less user friendly.

```just
[cache_keys(source_dir, build_dir, build-type, install_dir, projects, linker_arg, `$SOME_ENV`)]
[...]

Issue #1685, opened 2023-10-03
Track if tasks are up-to-date based on file timestamps or checksums


Nezteb:
Task/taskfile supports [several ways of preventing unncessary work](https://taskfile.dev/usage/#prevent-unnecessary-work), one of which is [by fingerprinting locally generated files and their sources](https://taskfile.dev/usage/#by-fingerprinting-locally-generated-files-and-their-sources).

If `just` could support this in some capacity, I think it'd help improve `just`'s user experience. 😄

laniakea64:
While waiting for `just` dev to respond, here's a possible workaround example: a recipe `foo` that only needs to run if files `a.txt` and `b.txt` have different checksums since last run -
```
foo:
	#!/bin/bash
	job=foo
	sha256sums={{quote(justfile_directory() / '.sha256sums-') + '"$job"'}}
	dependencies='a.txt b.txt'
	if [[ -f "$sha256sums" ]] && sha256sum --status -c "$sha256sums" ;then
	  echo "Job '$job' is up-to-date"
	  exit 0
	fi
	sha256sum $dependencies > "$sha256sums"
	echo "Running job '$job'..."
```
(This example assumes files `a.txt` and `b.txt` always exist.)

casey:
I think this is a dupe of #867, have a look.

qtfkwk:
I created https://crates.io/crates/mkrs primarily for this reason (and to write the *Makefile* in markdown); it seems like the same approach may not fit in just's design. Mkrs distinguishes between file and non-file targets, and runs the recipe for a file target if `-B` (force processing), file does not exist, or is *outdated* (any of target's source files are newer than the target file). Please feel free to use as it as a reference. Thanks!

PR #1906, opened 2024-02-16
Cached recipes

### LINKED ISSUE #867: files as dependency
Hi,
I'm glad to discover `just` since `make` is painful to write. However, I still like some dependencies to be actual files. The recipe should run if file exists or **has been updated**. "File exists" is easy to find a workaround and there are some workarounds for "if file changed" cases such as [`watchexec`](https://github.com/mattgreen/watchexec) or tricks mentioned in #424. But I was still wondering if it's possible to use a file as dependency so that dependees run if file has changed or file is missing?

Just a hypothetical example:
```
process-data: input.csv
    #!/usr/bin/env Rscript
    library(tidyverse)
    input <- read_csv("input.csv")
    #some analysis
    write_csv("output.csv")
```

Is it possible that `just process-data` does not run if `input.csv` has not changed?

nmay231:
Closes #867, #1861, cc #1685

Putting this out there to get some preliminary feedback, ask questions, and take advantage of self-inflicted peer pressure :grin:

### What's in this PR?

Adds a `[cached]` recipe attribute that marks that recipe as cacheable. A cacheable recipe only reruns when the body of the recipe changes **after variables and other `{{interpolations}}` are evaluated**. [See examples](https://github.com/casey/just/issues/867#issuecomment-1936921368) in #867.

### TODOs

- [x] Set up basic recipe caching.
- [ ] Finish work on how dependencies work with cached recipes.
- [ ] Add `--no-cache` to force run cached recipes.
- [ ] Add `--delete-cache` to delete the cache file for this project.
- [ ] Update README.
- [ ] Future PRs (I may or may not get to these):
  - [ ] Prevent double work in the evaluator when checking if the recipe needs to be run or not.
  - [ ] Expand `sha256_file()` and `blake3_file()` to take directories and/or globs, or add a more explicit `cache_files()`/`hash_files()` with this functionality.
  - [ ] Add `cache_env_vars()` to allow caching by a glob of different environment variables, e.g. `CARGO_.*`.
  - [ ] Add `cache_ignore()` to ignore certain content while caching that will always/often change such as `uuid()` and `just_pid()`.
  - [ ] Allow arguments like `[cached("expression")]` that 'stratify' a recipe so you can cache a recipe by independent targets.

### Decisions, decisions... Edge cases and questions

- [ ] Besides backtick expressions and calls to `uuid()` and `just_pid()`, are there other impure interpolations that you can think of off the top of your head? I've already scanned the README, but wanted to double check.
- [ ] Cached recipes with dependencies do not have perfectly obvious behaviors. Here is how I am going to go with it (I'm looking for feedback on these):
  - [ ] The *prior* dependencies of cached recipes must also be cached, otherwise, it's ambiguous if you would run a dependency if the main recipe matched the last cached run.
  - [ ] Subsequent dependencies (`&& after`) do not need to be cached and do not run if the recipe matches the last run. I don't think they need to run because they are typically clean-up jobs and are only needed if the recipe actually ran. If the user wants them to run every time, they should set up a separate recipe that depends on the main recipe and subsequent in that order.
  - [ ] An unchanged recipe with a changed dependency will re-run both.
  - [ ] A changed recipe with an unchanged dependency only re-run the former and not the dependency.
- [ ] Did I miss any stupidly obvious issue that needs addressing? I am writing this at 3 in the morning, so I wouldn't be surprised.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2995,12 +2995,13 @@ recipes also rerun if one of its prior (normal) dependencies has to rerun. A
 cache folder called `.justcache/` will be added to the directory containing the
 main `justfile`, and should NOT be added to version control.
 
-Cached recipes are currently an unstable feature (requires `--unstable` or
+Cached recipes are currently an unstable feature (Run with `--unstable` or
 similar) that only work with shebang or script recipes for now. Additionally,
 all prior dependencies must be cached recipes, but that is not required of
 subsequent deps.
 
 ```just
+set unstable
 variable := "My value could change by running `just variable=changed my-recipe ...`"
 
 [cached, script]
@@ -3032,7 +3033,7 @@ Until directories or glob patterns are supported as inputs to `sha256_file()`
 and `blake3_file()`, here is a useful pattern to run a cached recipe if any files in a
 directory change (or you can just pipe the output of find to sha256sum):
 
-```just
+```sh
 @# {{sha256(`find $DIR -type f -exec sha256sum "{}" ";"`)}}
 ```
 
--- a/README.md
+++ b/README.md
@@ -2087,6 +2087,7 @@ change their behavior.
 | `[private]`<sup>1.10.0</sup> | alias, recipe | Make recipe, alias, or variable private. See [Private Recipes](#private-recipes). |
 | `[script]`<sup>1.33.0</sup> | recipe | Execute recipe as script. See [script recipes](#script-recipes) for more details. |
 | `[script(COMMAND)]`<sup>1.32.0</sup> | recipe | Execute recipe as a script interpreted by `COMMAND`. See [script recipes](#script-recipes) for more details. |
+| `[cached]`<sup>TODO</sup> | recipe | Cache a recipe by the contents of its body. See [cached recipes](#cached-recipes) for more details. |
 | `[unix]`<sup>1.8.0</sup> | recipe | Enable recipe on Unixes. (Includes MacOS). |
 | `[windows]`<sup>1.8.0</sup> | recipe | Enable recipe on Windows. |
 | `[working-directory(PATH)]`<sup>1.38.0</sup> | recipe | Set recipe working directory. `PATH` may be relative or absolute. If relative, it is interpreted relative to the default working directory. |
@@ -2985,6 +2986,55 @@ The interpreter path `/bin/sh` will be translated to a Windows-style path using
 If the interpreter path does not contain a `/` it will be executed without
 being translated. This is useful if `cygpath` is not available, or you wish to
 pass a Windows-style path to the interpreter.
+
+### Cached recipes
+
+Recipes can be marked with `[cached]` to avoid unnecessary reruns. Cached
+recipes only run if its contents, *after evaluation*, change in any way. Cached
+recipes also rerun if one of its prior (normal) dependencies has to rerun. A
+cache folder called `.justcache/` will be added to the directory containing the
+main `justfile`, and should NOT be added to version control.
+
+Cached recipes are currently an unstable feature (requires `--unstable` or
+similar) that only work with shebang or script recipes for now. Additionally,
+all prior dependencies must be cached recipes, but that is not required of
+subsequent deps.
+
+```just
+variable := "My value could change by running `just variable=changed my-recipe ...`"
+
+[cached, script]
+my-recipe ARG: dependency
+  @# This recipe will rerun if any of the following change: {{variable}} {{ARG}} {{`$HOME`}} {{sha256_file("./input")}}
+  @# Also reruns if dependency would rerun
+  echo Recipe body contents have changed!
+  echo $RANDOM  # Even though the output of $RANDOM changes, the command is the same as far as `just` is concerned
+
+[cached, script]
+dependency:
+  ...
+```
+
+In the example above, you can see the many ways the *evaluated* recipe body can
+change thereby allowing a rerun. This allows you to precisely fine-tune exactly
+when a recipe needs to run again. Of course, manually changing the recipe also
+causes it to run the next time. While not necessary, it can be convenient to put
+the "inputs" to the cached recipe in a quiet comment like shown: `@# ...`.
+
+The last line of the above example might also unearth an easy misconception; the
+line `echo $RANDOM` is static as far as `just` is concerned and therefore
+doesn't invalidate the cache even if the output is always different. Instead,
+include `{{env(ENV_VAR, default)}}` somewhere in your recipe. On the flip side,
+including content that always changes like `{{uuid()}}` means the cache will
+always be invalidated.
+
+Until directories or glob patterns are supported as inputs to `sha256_file()`
+and `blake3_file()`, here is a useful pattern to run a cached recipe if any files in a
+directory change (or you can just pipe the output of find to sha256sum):
+
+```just
+@# {{sha256(`find $DIR -type f -exec sha256sum "{}" ";"`)}}
+```
 
 ### Setting Variables in a Recipe
 
--- a/src/analyzer.rs
+++ b/src/analyzer.rs
@@ -155,7 +155,8 @@ impl<'run, 'src> Analyzer<'run, 'src> {
     for recipe in recipes.values() {
       if recipe.attributes.contains(AttributeDiscriminant::Script) {
         unstable_features.insert(UnstableFeature::ScriptAttribute);
-        break;
+      } else if recipe.attributes.contains(AttributeDiscriminant::Cached) {
+        unstable_features.insert(UnstableFeature::CachedRecipes);
       }
     }
 
--- a/src/attribute.rs
+++ b/src/attribute.rs
@@ -9,6 +9,7 @@ use super::*;
 #[strum_discriminants(derive(EnumString, Ord, PartialOrd))]
 #[strum_discriminants(strum(serialize_all = "kebab-case"))]
 pub(crate) enum Attribute<'src> {
+  Cached,
   Confirm(Option<StringLiteral<'src>>),
   Doc(Option<StringLiteral<'src>>),
   ExitMessage,
@@ -33,7 +34,8 @@ impl AttributeDiscriminant {
     match self {
       Self::Confirm | Self::Doc => 0..=1,
       Self::Group | Self::Extension | Self::WorkingDirectory => 1..=1,
-      Self::ExitMessage
+      Self::Cached
+      | Self::ExitMessage
       | Self::Linux
       | Self::Macos
       | Self::NoCd
@@ -78,6 +80,7 @@ impl<'src> Attribute<'src> {
     }
 
     Ok(match discriminant {
+      AttributeDiscriminant::Cached => Self::Cached,
       AttributeDiscriminant::Confirm => Self::Confirm(arguments.into_iter().next()),
       AttributeDiscriminant::Doc => Self::Doc(arguments.into_iter().next()),
       AttributeDiscriminant::ExitMessage => Self::ExitMessage,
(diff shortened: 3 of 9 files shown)


casey:
Nice! Some feedback:

- This initial PR should be as simple as possible, so it's easy to review and land, and additional features can be added in follow-up PRs. So I would leave `--no-cache`, `--delete-cache`, and customizing the cache filename for later.

- What is the `Error::InvalidCachedRecipe` error for?

- Can you describe the structure of the cache file JSON?

src/cache.rs
```diff
@@ -0,0 +1,25 @@
+use std::collections::HashMap;
+use std::path::PathBuf;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) struct JustfileCache {
+  /// Only serialized for user debugging
+  pub(crate) working_directory: PathBuf,
```
tgross35: This should probably be the file path, since you can have multiple justfiles in the same directory

src/cache.rs
```diff
@@ -0,0 +1,25 @@
...
+  pub(crate) recipe_caches: HashMap<String, RecipeCache>,
```
tgross35: Maybe just `recipes` since the struct is only caches?

tgross35:
This looks awesome!

I still think it could be better to have an explicit "rerun recipe if the value of this thing changes" function so the user can determine what is important, rather than having `just` need to determine whether a function is impure. This would also mean that you could have backticked expressions that don't invalidate the cache.

Don't think this would be too bad if the indicator function also returns the value, so you could do:

```just
config:
    # There is probably a snappier name than rerun_if_changed
    echo "running for configuration {{ rerun_if_changed(env("PROJECT_CONFIG")) }}"
    # ...
```

casey:
> I still think it could be better to have an explicit "rerun recipe if the value of this thing changes" function so the user can determine what is important, rather than having just need to determine whether a function is impure. This would also mean that you could have backticked expressions that don't invalidate the cache.

This is definitely an interesting idea, but I'd like to explore how things look without explicit annotations. If it works well without having to annotate anything, that would be ideal.

src/cache.rs
nmay231: Yes, but the same justfile could be used in multiple directories :)

What I probably need to do is use the justfile path and working directory to be part of the hash that changes the cache filename.

nmay231:
@casey 

> What is the `Error::InvalidCachedRecipe` error for?

Having a call to `uuid()`, and the like in a cached recipe means the recipe will always change and never be cached. So they should be disallowed completely until `cache_ignore()` is implemented.

I also disallowed backtick interpolation because currently the command would be executed twice (one of the backlog problems). Even when it is evaluated once, it won't be run in the same execution order as `run_linewise` which only runs interpolations line by line. If the user determines their backtick expression is pure (like a version check `rustc --version`) then they can put the expression outside the recipe body. If it is impure (relies on previous state) then they can wrap it in `cache_ignore()` to only run during the main execution.

> Can you describe the structure of the cache file JSON?

The structure itself is `JustfileCache`. As mentioned in [my response](https://github.com/casey/just/pull/1906#discussion_r1493049390) to @tgross35 above, I probably need to change things a bit.

After this comment, I'll make it so there is a unique cache file per justfile path and working directory combo. The filename on my system is `~/.cache/justfiles/{project_name}-{&path_hash[..16]}.json`. The `working_directory` and (soon to be added) `justfile_path` will only be there for user debuggability.

> I would leave [...] customizing the cache filename for later.

I'm actually using the filename customization to not pollute the contributor's workspace during tests, so I'll leave it for now.

casey:
> Having a call to `uuid()`, and the like in a cached recipe means the recipe will always change and never be cached. So they should be disallowed completely until `cache_ignore()` is implemented.

I think it's fine to just not detect this and the recipe doesn't get cached.

> I also disallowed backtick interpolation because currently the command would be executed twice (one of the backlog problems). Even when it is evaluated once, it won't be run in the same execution order as `run_linewise` which only runs interpolations line by line. If the user determines their backtick expression is pure (like a version check `rustc --version`) then they can put the expression outside the recipe body. If it is impure (relies on previous state) then they can wrap it in `cache_ignore()` to only run during the main execution.

Let's punt on linewise recipes and only allow `[cache]` on shebang recipes for now. We can allow backticks in that case and figure out linewise recipes later. Shebang recipes will be easiest to cache, since we just evaluate the text once and can make a decision to execute or skip based on the evaluated text. I don't want to worry about a `cache_ignore` function for the time being.

> The structure itself is `JustfileCache`. As mentioned in my response to @tgross35 above, I probably need to change things a bit.

I think we should aim for an initial very simple version which has shortcomings and then fix those one by one.

For now, what do you think about just having a single `.justcache` directory, which has files named after the sha hash of the evaluated recipe body, which contain the evaluated text of the recipe body?

My thinking is that:

- It's useful to be able to see what's cached with `ls .justcache` and seeing evaluated recipe bodies with `cat .justcache/HASH`.
- What exactly should be included in the cache key is tricky. There is the evaluated recipe body, the working directory, and justfile settings. Anything else?

We can initially land this feature behind `set unstable`, and then we can think about what needs to be in the cache keys one at a time. So for example, we can go, "Okay, the working directory needs to be in the cache key, let's do a PR for that.", and make incremental improvements to the feature, until it's ready to be made stable.

nmay231:
> I think it's fine to just not detect this and the recipe doesn't get cached.

Alrighty. I'm just trying to prevent open issues like "`[cached]` doesn't work". If you prefer a simpler implementation, I'll do that.

> For now, what do you think about just having a single `.justcache` directory, which has files named after the sha hash of the evaluated recipe body, which contain the evaluated text of the recipe body?

I am very opposed to this. While it sounds good in principle to be able to read the previous cached recipes, most of the time the hash is gonna change a lot because the recipe contains `sha256_file(...)` while the rest of the recipe body remains the same. I don't want to have to delete the cache all the time.

That said, I would not mind if the exact format changed somewhat. For example, I wouldn't mind having the cache file/directory in the working directory instead of the user's general cache directory.

> What exactly should be included in the cache key is tricky. There is the evaluated recipe body, the working directory, and justfile settings. Anything else?

The recipe name and/or justfile path should also be included in the cache input. If we do put the cache file/directory in the working directory, that would not need to be part of the input. I can't think of anything else.

> We can initially land this feature behind set unstable

:+1:

amarao:
Don't forget to put version check into cache. Eventually it will be version 2 of the cache, and someone will run older just with newer cache.

Expected behavior: either invalidate cache, or refuse to run.
Bad behavior: ignore version 2 in the cache file and try to re-use it.

The main issue is that it's impossible to add behavior to older versions (they are old by definition), so  safeguard should be from from the very beginning.

nmay231:
@casey Two quick things:

1. Do we really need to constrain this to only shebang recipes to prevent double evaluation if this is unstable only? I feel it's better to allow shebang and linewise recipes since it requires `--unstable` anyways. Then users can keep the benifits of linewise recipes (early exit on failure, linewise command printing) while trying out caching. Of course, I'll work on preventing double evaluation in a separate PR.
2. Is the current cache format a show-stopper for you, or is the json file okay? I already expressed why I didn't like it in the previous comment. I did simplify it to a file in a local `.justcache/` directory, where there's a json file for every justfile * working_dir combo.

(Still need to handle recipe dependencies and add docs, but this should be done other than that)

nmay231:
@amarao I decided to not include an explicit version in the cache format to simplify things and instead let the schema be the version. Obviously, we still handle outdated/invalid schemas by invalidating the cache.

Besides, running `just` with a cache file that is the wrong version is much rarer than you might think since `.justcache/` *should* be local to each machine and not included in git history.

nmay231:
I still have the [two previous questions](https://github.com/casey/just/pull/1906#issuecomment-1990643471) that need to be answered, but it's ready besides that. Whenever you have time for this full-time unpaid job, of course :)

One of those questions is about double evaluation, but that feels much easier to handle in the context of how `cache_ignore()` will be implemented (since that will basically need to be evaluated once during cache checks and once again during runtime). I will probably implement that and future features as a stacked diff (multiple PRs).

The only other thing to consider (that I know about) is that this might heavily conflict with #1562, however, that seems to be paused at the moment due to the author's lack of time.

casey:
> 1. Do we really need to constrain this to only shebang recipes to prevent double evaluation if this is unstable only? I feel it's better to allow shebang and linewise recipes since it requires `--unstable` anyways. Then users can keep the benifits of linewise recipes (early exit on failure, linewise command printing) while trying out caching. Of course, I'll work on preventing double evaluation in a separate PR.

I think I'd rather constrain it to shebang recipes at the moment. Linewise recipes seem much more complicated. One issue is that linewise recipes are evaluated line-by-line, so an error in an early line will prevent a later line from running. If a later line relies on an earlier line, for example because it creates a file that is read in an expression in a later line, then evaluating it ahead of time would produce an error. We could, in the case of cached recipes, evaluate linewise recipe lines ahead of time, and only re-run the recipe if no lines have changed, which is probably what we want to do, but would be a larger change, and I'd rather get this PR in sooner rather than later.

> 2. Is the current cache format a show-stopper for you, or is the json file okay? I already expressed why I didn't like it in the previous comment. I did simplify it to a file in a local `.justcache/` directory, where there's a json file for every justfile * working_dir combo.

Ultimately, I think I'd like to use [redb](https://github.com/cberner/redb/) an embedded pure-rust key/value store, for the cache. I'm happy to do that myself though, so whatever format is easiest for you for now and easiest to review is fine. Since `redb` databases block on concurrent access, this would also prevent errors where two instances of `just` try to modify the cache at the same time, and potentially also allow for concurrent access, if something like #1562 is ever merged.

> One of those questions is about double evaluation, but that feels much easier to handle in the context of how cache_ignore() will be implemented (since that will basically need to be evaluated once during cache checks and once again during runtime). I will probably implement that and future features as a stacked diff (multiple PRs).

I'm not sure I understand this point. Can't recipes be evaluated only once, before running, and the evaluated recipe body used both for cache checks and execution?

casey:
There are some conflicts, but they shouldn't be too bad to resolve. Let me know when this is ready for review!

casey:
Converted to a draft just so I can keep track of what's ready to go. Feel free to undraft!

nmay231:
Sounds good! I recently changed jobs, but I'll try to get back to this this week.

> One issue is that linewise recipes are evaluated line-by-line, so an error in an early line will prevent a later line from running. If a later line relies on an earlier line, for example because it creates a file that is read in an expression in a later line, then evaluating it ahead of time would produce an error.

True, but that's only if the user is using backtick expressions to mutate things "during compile time" which *should be* discouraged. This is one reason I wanted to [disallow backtick expressions initially](https://github.com/casey/just/pull/1906#issuecomment-1949462884) until `cache_ignore()` was fully implemented. (To prevent confusion, I mean ```expr {{`with`}} backticks``` since regular backticks ```expr `with` backticks``` are passed to the shell directly and safely run line-by-line.)

In any case, I agree that cached line-by-line recipes should have the same semantics as shebang recipes (all expressions evaluated at start before passed to shell). Ideally, cached line-by-line recipes would run *exactly the same way* as cached shebang recipes, but you can still reap the benefits of changing line-echos with `@`, spreading commands over multiple lines with `\`, etc. In other words, line-by-line recipes would be the odd one out.

> > One of those questions is about double evaluation, but that feels much easier to handle in the context of how cache_ignore() will be implemented (since that will basically need to be evaluated once during cache checks and once again during runtime). I will probably implement that and future features as a stacked diff (multiple PRs).
> 
> I'm not sure I understand this point. Can't recipes be evaluated only once, before running, and the evaluated recipe body used both for cache checks and execution?

I initially thought I would do two evaluation runs to implement `cache_ignore()`, but it would be much simpler to produce two strings, one for hashing and comparing to the cache, and one to pass to the shell to run.

EDIT: So much for that deadline, haha. I think I'll have time to look at this again soon.

nmay231:
Due to lack of time and lack of personal interest/usecases, I'm going to close this PR. I hope someone else finds the initial work I did useful. It would help to look at the [original version](https://github.com/nmay231/just/compare/testing...nmay231:just:old.cached-recipes?expand=1) that worked more or less. The code that I force pushed just now was when I started to update to the latest changes, but stopped.

casey:
Thanks for following up! Sad to see this closed, but open source is volunteer work, so life takes precedence!

snoweuph:
Out of curiosity, whats currently missing from this PR if somebody would want to pick it up? :eyes:

PR #1908, opened 2024-02-17
Adding make-style multivalue assignments

### LINKED ISSUE #1889: [Feature request] Array variables
In Make, one can do something like `val := ent1 ent2 ent3 ...`, which Make will interpret as an array. Can we have something similar to it that'll also be compatible with variadic recipes and functions?

Bonus point for "post setting",  eg
```make
var := initial
# yada yada
var += additions
```

gyreas:
This PR adds support for one of the iconic features of Makefiles: multivalued assignments, where a single variable holds a list of values separated by whitespace.
The trick is to make `key := value` a special case of `key := value1 value2 ...`, which is the easiest way out. 

This PR hopefully enables new use cases for Just.

EDIT: 
Broken tests are now passing. 
A follow-up PR will enable Just functions to operate on multivalued variables and add more Just functions that work on multivalued variables. Another one will make variables modifiable with `+=` syntax (or something similar).

Closes #1889


casey:
Thanks for the PR! I might be missing something, but I don't think that this buys us very much. It's essentially a shortcut for creating space-separated variables. As is, I don't think it's very useful, and would much prefer using #11 to do that.

gyreas:
I guess I misunderstood some things. I can't find the branch you referenced in #11, tho. Also, I'll shift my focus to porting some convenience functions from make, then. Furthermore, I'll probably look into #11 later on.

casey:
The main issue is that if we add:

```
foo := bar baz
```

But all it's really doing is:

```
foo := bar + " " + baz
```

Then it's not very useful.

My point about formatting strings is that if the user wants a space-separated list, they could do:

```
foo := f'{bar} {baz}'
```

Which seems nicer than extending the grammar.

I think that when people want lists of things, the thing that makes them useful is the ability to access elements of the list, modify the list, and perform other operations on the list:

```
foo := ["a", "b"]
bar := foo[0]
baz := foo[1]
all := join(foo, " ")
```

The issue with that, however, is that in order to figure out if `foo[0]` is allowed, `just` would need to know whether `foo` is a string or a list of strings. This would require a more advanced type system than just currently has. Currently, just has a static type system, in the sense that it knows which variables exist, but all variables have type string, so this would require extending just to allow multiple types of variables, and to allow both strings and lists of string. This is desirable, but it's a much bigger change.

#11 is an issue for allowing interpolation inside of strings, so you could do something like `f'{foo} and {bar}'`, so you could interpolate inside of strings instead of having to do `foo + " and " + bar`, which is longer and more annoying.

gyreas:
I understand. I can't tackle it right now, however; instead, I'll work on adding more convenience functions from make. Are cool with that?

casey:
Of course! If there's something you want to add, open an issue first to discuss the feature and design, so we can make sure it's something that'll get merged.

PR #1910, opened 2024-02-19
Add -E option (alias for --dotenv-path)

amarao:
I found that I use few env files to manage the same recipes between few environments. I do this very often and it's like 'ansible inventory' to me.

Typing `--dotenv-filename` each time is really tiresome. Can we use `-E` as shorthand for it?


casey:
Thanks for the PR! I definitely agree that we should have a flag for this. But I'm not sure if it should be `--dotenv-path` or `--dotenv-filename` `--dotenv-path` lets you load an environment file anywhere on disk, whereas `--dotenv-filename` sets a different filename which just searches for in the usual places. Which do you think is more useful?

amarao:
Oh. Wow. This is, actually, very unexpected to me. I thought that `--dotenv-path` is changing path, and `--dotenv-filename` allows to specify the path to the file, including absolute path.

... I checked, it works as expected. `just --dotenv-filename=/path/to/file` uses /path/to/file correctly. (I was afraid of having something non-working here).

In my use-case, we use files (e.g. `--dotenv-filename`). Maybe some people use directories too, and it worth adding short option for it too.

Actually, I already use it for a week with hand-build patched version with `-E` and it really raises ergonomics.

(I'll look on failed linter now).

amarao:
Oh, I found the autocomplete issue. It's autogenerated. Fixed.

casey:
> Oh. Wow. This is, actually, very unexpected to me. I thought that `--dotenv-path` is changing path, and `--dotenv-filename` allows to specify the path to the file, including absolute path.
> 
> ... I checked, it works as expected. `just --dotenv-filename=/path/to/file` uses /path/to/file correctly. (I was afraid of having something non-working here).

This works, but for a reason you might not expect.

`--dotenv-path` specifies a single path to search for a dotenv file, which may be relative or absolute. If `just` doesn't find that path, it throws an error.

`--dotenv-filename` specifies a path to search for a dotenv file, which, again, maybe relative or absolute. While searching for a `.env` file, just will go through the ancestors of the current working directory (`.`, `../`, `../../`, etc), and join the path with those directories, and see if there's a env file there.

Now, the reason that `--dotenv-filename` worked with `--dotenv-filename /path/to/file` is because when a path is joined with an absolute path, the result is the absolute path. `../` joined with `/path/to/file` is just `/path/to/file`.

The place where you would see a difference is when using a relative path with both, and where the env file is in a parent directory. Here's a somewhat contrived example:

```
$ tree
test
├─ dir
│  └─ justfile
└─ foo
$ cat foo
FOO=bar
$ cat dir/justfile
default:
  echo $FOO
$ cd dir
$ just --dotenv-filename foo
echo $FOO
bar
$ just --dotenv-path foo
error: Failed to load environment file: No such file or directory (os error 2)
```

In this case, `just --dotenv-filename foo` works, because just first looks in `./foo`, then `../foo`, and finds the environment file.

`just --dotenv-path foo` does not work, because just first looks in `./foo`, does not find the file, and then gives up.

If you're using `--dotenv-filename` with an absolute path, `--dotenv-path` will work too.

One important difference is that `--dotenv-path` will complain if the file doesn't exist, whereas `--dotenv-filename` won't. I suspect that the behavior of `--dotenv-path` is probably the one that should be the default, since if a user gives `--dotenv-filename` with a relative path which doesn't exist, then just won't complain if it doesn't find a env file, which might be confusing.

amarao:
Yes, I've noticed the odd behavior of --dotenv-filename when file is missing. May be, it worth  an additional option, like `--dotenv-file`, which will accept only the normal path (e.g. local to the current workdir or absolute), and which WILL complain if this file is missing? In this case '-E' may go into that option.

I also wonder, if it worth an environment variable.

casey:
> Yes, I've noticed the odd behavior of --dotenv-filename when file is missing. May be, it worth an additional option, like `--dotenv-file`, which will accept only the normal path (e.g. local to the current workdir or absolute), and which WILL complain if this file is missing?

I might be misunderstanding, but isn't this what `--dotenv-path` does?

amarao:
F....

Okay, I got it. It's misunderstanding from my side. I thought --dotenv-path is the path to the directory where to search for .env file.

Here the help:

`--dotenv-path <DOTENV-PATH>` Load environment file at `<DOTENV-PATH>` instead of searching for one

The problem is with 'at path'. This sentence (as far as I got it) means 'load file' (located) at DOTENV-PATH.

I propose to move '-E' to `--dotenv-path`, and update help for it to be:

`--dotenv-path <DOTENV-PATH>` Load `<DOTENV-PATH>` as environment file  instead of searching for one.

Is it okay?

And sorry for misunderstanding and commotion.

amarao:
I've updated my branch, but github is showing old content. Really, really odd. The actual diff looks like this: https://github.com/amarao/just/pull/2/files

amarao:
@casey, can you see it? I've updated to be alias for `--dotenv-path`

amarao:
Sorry for nagging, @casey, can you look at it? I use forked/selfcompiled version with '-E' option for convenience, but I very much to prefer to stay with upstream.

casey:
LGTM!

PR #1911, opened 2024-02-19
Enable command color for shebang recipes

### LINKED ISSUE #1903: Use command-color for shebang recipes
The recent `command-color` flag works nicely for regular just recipes, but it's not used for shebang recipes (e.g. those which start with `#!`).

So if I do `just --command-color purple foo`, it will echo recipe lines in purple if `foo` is a regular recipe, but will still use white if `foo` is a shebang recipe.

Unless anyone thinks this is the desired behavior, I can probably work on a fix for this.

avi-cenna:
Small MR which resolves #1903, Adding `command-color` for shebang recipes. Now if your shebang recipe is prefixed with `@` and you specify a `command-color`, it will use that when printing out the shebang recipe.


casey:
LGTM, thank you for the PR!

casey:
I simplified it a little bit by just always using `config.command_color`, since when it's `None`, it's equivalent to no color.

PR #1912, opened 2024-02-19
install.sh: fix shellcheck and convert to sh

tgross35:
This fixes all errors from https://www.shellcheck.net/. Also change the shebang to POSIX sh, which helps improve compatibility (e.g. this can now run in Alpine contianers).

The only bash-specific thing was `pipefail` which is easy to ignore. Most of these other fixes are just [SC2086](https://www.shellcheck.net/wiki/SC2086), and remove the unused `git` variable.

Tested on an alpine docker image.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,11 +36,16 @@ jobs:
         git diff --no-ext-diff --quiet --exit-code
         ./tests/completions/just.bash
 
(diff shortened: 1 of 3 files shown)


casey (approved):
LGTM!

casey:
I also added a CI lint step to run shellcheck on install.sh.

PR #1913, opened 2024-02-20
Add a fallback to wget if curl is not available

tgross35:
This allows using the install script on Alpine containers and other busybox-based systems with no additional dependencies.

The script attempts to use `--https-only --secure-protocol=TLSv1_3 --no-verbose` if available, otherwise falls back to a portable wget call.

Additionally, error on invalid arguments (currently syntax errors silently continue).

This is based on top of https://github.com/casey/just/pull/1912, which should be merged first. The main change for this PR is in the second commit.

Tested on ubuntu (wget is GNU and sh is Dash, no pipefail) and alpine (wget and sh are both alpine, yes pipefail)


www/install.sh
```diff
@@ -51,6 +51,27 @@ need() {
...
 
+download() {
+  url="$1"
+  output="$2"
+
+  # use curl if available, wget otherwise
+  if command -v curl > /dev/null; then
+    curl --proto =https --tlsv1.2 -sSfL "$url" "-o$output"
+  else
+    # Some wget implementations support these flags, but busybox
```
casey: How about trying `wget` with all options and then falling back to using none of the options if that fails? We might try the download twice, if there's some other error, like the network not being up, but that might be better than checking each flag individually?

casey (changes requested):
Seems reasonable to me! There are conflicts because along with your shellcheck PR, I also added a check on CI which runs shellcheck, but they should to be easy to merge.

www/install.sh
tgross35: That sounds reasonable, I updated to do this and tested with both ubuntu and alpine. I had to suppress stderr from the first run so it doesn't print the help message on Alpine, but that should be fine.

tgross35: Actually... maybe we don't want to just try twice? With the change, if HTTPS fails or TLS is outdated then it will just rerun without the flags :)

I don't think the flags actually add much since the URL is https, your call for what to do

casey:
Is it just old versions of wget which don't support those flags? I think perhaps we should always use those flags, since I'm a little bit wary of not using them just because they aren't supported.

casey (approved):
I removed the fallback to insecure wget. This fallback would have triggered if there was a cert error, so a MITM could potentially have intentionally downgraded the connection to insecure wget.

If it's necessary to support boxes that don't have a secure curl or wget, we should add a `--insecure` flag download with forcing TLS.

PR #1915, opened 2024-02-20
docs: add example of using GNU parallel to run tasks in parallel

amarao:
I found amazing synergy between GNU Parallel and just (basically, use parallel in `--shebang` mode), adding example into examples/.


casey:
This is great! I think we should actually add this to the readme, since it seems super useful.

A few things:

- There's a `num_cpus()` function, which can be used to set the number of jobs appropriately.
- I think it's a good idea to use long flags instead of short flags, so `--jobs` instead of `-j`.
- With `--ungroup`, `parallel` prints job output right away, instead of waiting for the job to finish, which lets you see when jobs start up.
- `/usr/bin/env` will run `parellel` regardless of where it is.
- I think we should call the recipe `parallel`, since `churn` is a bit uninformative.

With these changes, the recipe looks like this:

```just
parallel:
  #!/usr/bin/env parallel --shebang --ungroup --jobs {{ num_cpus() }}
  echo task 1 start; sleep 1; echo task 1 done
  echo task 2 start; sleep 2; echo task 2 done
  echo task 3 start; sleep 3; echo task 3 done
  echo task 4 start; sleep 4; echo task 4 done
```

amarao:
I've updated the example.

amarao:
Oh, and I found how to force env and parallel to work together in shebang mode.

Thanks for feedback.

casey:
I moved this into the readme, since I think it'll be more discoverable there. Thanks for the PR!

PR #1916, opened 2024-02-20
Update depedencies, remove atty and general cleanup

rosymati:
This pr is fairly simple. 
- It updates all the dependencies to their latest version, purposely omitting clap since that needs a proper migrations. 
- It  makes the manifest much cleaner by making all the dependencies versions consistent with each other whereas before some of them only specified the major version. 
- The atty crate is also removed since it is no longer maintained and has security vulnerabilities, instead it replaces it with the std implementation.
- Finally it also makes the release profile more optimized and smaller by reducing the codegen units and stripping

I plan to make more prs to modernize the codebase and replace other dependencies. This project is old enough (relatively speaking) that the rust ecosystem has grown a lot since. This pr is a good starting point to ensure a clean and updated base to work on

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,9 +22,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.11"
+version = "0.6.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e2e1ebcb11de5c03c67de28a7df593d32191b44939c482e97702baaaa6ab6a5"
+checksum = "96b09b5178381e0874812a9b157f7fe84982617e48f71f4e3235482775e5b540"
 dependencies = [
  "anstyle",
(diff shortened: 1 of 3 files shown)


casey:
The release optimizations and getting rid of atty are both great. I'm not sure about the changes to the `Cargo.toml` file though. My thought is that when packaging `just`, sometimes packagers have to package each dependency along with `just`, and requiring the latest version of each dependency may create more work for them, because they potentially have to package new versions of each dependency, when in fact, the older, previously packaged dependency version would have done. I know that FreeBSD and Debian both do this.

How about putting everything but the changes to `Cargo.toml` in a separate PR which I can merge now?

@yurivict (packager of `just` for FreeBSD) do you have any thoughts about this?

casey:
I merged just the atty removal in #2066. Thank you for the PR!

PR #1917, opened 2024-02-20
Replace libc with rustix

rosymati:
This pr replaces the libc crate with the rustix crate. [Rustix](https://github.com/bytecodealliance/rustix) provides safe binding to posix/linux apis. On linux it uses direct syscall and falls back to libc on other systems.
Switching to rustix allows to remove unsafe usage and potentially unlock safer abstractions in the future.
The change is completely transparent and doesn't add or remove any dependency since most crates used by just already used rustix internally

Depends on #1916


tests/interrupts.rs
```diff
@@ -3,11 +3,7 @@ use {
   std::time::{Duration, Instant},
 };
 
-fn kill(process_id: u32) {
-  unsafe {
-    libc::kill(process_id as i32, libc::SIGINT);
-  }
-}
+use rustix::process::{kill_process, Pid, Signal};
```
casey, line 6: These use statements can be combined.

tests/interrupts.rs
```diff
@@ -25,7 +21,8 @@ fn interrupt_test(arguments: &[&str], justfile: &str) {
 
   while start.elapsed() < Duration::from_millis(500) {}
 
-  kill(child.id());
+  // FIXME: the old libc implementation ignored errors, this does the same but it's probably not the best idea
+  let _ = kill_process(Pid::from_child(&child), Signal::Int);
```
casey, line 25: Try unwrapping the error. As long as the tests pass, it's probably fine.

casey (changes requested):
Looks mostly good to me! See comments.

casey:
I think I'm inclined not to do this. `libc` is pretty widely used and tested, so absent a concrete issue that `rustix` would fix, I'm not sure it's worth it.

PR #1918, opened 2024-02-21
Reordered Packages table, and added openSUSE

sfalken:
I actually only meant to add openSUSE, but the table organization was bugging the heck out of me.

It's now sorted by Column A, then Column B alphabetically, which just makes more sense.


casey:
Seems reasonable to me!

Issue #1827, opened 2024-01-09
"error: Variable `project_name` has multiple definitions" when importing justfiles


sylvorg:
Hello!

I am getting the error in the title when trying to import a `justfile` with the same variable as in the parent file; is there any setting, perhaps similar to `allow-duplicate-recipes`, that works for variables as well, such as `allow-duplicate-variables`, maybe?

Thank you kindly for the help!

casey:
I'm not sure I understand, can you copy and paste your justfiles so I can take a look?

sylvorg:
Not at my computer right now, _but,_ it's something like this:

`p/a/justfile:`
```make
x := `basename {{justfile_directory()}}`
```
Where `x` results in `a`, and:

`p/b/justfile:`
```make
import "../a/justfile"
x := `basename {{justfile_directory()}}`
```
Where `x` is _supposed_ to result in `b`, but `just` errors out with what's in the title of the issue.

casey:
Gotcha. I think that `set allow-duplicate-variables` would be reasonable to add, and not too much work.

casey:
I marked this as good first issue, in case anyone wants to take a crack at it.

sylvorg:
Wish I could help, but I know nothing of rust, though I've been wanting to learn it for a while! 😅

bukowa:
Can this be closed as completed?

sylvorg:
I have not tested the changes yet, as I'm working on something else at the moment, unfortunately! 😅 We could possibly reopen it if necessary?

bukowa:
> I have not tested the changes yet, as I'm working on something else at the moment, unfortunately! 😅 We could possibly reopen it if necessary?

`set allow-duplicate-variables := true`

I am just using it ^^ I landed here trying to find out why it didn't work before

sylvorg:
Then we should be good to go, I think! My thanks to everyone involved here! 😸

teto:
is there a way to check if the variable is already set and dont override it instead ?

PR #1922, opened 2024-02-21
Implemented 'set allow-duplicate-variables'

Mijago:
I ran into the same issue as https://github.com/casey/just/issues/1827 and decided to make a quick draft for it.
It's my first rodeo with rust (and I am not a big fan so far, but it may grow onto me), so I appreciate feedback.

This issue allows us to overwrite assignments, even when they are imported from other justfiles. Previously it would have thrown an error. We now can use `set allow-duplicate-variables` to enable this behavior (in the same style as the existing `set allow-duplicate-recipes`).

# Example: Default behavior
```
a := "foo"
a := "bar" 
#  Example: Raises an exception
```

# Example: Using the flag to overwrite a variable
```
set allow-duplicate-variables

a := "foo"
a := "bar"
# a is now "bar"
```

# Example: Using the flag in an import
```
# imp.just
a := "foo"
```

```
# some.just
import "./imp.just"
set allow-duplicate-variables

a := "bar"
# a is now "bar"
```

## Feedback appreciated
I am not quite happy with the detour I had to take with `DepthAwareBinding`. I introduced it because the assignments need to be aware of their depth, otherwise imported modules would overwrite assignments that appear later in the justfile. As I did not use rust before, I do not know the best practices for this in this language and just wanted to get the draft done (for now). 

### Some other thing
I ran into `just` just (hehe) a few months ago - and fell in love. Thanks for this tool!


Mijago:
@casey Hi! I do not want to come off rude or impatient - but is there anything I should change in order to progress this PR further?

casey:
Sorry for letting this sit! This looks great. I made a couple changes:

- I removed `DepthAwareBinding` and just added a depth, which is set to 0 when not needed, to `Binding`.
- I removed the `modules_use_module_settings_variables` test. Since there is no code specific to different settings which implements this, I don't think a new test is needed for each setting.

PR #1924, opened 2024-02-23
chore(deps): update clap

poliorcetics:
Steps:

- [x] Move from clap 2 to 3
  - [x] Make it compile
  - [x] Make all tests pass 
- [x] Resolve deprecation warnings
  - [x]  Make it compile
  - [x] Make all tests pass
- [x] Move from clap 3 to 4
  - [x] Make it compile
  - [x] Make all tests pass

I know there is #1865 but I think it does it too quickly to make it possible. I'm following the changelog instructions from clap: https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrating, which makes it simpler


poliorcetics:
Note: current commit has a deprecation warning already, the fix is planned for the next commit where I'll resolve such warnings

casey:
Nice, this is a great approach, because it makes each commit easy to review independently.

The completion scripts are probably going to be a huge problem. The completion scripts are generated by clap, but then we patch them to add features. So if the completion scripts have changed a lot, those patchs might not apply cleaning, and would be hard to resolve.

If going from clap v3 to clap v4 winds up being hard, we can also merge just the move from v2 to v3.

poliorcetics:
> The completion scripts are probably going to be a huge problem. The completion scripts are generated by clap, but then we patch them to add features. So if the completion scripts have changed a lot, those patchs might not apply cleaning, and would be hard to resolve.

The issue is not the cleaning, I fixed that easily enough, I just can't get `clap_complete` to output to anything but `stdout`, which is obviously annoying for us. It happens with clap 3 too, so even that upgrade is blocked

casey:
Which function generates the completion script, and which version of clap_complete? Looking at the latest clap_complete, it looks like the generate function takes a writer to write to. [clap_complete::generator::generate](https://docs.rs/clap_complete/latest/clap_complete/generator/fn.generate.html).

poliorcetics:
Look at https://github.com/clap-rs/clap/issues/5372 for details, it's not related to just's usage of it but is a wider issue

tgross35:
To workaround that issue, could you generate the completion scripts in `build.rs`, patch them there, then `include_str!` them as needed?

poliorcetics:
Found the issue, I'll fix it and continue this pr

poliorcetics:
Okay I think I did everything I was supposed to for a clean clap upgrade (congrats to the clap changelog, it's very well made) so this is ready for review :D

casey:
Looks good to me! Thank you for the epic PR!

PR #1933, opened 2024-03-03
Implement loading justfile from stdin

drzymalanet:
I think it would be nice to have the possibility to pipe justfiles into just, like so:
```bash
cat input.txt | just --dump --justfile - > output.txt
```
It would work by detecting a special symbol "-" as the -f / --justfile argument.

Benefits:
 - Some code editors like helix require this in order to make auto formatting to work,
 - Many command line tools work like this and it seems it will not hurt to have,

The plan:
 1. Move `Search{}` functionality into `Source{}`
 2. Implement `Source::from_stdin()`
 3. Add "-" detection to the argument parser.
 
 Here is the first of the commits, more commits will be added while work progresses.

### DIFF
--- a/src/search.rs
+++ b/src/search.rs
@@ -6,7 +6,7 @@ const PROJECT_ROOT_CHILDREN: &[&str] = &[".bzr", ".git", ".hg", ".svn", "_darcs"
 
 #[derive(Debug)]
 pub(crate) struct Search {
-  pub(crate) justfile: PathBuf,
+  pub(crate) justfile: Option<PathBuf>,
   pub(crate) working_directory: PathBuf,
 }
 
@@ -51,15 +51,15 @@ impl Search {
         })
       }
       SearchConfig::GlobalJustfile => Ok(Self {
-        justfile: Self::global_justfile_paths()
+        justfile: Some(Self::global_justfile_paths()
           .iter()
           .find(|path| path.exists())
           .cloned()
-          .ok_or(SearchError::GlobalJustfileNotFound)?,
+          .ok_or(SearchError::GlobalJustfileNotFound)?),
         working_directory: Self::project_root(invocation_directory)?,
       }),
       SearchConfig::WithJustfile { justfile } => {
-        let justfile = Self::clean(invocation_directory, justfile);
+        let justfile = Some(Self::clean(invocation_directory, justfile));
         let working_directory = Self::working_directory_from_justfile(&justfile)?;
         Ok(Self {
           justfile,
@@ -70,7 +70,7 @@ impl Search {
         justfile,
         working_directory,
       } => Ok(Self {
-        justfile: Self::clean(invocation_directory, justfile),
+        justfile: Some(Self::clean(invocation_directory, justfile)),
         working_directory: Self::clean(invocation_directory, working_directory),
       }),
     }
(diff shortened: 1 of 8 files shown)


casey:
This seems reasonable, but this PR looks a bit too complicated and I think some of the changes are unnecessary.

I would create two new SearchConfig variants, `WithStdin`, and `WithStdinAndWorkingDirectory`, and handle those two cases separately in `Search::find`.

drzymalanet:
@casey I have added the variants, now we need to change `Search.justfile` type because it is a `PathBuf` which does not fit to the stdin stream. It seems to me that changing it to an `Rc<dyn std::io::Read>` or a custom `JustfileSource` would be some options. What do you think?

casey:
There are two options I can think of:

1. Add an additional field to `Search`, which is a string containing the loaded source
2. Make `Search::justfile` an enum, like `JustfileSource`

I would try out 1 first and see how it looks and see if it causes any problems, or requires a big refactor.

drzymalanet:
Sorry for closing and reopening. It was by mistake. 

@casey Please have a look at the new commit. I have added a `JustfileKind` enum to the `Search` struct. It does not feel right, but I have no idea how to make it better without a bigger refactor. Let me know what you think.

casey:
This looks like a good approach to me!

drzymalanet:
Hi @casey I have changed a little the implementation of this PR. Previous approach did not felt right for me. I think the new approach makes more sense. However, I stumbled upon a failing integration test: `working_directory::justfile_without_working_directory`. It seems it runs the command in a subshell, which makes it hard to find out what's the problem. Any advice on how to trace this?

Let me know what you think about the new changes.

Thanks!

src/search.rs
```diff
@@ -6,7 +6,7 @@ const PROJECT_ROOT_CHILDREN: &[&str] = &[".bzr", ".git", ".hg", ".svn", "_darcs"
 
 #[derive(Debug)]
 pub(crate) struct Search {
-  pub(crate) justfile: PathBuf,
+  pub(crate) justfile_path: Option<PathBuf>,
```
casey: I think this should just be `justfile`, since the type is already a path.

src/search.rs
```diff
@@ -198,18 +194,81 @@ impl Search {
...
-        .ok_or_else(|| SearchError::JustfileHadNoParent {
-          path: justfile.to_path_buf(),
-        })?
-        .to_owned(),
-    )
+  /// Returns the path to the directory where the jusfile is located.
+  /// Returns an empty path if the justfile path is relative and has only one component.
+  /// Returns an empty path if the justfile path is itself empty.
+  /// Returns the current working directory if the justfile is not a file on disk.
+  pub fn justfile_directory(&self) -> PathBuf {
```
casey: These functions are kind of footguns. If the justfile was read from stdin, then there is no path or directory to the justfile. I understand that these are kind of compatibility shims, but I think the code just has to be adapted to deal with the fact that the justfile path might be None.

casey (changes requested):
I can't really review this PR yet, because it doesn't actually implement reading justfiles from stdin, it's just kind of preliminaries, which, without seeing the implementation, I'm can't give good feedback on.

src/search.rs
drzymalanet: I agree. Fixed.

src/search.rs
drzymalanet: My reasoning is that you still need to know the working directory even if you load the justfile from stdin. On the other hand, many functions don't need to know both the working directory and the optional justfile path. That's what this function does. It abstracts away the justfile directory over the `Search` struct. I don't have strong opinion on this, but the abstraction seems reasonable. How do you see it?

casey: `working_directory` is fine, and should always be set. It's `justfile` and `justfile_directory` that I think are the issue, since there is no justfile, and no directory containing it.

casey:
Closing, since this seems to have stalled.

PR #1936, opened 2024-03-06
Correct Typo in README

Justintime50:
As I was reading through the README, I noticed there were two words repeated immediately after one another. This fixes the typo.


casey:
Nice, good catch!

Issue #1748, opened 2023-11-27
Allow loading multiple dotenv files


evandam:
Related: https://github.com/casey/just/issues/945#issuecomment-1828806192

It would be great to be able to source multiple dotenv files in a "cascading" manner, similar to https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use. This allows per-environment overrides, and a `.env` to provide defaults to fall back on (meaning `.env.test`, `.env.qa`, `.env.prod` don't need to repeat the same common env vars that can be defined once in `.env`).

I think there may be some related work that would need to be done to be able to interpolate filenames in `dotenv-files`, though.

```
ENV := "prod"
set dotenv-files := [".env.{{ENV}}", ".env"]
```

And call with `just ENV=prod` to load env vars from `.env.prod`, and then `.env` (assuming dotenv files don't overwrite existing environment variables).

As far as behavior goes, it probably makes sense to fail silently (or log it) if a file in the array isn't found since they can just be "candidates" for dotenv files to load. For example, `.env.local` may not always be present in an environment, but we'll try to load it if it does happen to exist.

azriel49:
it would be super useful to be able to override the default values ​​of the .env using a .env.local !!
Like the OP, i usually put my .env into the repository with default values, and override some confs with an environment specific .env.local

r3nor:
This would be great, many project use multiple dotenvs and I have to use bash scripts every time

sbeckeriv:
how about allowing more then on dotenv-filename config? 

`just --dotenv-filename=.env --dotenv-filename=.env.local command`

and 

```
      ENV := "prod"
      set dotenv-load
      set dotenv-filename := ".env"
      set dotenv-filename := ".env.local"
      set dotenv-filename := ".env.{{ENV}}"
      set dotenv-path := "path"
```

if they can be evaluated in order, last one wins. I was reading the code and I think updating the config to allow an vec of filenames and then modifying the current dotenv load process might be a smaller change. The code path would always be used making it less likely to have diverging behaviors. one could even skip supporting the command line part and make it a set config only. 

[edit]
there are a number of things that would need to be cleaned up, however, this is a diff of it working with arguments and settings. I would want to know if this was an acceptable way forward before spending more time on it. 

https://gist.github.com/sbeckeriv/664c3dd8a38c84c877a45f50d44a0f20 

While making this diff I learned that `[]` syntax is not supported currently. I believe there is other issues around supporting 'arrays' of values.

[...]

PR #1940, opened 2024-03-08
1748-support-multiple-dotenv-files

sbeckeriv:
Dearest Reviewer,

I use more then dotenv for different reasons. I saw the open issue #1748 and took a stab at it. This allow for more then one set value but only for the one type.

It was suggested in the issue to support an array syntax ["file", "file2"]. This is not supported today and I didnt look in to changing the parser to support that.

I did consider supporting a csv value like "file,file2" but I didnt move forward on that.

Some things to note:
It does expect the set lines to be parsed in order. I am unsure if that is true today. It could load files from different subpaths. This could be changed to only load from the path where the first file is found.

I do like that one file or 10 files they all use the same path.

I like using just. Thank you for your work.
Becker

N.B. The test functions::env_var_functions failed for me on master and my branch.


woutervh:
@casey What is blocking this review? It would be nice to support multiple dotenvs.

nk9:
The way this PR works is spelled out in #1748, namely:

`just --dotenv-filename=.env --dotenv-filename=.env.local command`

or

```
      ENV := "prod"
      set dotenv-load
      set dotenv-filename := ".env"
      set dotenv-filename := ".env.local"
      set dotenv-filename := ".env.{{ENV}}"
      set dotenv-path := "path"
```

I personally think this syntax is a bit confusing since it seems like you're replacing the filename, when in fact all the files will be loaded sequentailly. I think the array syntax described in the original issue would be much clearer.

sbeckeriv:
@nk9 thanks for the feedback. I will close this.

PR #1961, opened 2024-03-12
add --no-aliases flag to not show aliases when using --list

### LINKED ISSUE #1891: just --list  --exclude alias
I 'm defining lots of aliases for multiple commands

but this pollutes my default list-output,
filtering the output in windows looses the colored output

I would be nice to have an option to exclude aliases:
```
# list available recipes
default:
    @ just --list --justfile justfile --unsorted --no-alias
```

WJehee:
Hey there, this is my first contribution to this project, please let me know if any changes need to be made.

This PR aims to fix https://github.com/casey/just/issues/1891, since I came across the same problem.
I implemented the completions files, but I wasn't able to test most of them, I use zsh myself but the completions did not seem to work, although I could see them when passing `--completions=zsh`, perhaps you know what the problem is?


casey:
Looks good to me! This just needs a test.

WJehee:
Added a test, I'm not sure if the standard is to create a new file for this but that is what I did

casey:
Looks like there's an issue with formatting.

casey:
Merged!

PR #1966, opened 2024-03-15
I think accepting an all as param is misguiding.

aarmn:
Delete is taking param all, which is not used, and this may cause confusion as if "delete all" is name of the command. so, for clarification, I added a hyphen.


casey:
Good catch, thank you!

PR #1971, opened 2024-03-16
add homebrew to readme

imevro:
found formula in homebrew, thought it will be useful to be reflected in readme

 https://formulae.brew.sh/formula/just

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ most Windows users.)
     </tr>
     <tr>
       <td><a href=https://en.wikipedia.org/wiki/MacOS>macOS</a></td>
-      <td><a href=https://www.macports.org>Homebrew</a></td>
+      <td><a href=https://brew.sh>Homebrew</a></td>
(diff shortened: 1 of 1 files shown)


casey:
If I'm not mistaken, homebrew is already listed in the readme: https://github.com/casey/just/?tab=readme-ov-file#packages

imevro:
ah my bad

thanks!

cskeeters:
Funny, I just made the same change and almost submitted a pull request for it.  I wonder if the maintainers have considered having a separate table at the top of the packages section for cross-platform package managers?  Breaking it into two tables would help users conceptually understand that they need to check two spots.  As it is, I searched for my OS and then concluded it wasn't listed.  Apparently, I'm not the only one.

casey:
I wouldn't mind a PR which splits the table into two, with cross-OS options in the first table, and the OS-specific options in the second.

PR #1998, opened 2024-04-09
Update line continuations documentation

laniakea64:
Not sure when this changed, but this is valid now -
```
foo:
	echo {{ \
 "test" \
 }}
```

This is still invalid -
```
foo2:
	echo {{ \
"test" \
 }}
```
```
error: Expected backtick, identifier, '(', '/', or string, but found dedent
 ——▶ justfile:3:1
  │
3 │ "test" \
  │ ^
```


casey:
LGTM!

PR #2012, opened 2024-04-22
Port some functions from make

gyreas:
This PR adds some useful functions found in [make](https://www.gnu.org/software/make/).

Please read the corresponding commit's message for how each works
#### To be added
- [ ] addprefix() ([#8c9a965](https://github.com/casey/just/pull/2012/commits/8c9a965e7304d8a2310c069630fc85af24ca7060))
- [ ] addsuffix() ([#766c817](https://github.com/casey/just/pull/2012/commits/766c817a4069e844b07fa8aa81ba59126624bcdf))
- [ ] shell() ([#f3fcf7a](https://github.com/casey/just/pull/2012/commits/f3fcf7aa0805ab161dbd9169e647269b3c24b86c)) - let one write something similar to this: 
  ```just
  var := `ls -hl {{ join('/usr/','bin','python3') }} {{ addsuffix('.rs', 'lib main runner') }}`
  ```
  as
  ```just
  var := shell("ls -hl $1 $2", 'justscript', join('/usr/', 'bin/', 'python3'), addsuffix('.rs', 'lib main runner'))
  ```
- [ ] call() - user-defined functions, when introduced, will benefit from this
- [ ] basename()
- [ ] wildcard()

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -12,6 +12,7 @@ pub(crate) enum Function {
   Nullary(fn(&FunctionContext) -> Result<String, String>),
   Unary(fn(&FunctionContext, &str) -> Result<String, String>),
   UnaryOpt(fn(&FunctionContext, &str, Option<&str>) -> Result<String, String>),
+  UnaryPlus(fn(&FunctionContext, &str, &[String]) -> Result<String, String>),
   Binary(fn(&FunctionContext, &str, &str) -> Result<String, String>),
   BinaryPlus(fn(&FunctionContext, &str, &str, &[String]) -> Result<String, String>),
   Ternary(fn(&FunctionContext, &str, &str, &str) -> Result<String, String>),
@@ -63,6 +64,7 @@ pub(crate) fn get(name: &str) -> Option<Function> {
     "semver_matches" => Binary(semver_matches),
     "sha256" => Unary(sha256),
     "sha256_file" => Unary(sha256_file),
+    "shell" => UnaryPlus(shell),
     "shoutykebabcase" => Unary(shoutykebabcase),
     "shoutysnakecase" => Unary(shoutysnakecase),
     "snakecase" => Unary(snakecase),
@@ -89,6 +91,7 @@ impl Function {
       Nullary(_) => 0..0,
       Unary(_) => 1..1,
       UnaryOpt(_) => 1..2,
+      UnaryPlus(_) => 1..usize::MAX,
       Binary(_) => 2..2,
       BinaryPlus(_) => 2..usize::MAX,
       Ternary(_) => 3..3,
@@ -422,6 +425,31 @@ fn sha256_file(context: &FunctionContext, path: &str) -> Result<String, String> 
     .map_err(|err| format!("Failed to read `{}`: {err}", path.display()))?;
   let hash = hasher.finalize();
   Ok(format!("{hash:x}"))
+}
+
+fn shell(context: &FunctionContext, cmdlike: &str, extras: &[String]) -> Result<String, String> {
+  let (command, args) = context.settings.shell(context.config);
+  let mut shelled_cmd = Command::new(command);
+  shelled_cmd.args(args); // for the shell
+
+  shelled_cmd.arg(cmdlike);
+  if extras.len() > 0 {
+    shelled_cmd.args(&extras[..]);
+  }
+  shelled_cmd
+    .current_dir(&context.search.working_directory)
+    .stdin(Stdio::inherit())
+    .stderr(Stdio::inherit());
+
+  let stdout = InterruptHandler::guard(|| {
+    output(shelled_cmd).map_err(|output_error| output_error.to_string())
+  })?
+  .trim() // as per Make
+  .lines()
+  .collect::<Vec<&str>>()
+  .join(" "); // as per Make
+
+  Ok(stdout)
 }
 
 fn shoutykebabcase(_context: &FunctionContext, s: &str) -> Result<String, String> {
--- a/src/function.rs
+++ b/src/function.rs
@@ -443,11 +443,7 @@ fn shell(context: &FunctionContext, cmdlike: &str, extras: &[String]) -> Result<
 
   let stdout = InterruptHandler::guard(|| {
     output(shelled_cmd).map_err(|output_error| output_error.to_string())
-  })?
-  .trim() // as per Make
-  .lines()
-  .collect::<Vec<&str>>()
-  .join(" "); // as per Make
+  })?;
 
   Ok(stdout)
 }
--- a/src/function.rs
+++ b/src/function.rs
@@ -21,6 +21,7 @@ pub(crate) fn get(name: &str) -> Option<Function> {
   let function = match name {
     "absolute_path" => Unary(absolute_path),
     "arch" => Nullary(arch),
+    "addprefix" => Binary(addprefix),
     "blake3" => Unary(blake3),
     "blake3_file" => Unary(blake3_file),
     "canonicalize" => Unary(canonicalize),
@@ -253,6 +254,16 @@ fn invocation_directory_native(context: &FunctionContext) -> Result<String, Stri
         context.invocation_directory.display()
       )
     })
+}
+
+fn addprefix(_context: &FunctionContext, pref: &str, base: &str) -> Result<String, String> {
+  Ok(
+    base
+      .split(" ")
+      .map(|s| format!("{pref}{s}"))
+      .collect::<Vec<String>>()
+      .join(" "),
+  )
 }
 
 fn join(
--- a/src/function.rs
+++ b/src/function.rs
@@ -22,6 +22,7 @@ pub(crate) fn get(name: &str) -> Option<Function> {
     "absolute_path" => Unary(absolute_path),
     "arch" => Nullary(arch),
     "addprefix" => Binary(addprefix),
+    "addsuffix" => Binary(addsuffix),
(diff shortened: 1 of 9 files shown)


laniakea64:
> shell() -

This would be awesome, thanks!  Would provide a solution for https://github.com/casey/just/issues/11#issuecomment-2015584186 :+1: 

> call() - this might require user-defined functions

Why would this require user-defined functions?  Even without that, seems like `call()` would be a helpful tool.

> wildcard()

Already achievable as
```
foo := `echo *.c`
```

> addprefix()
> 
> addsuffix()
> 
> basename()

These seem easily implemented with `replace_regex()`, and doing it this way is more flexible.  Using the examples from [`make` documentation](https://www.gnu.org/software/make/manual/make.html#index-addsuffix) -
```
addprefix := replace_regex('foo bar', '(^|\s+)(\S+)', '${1}src/${2}')
addsuffix := replace_regex('foo bar', '(^|\s+)(\S+)', '${1}${2}.c')
basename := replace_regex('src/foo.c src-1.0/bar hacks', '(\S+)\.[^/ ]*(\s+|$)', '${1}${2}')
```

gyreas:
> Why would this require user-defined functions? Even without that, seems like `call()` would be a helpful tool.

My bad. I phrased it wrongly; I guess the proper phrasing would be: "`call()` is much useful with user-defined functions". I'll rephrase the description. 

> Already achievable as
> 
> ```
> foo := `echo *.c`
> ```
The point of these functions is to be mixed and matched. So, while `echo *.c` and its variations covers those cases for wildcard(), more complex mixing between `wildcard()` and the others might not be so easily covered. For example, from [make documentation](https://www.gnu.org/software/make/manual/make.html#Call-Function)
```$(wildcard $(addsuffix suf,$(subst :, ,$(PATH))))```
     |
     v
```wildcard(addsuffix("suf", replace_all(":", PATH)))```

> ```
> addprefix := replace_regex('foo bar', '(^|\s+)(\S+)', '${1}src/${2}')
> addsuffix := replace_regex('foo bar', '(^|\s+)(\S+)', '${1}${2}.c')
> basename := replace_regex('src/foo.c src-1.0/bar hacks', '(\S+)\.[^/ ]*(\s+|$)', '${1}${2}')
> ```
It's certainly much readable to simply use `addprefix()`, `addsuffix()`, `basename()`, right?



Thanks for the feedback.

**Comment by laniakea64** [comment]
> The point of these functions is to be mixed and matched. So, while `echo *.c` and its variations covers those cases for wildcard(), more complex mixing between `wildcard()` and the others might not be so easily covered.

With `shell()` implemented, it could be done something like
```
shell('echo $@', addsuffix("/suf*", replace(env_var('PATH'), ":", ' ')))
```

**Comment by gyreas** [comment]
> With `shell()` implemented, it could be done something like
> 
> ```
> shell('echo $@', addsuffix("/suf*", replace(env_var('PATH'), ":", ' ')))
> ```
I guess.

What do you think about using backslash for escaping? (Does just support raw strings?)

**Comment by laniakea64** [comment]
> What do you think about using backslash for escaping?

escaping what?

> (Does just support raw strings?)

Yes - https://just.systems/man/en/chapter_29.html

**Comment by gyreas** [comment]
> escaping what?

Placeholder escaping in `shell()`. I'm using `$\d+` for argument substitution, and I'll naturally need to deal with escaping of $. 


Also, is `$\d+` preferred over `${\d+}` ? (I prefer the former).

**Comment by laniakea64** [comment]
> Placeholder escaping in `shell()`. I'm using `$\d+` for argument substitution, and I'll naturally need to deal with escaping of $.

I think it would be best to leave that to the shell and just pass in the arguments as command-line arguments.  In `just` the shell could be set to anything, even something that isn't nominally a "shell" e.g. a Python interpreter.

So for example, with bash -
```
set shell := ['bash', '-uc']

# in bash the first argument is $0
# provide the justfile as dummy argument so that desired arguments are all in $@
foo := shell('ls -la "$@"', justfile(), 'actual_first_arg')
```
would run `bash -uc 'ls -la "$@"' /path/to/justfile 'actual_first_arg'`

With Python -
```
set shell := ['python3', '-c']

# argument 0 is the `-c` flag from above
foo := shell('''
  import sys
  print(sys.argv[1:])
''', 'argument 1', 'argument 2')
```
would run
```
python3 -c 'import sys
print(sys.argv[1:])
' 'argument 1' 'argument 2'
```

**Comment by laniakea64** [comment]
Realizing that even with `shell()` using command-line to pass arguments, your question applies to `call()`.

There is precedent for escaping `$` and related in [Replacement string syntax of `replace_regex()`](https://just.systems/man/en/chapter_31.html#string-manipulation), where escaped `$` is `$$`.

> Also, is `$\d+` preferred over `${\d+}` ? (I prefer the former).

Usually yes.  `${\d+}` is secondary alternative form for disambiguation - for example, writing `$1` immediately followed by a literal 2 is `${1}2` to distinguish it from twelfth argument `$12`.

**Comment by gyreas** [comment]
> ```
> set shell := ['python3', '-c']
> 
> # argument 0 is the `-c` flag from above
> foo := shell('''
>   import sys
>   print(sys.argv[1:])
> ''', 'argument 1', 'argument 2')
> ```
This...
> ```
> python3 -c 'import sys
> print(sys.argv[1:])
> ' 'argument 1' 'argument 2'
> ```
... and this expose something I haven't thought of. I'll probably go this route since it's similar to how \` \` works. That means shell() is a convenient frontend to \` \`, no?

> I think it would be best to leave that to the shell and just pass in the arguments as command-line arguments. In `just` the shell could be set to anything, even something that isn't nominally a "shell" e.g. a Python interpreter.

Yeah. I'll make the necessary changes and push that.

**Comment by gyreas** [comment]
> ```
> set shell := ['python3', '-c']
> 
> # argument 0 is the `-c` flag from above
> foo := shell('''
>   import sys
>   print(sys.argv[1:])
> ''', 'argument 1', 'argument 2')
> ```
This...
> ```
> python3 -c 'import sys
> print(sys.argv[1:])
> ' 'argument 1' 'argument 2'
> ```
... and this expose something I haven't thought of. I'll probably go this route since it's similar to how \` \` works. That means `shell()` is a convenient frontend to \` \`, no?

> I think it would be best to leave that to the shell and just pass in the arguments as command-line arguments. In `just` the shell could be set to anything, even something that isn't nominally a "shell" e.g. a Python interpreter.

Yeah. I'll make the necessary changes and push that.

**Comment by laniakea64** [comment]
> That means shell() is a convenient frontend to \` \`, no?

It's much better than that:
 - Backticks are fixed strings written in the justfile.  All arguments to `shell()` can (hopefully) be arbitrary `just` expressions, allowing flexible construction of commands and variable parameters.
 - With `shell()` passing in command-line arguments, it's even better than the feature proposed in https://github.com/casey/just/issues/11 .  The `shell()` approach would be more easily resistant to shell code injection and make it more straightforward to guarantee preserving values as intended.

**Comment by gyreas** [comment]
Alright. 

I think we're in sync. 



PS: I think string interpolation can be done via a function fe `fmt()` that uses Rust/Python's formatting scheme. Maybe I should open a discussion for this (or just chime in to #11)?

**Comment by laniakea64** [comment]
> PS: I think string interpolation can be done via a function fe `fmt()` that uses Rust/Python's formatting scheme. Maybe I should open a discussion for this (or just chime in to #11)?

I don't know which option would be best.  In your place I would probably start by replying to https://github.com/casey/just/issues/11#issuecomment-1506818296

File: src/function.rs [hunk 1]
```diff
@@ -402,6 +427,31 @@ fn sha256_file(context: &FunctionContext, path: &str) -> Result<String, String>
...
+    .stdin(Stdio::inherit())
+    .stderr(Stdio::inherit());
+
+  let stdout = InterruptHandler::guard(|| {
+    output(shelled_cmd).map_err(|output_error| output_error.to_string())
+  })?
+  .trim() // as per Make
+  .lines()
+  .collect::<Vec<&str>>()
+  .join(" "); // as per Make
```
laniakea64: Please don't force this problematic make-ism.  `just` does not need strict `make` compatibility, and an individual justfile can easily opt into `make`'s behavior where desired by
```
replace(trim(...), "\n", ' ')
```

I couldn't resist trying your PR (with this output post-processing removed) on one of my justfiles where not having `shell()` was problematic, re-writing the justfile to use `shell()`.  Wow is `shell()` nice to use!  I hope casey agrees when he gets to this PR.  This feature is sorely missing in mainline `just`.  Thanks for sharing your implementation! :grin:

gyreas:
>> +  .trim() // as per Make
>+  .lines()
>+  .collect::<Vec<&str>>()
>+  .join(" "); // as per Make
>
>Please don't force this problematic make-ism.  `just` does not need strict `make` compatibility, and an individual justfile can easily opt into `make`'s behavior where desired by `replace(trim(...), "\n", ' ')`.

Fair enough. I only did that since it's support to be port, I'm just trying to replicate that.

Saed

casey:
I like these, but this PR is pretty big with all of them together. Would you mind splitting them out into individual PRs?

gyreas:
> I like these, but this PR is pretty big with all of them together. Would you mind splitting them out into individual PRs?

Alright. I'll break them apart. Stay tuned!

casey:
> Alright. I'll break them apart. Stay tuned!

Thank you!

laniakea64:
A use case for `wildcard()` that isn't fully available with existing functionality: https://github.com/casey/just/issues/2854

PR #2022, opened 2024-05-02
1748-dotenv-files

sbeckeriv:
Dearest Reviewer,

I was looking at the parser and the format for a list of things [] was already support for shells. I reused that concept to make an ordered list of strings.

This now works
```
set dotenv-files := [".env1", ".env2"]
```

As does the command line
```
just --dotenv-files .env1 --dotenv-files .env2 echo
```

The new setting is applied last. It does not change how the other two settings currently work.

I do have a problem I cant get this to work
```
ENV := "prod"
set dotenv-files := [".env.{{ENV}}", ".env"]
```
it the `{{ENV}}` never turns in to prod.  I am using parse_string_literal.

the issue https://github.com/casey/just/issues/1748 asked about fallback justfiles. I have not used them and I would take any advice on what to change for that as well.

Any advice would be appreciated

Thanks
Becker

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -23,6 +23,7 @@ pub(crate) struct Config {
   pub(crate) color: Color,
   pub(crate) command_color: Option<ansi_term::Color>,
   pub(crate) dotenv_filename: Option<String>,
+  pub(crate) dotenv_files: Option<Vec<PathBuf>>,
   pub(crate) dotenv_path: Option<PathBuf>,
   pub(crate) dry_run: bool,
   pub(crate) dump_format: DumpFormat,
@@ -98,6 +99,7 @@ mod arg {
   pub(crate) const COMMAND_COLOR: &str = "COMMAND-COLOR";
   pub(crate) const DOTENV_FILENAME: &str = "DOTENV-FILENAME";
   pub(crate) const DOTENV_PATH: &str = "DOTENV-PATH";
+  pub(crate) const DOTENV_FILES: &str = "DOTENV-FILES";
   pub(crate) const DRY_RUN: &str = "DRY-RUN";
   pub(crate) const DUMP_FORMAT: &str = "DUMP-FORMAT";
   pub(crate) const HIGHLIGHT: &str = "HIGHLIGHT";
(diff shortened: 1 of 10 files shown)


omar25h:
Not very related to the PR, but to my knowledge, the ".env.{{ENV}}" is working as expected since interpolation only works in recipes. The alternative would be to write something like:

```just
ENV := "prod"
set dotenv-files := [".env."+ENV, ".env"]
```

casey:
Does this need to be a separate setting? I strikes me that `set dotenv-file` could be extended to accept either a single item, or an array of items, with a single item being equivalent to an array containing that single item. Additionally, the `--dotenv-filenname` could be changed to accept multiple items.

sbeckeriv:
> Does this need to be a separate setting? I strikes me that `set dotenv-file` could be extended to accept either a single item, or an array of items, with a single item being equivalent to an array containing that single item. Additionally, the `--dotenv-filenname` could be changed to accept multiple items.

I made the new setting in response to the comment here https://github.com/casey/just/issues/1748#issuecomment-2081443912. 
 
Originally I had setting it more then once here https://github.com/casey/just/pull/1940. 

I could redo dotenv-file take a string or list with the current dotenv-file and accept the command line multiple times. Kind of munging the two PRs together. 

The parse_ordered_list i wrote only works with static strings. I have not figured out how to get an expression working ie: `".env."+ENV`. I would need some pointers or an example on how i can handle that correctly.

casey:
> I made the new setting in response to the comment here [#1748 (comment)](https://github.com/casey/just/issues/1748#issuecomment-2081443912).
> 
> Originally I had setting it more then once here #1940.

I think having a single setting that takes an array is a better idea. Otherwise, the two settings conflict.

> The parse_ordered_list i wrote only works with static strings. I have not figured out how to get an expression working ie: `".env."+ENV`. I would need some pointers or an example on how i can handle that correctly.

It will have to take a static string. dotenv file settings are applied before anything in the justfile is evaluated, so it can't use variables. I did merge #2055 recently, which adds a new kind of string literal which can contain environment variables, so those can be used.

casey:
Also, this needs tests, including tests which cover how different ways dotenv files can be specified interact with each other.

casey:
Converting to a draft so I can keep track of what's ready to go. Feel free to undraft when it's ready!

casey:
Closing, since this has gotten pretty stale.

Issue #1934, opened 2024-03-04
Read a `JUST_VERBOSE` environment variable to always enable the `--verbose` flag


tgross35:
For CI, I run `just --verbose` for everything. Would be great if I could just set the environment to `JUST_VERBOSE=1` :)

casey:
This seems totally reasonable to me. Would be a great first issue if anyone wants to take a stab at a PR.

CramBL:
Close this? 🤓

tgross35:
Ah yeah this is fixed, but not documented in the main README. Probably should update that with a list of the changes in https://github.com/casey/just/pull/2044/files#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecd before closing

casey:
This is sort-of documented [here](https://github.com/casey/just?tab=readme-ov-file#command-line-options). We could add a list of environment variables to the readme, but that might wind up getting out of date, so I just mention to check the output of `--help`.

PR #2044, opened 2024-05-16
Set just flags through environment variables

### LINKED ISSUE #1979: set just flags through environment variables
Hello

It seems like some flags/options passed through parameter, such as `--command-color`, cannot be set by environment variable. It could be good to allow setting parameter also from environment variable (some other tickets refered this subject #1829 and #1702).

I would need the following environment variables:
- `--dry-run` / `JUST_DRY_RUN=1`
- `--quiet` / `JUST_QUIET=1`
- `--verbose` / `JUST_VERBOSE=1`
- `--color=always` / `JUST_COLOR=always`
- `--command-color=green` / `JUST_COMMAND_COLOR=green`
- `--justfile justfile` / `JUST_JUSTFILE=justfile`
- `--working-directory workdir` / `JUST_WORKING_DIRECTORY=workdir`

For example on Gitlab / Github, colors can be forced enabled by setting environment variable `JUST_COLOR=always`

Optionally, I would like to have default values for some flags in the `justfile` itself.

For instance, if the CI variables is defined (in Gitlab), set the JUST_COLOR to alway. And also, always set JUST_COMMAND_COLOR=green no matter the env.

neunenak:
Configure clap to set some configuration options via `JUST_<name>` environment variables.

Resolves https://github.com/casey/just/issues/1979 and https://github.com/casey/just/issues/1934

### DIFF
--- a/README.md
+++ b/README.md
@@ -627,7 +627,6 @@ cc main.c foo.c bar.c -o main
 testing… all tests passed!
 ```
 
-
 Examples
 --------
 
@@ -1395,7 +1394,6 @@ pid:
 $ just
 The process ID is: 420
 ```
-
 
 #### String Manipulation
 
--- a/README.md
+++ b/README.md
@@ -2515,22 +2515,21 @@ $ just --show polyglot
 polyglot: python js perl sh ruby
 ```
 
-The value of some command-line options can be read from environment variables
-of the form `JUST_<option name>`. For instance:
+Some command-line options can be set with environment variables. For example:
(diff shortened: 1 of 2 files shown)


casey:
Looks good to me! Can you add a section in the readme which documents the fact that some options can be set via environment variables? I think we probably shouldn't list the specific options which can be set, since users can look in the `--help` message to see what exactly is supported.

casey:
LGTM!

PR #2045, opened 2024-05-16
Support prepend()

gyreas:
This PR adds a convenient `prepend` function.
 
`prepend(prefix, base)` takes a `prefix` and a `base` string with space-separated words,
and prepends `prefix` to the beginning of each word in the base string.

For example: `prepend('src', 'foo bar baz')` -> 'src/foo src/bar src/baz'


casey:
Seems reasonable to me! How about calling it `prepend`, since that's shorter, and using `split_whitespace` instead of `split(" ")`?

casey:
This should also be documented in the readme (along with addsuffix).

gyreas:
> This should also be documented in the readme (along with addsuffix).

README? Ah! No wonder I feel something is still missing! Alright, I do that across the PRs. Thanks!

> Seems reasonable to me! How about calling it prepend, since that's shorter, and using split_whitespace instead of split(" ")?

That's a shorter name, okay!

casey:
LGTM!

PR #2046, opened 2024-05-16
Support append()

gyreas:
`append(suffix, base)` takes a `append` and a `base` string with space-separated words,
and append `suffix` to the end of each word in the base string.

For example, `append('.c', 'foo bar baz')` -> 'foo.c bar.c baz.c'

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -106,11 +106,10 @@ fn absolute_path(context: &FunctionContext, path: &str) -> Result<String, String
   }
 }
 
-fn append(_context: &FunctionContext, suf: &str, base: &str) -> Result<String, String> {
+fn append(_context: &FunctionContext, suffix: &str, s: &str) -> Result<String, String> {
   Ok(
-    base
-      .split_whitespace()
-      .map(|s| format!("{s}{suf}"))
+    s.split_whitespace()
+      .map(|s| format!("{s}{suffix}"))
       .collect::<Vec<String>>()
       .join(" "),
   )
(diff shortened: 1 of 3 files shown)


src/function.rs
```diff
@@ -105,6 +106,16 @@ fn absolute_path(context: &FunctionContext, path: &str) -> Result<String, String
   }
 }
 
+fn addsuffix(_context: &FunctionContext, suf: &str, base: &str) -> Result<String, String> {
+  Ok(
+    base
+      .split(" ")
```
laniakea64: Would it be better to use [`.split_whitespace()`](https://doc.rust-lang.org/std/primitive.str.html#method.split_whitespace) here for `addprefix()` and `addsuffix()`?

casey:
Seems reasonable to me! How about calling it `append`, since that's shorter, and using `split_whitespace` instead of `split(" ")`?

gyreas:
Shorter names are okay too. This is carried over from when it was about porting make functions.

I'll incorporate the suggestion changes. Thanks for your feedbacks

laniakea64 (commented):
I'm not `just` maintainer, did you mean to request review from casey?  Anyway this looks good to me, `append()` now behaves according to [`make` documentation of `addsuffix`](https://www.gnu.org/software/make/manual/make.html#index-addsuffix) as intended. :+1:

casey:
Nice, LGTM! I added `<sup>master</sup>` in the readme, which indicates that it still hasn't been released. I'll remove that when it makes it into a release.

casey:
Gah, didn't mean to close.

gyreas:
> Nice, LGTM! I added `<sup>master</sup>` in the readme, which indicates that it still hasn't been released. I'll remove that when it makes it into a release.

Got it!

PR #2047, opened 2024-05-16
Support shell()

gyreas:
This PR add support for the `shell()`, which expands to the output of running its given command. 

Example:
```just
set shell := ["fish", "-c"]
v := shell("
set p $argv
for i in $p
	echo "$i"
end
",  "Nightmares", "on", "Elms", "Street")

_:
  @echo "{{ v }}"
```
output:
```console
Nightmares
on
Elms
Street
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1340,6 +1340,53 @@ that work on various operating systems. For an example, see
 [cross-platform.just](https://github.com/casey/just/blob/master/examples/cross-platform.just)
 file.
 
+#### Running External Commands/Scripts
+- `shell(command, args...)` expands to the output of running its given command.
+  - `command` is a script snippet or command. Use the catch-all parameter/argument to get all
+    passed arguments/options, '$@'(bash), '$argv' (fish), 'argv' etc.
+
+  > *Note*
+  > 1. Due to the way Just currently works, `set shell` overrides all assumed shell in `shell()` calls.
+  > 2. Any statements in the script-like command that writes to standard output will affect the output.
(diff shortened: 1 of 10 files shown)


casey:
This seems really nice! What do you think about getting rid of the `name` argument for now? Since a lot of uses of this function won't use it, it seems unfortunate to force users to pass an additional dummy argument if they want to pass arguments. I'd like to enable it in the future with a dedicated named argument syntax, i.e., `shell(CMD, name=NAME, ARGS)`, which is generally useful and can be done in a future PR.

casey:
Also, this should be documented in the readme.

gyreas:
Alright. I clean stuff up, I think I have a solution to remove that inconvenience. Thanks

casey:
I think this should be refactored to use the same code as `Evaluator::run_backtick`. `run_backtick` does a few additional things, like setting the working directory, exporting environment variables, and the destination of stderr.

To make this easier, I just landed #2048, which gets rid of `FunctionContext` and instead just passes the `Evaluator`, since all fields on `FunctionContext` are just taken from `Evaluator` anyways.

The signature of `run_backtick` can be changed to something like:

```
fn run_command(&self, command: &str, args: &[String], token: Option<&Token<'src>>) -> RunResult<'src, String> {
```

Which produces an `Error::Backtick` when `token` is present, and an `Error::FunctionCall` when it isn't.

casey:
Nice, LGTM!

gyreas:
Thank you!

casey:
Thank you! `shell()` is a great addition!

Issue #1991, opened 2024-04-05
Support for secure random sequence generation, with a customizable alphabet


t3hmrman:
Thanks (as always) for the awesome tool!

One thing I'd like to have is secure random sequence generation in `just`. Right now we have `uuid()` which is OK for generating random phrases and stuff, but it would be nice to have a utility that generates a secure random (i.e. from the appropriate OS-provided entropy source) value, of a customizable length, with a customizable alphabet.

This is quite a big feature, but it *would* help remove a bunch of hacks around this that I think many people are doing. Maybe I'm the only one piping `openssl` output through various tools like `tr` trying to get reasonable random values, but it would be nice if this was handled by `just` natively.

casey:
Seems reasonable to me! I think it would be a nice function to add. To keep things simple, I think the first version should just be a function `rand()` which returns a 64-character random hex string. This corresponds to 256 bits of entropy, which is reasonable for almost all cases, and would be a good default. Later, the amount of entropy and alphabet can be customized, perhaps by adding an optional argument, like `rand('01')` if you want a binary string, and `rand('01', 64)` if you want 64 bits of entropy. Syntax TBD.

laniakea64:
Hi, I was curious to see if I could do this and think I might have got somewhere.

[`rand::thread_rng()`](https://docs.rs/rand/latest/rand/fn.thread_rng.html) return value [implements CryptoRng](https://docs.rs/rand/latest/rand/rngs/struct.ThreadRng.html#impl-CryptoRng), which is ["A marker trait used to indicate that an RngCore or BlockRngCore implementation is supposed to be cryptographically secure."](https://docs.rs/rand/latest/rand/trait.CryptoRng.html)  So it seemed reasonable to use that here.

I found it simplest to implement the function as:

`rand(len)` -> return string of `len` random hex digit characters
`rand(len, alphabet)` -> return string of `len` random characters chosen from the characters in `alphabet`

I know this is not quite what casey proposed, so to check:
 - @t3hmrman , would this serve your case?
 - @casey , would this version of `rand()` function be wanted as a PR?

casey:
[...]

PR #2049, opened 2024-05-18
Add function for generating random values

laniakea64:
Adds `pick()` function as discussed in https://github.com/casey/just/issues/1991#issuecomment-2118496078

I'm still relatively new to Rust, so code review would be much appreciated :slightly_smiling_face: 

cc @t3hmrman

### DIFF
--- a/README.md
+++ b/README.md
@@ -1494,9 +1494,10 @@ which will halt execution.
 
 #### Random
 
-- `choose(len)`<sup>master</sup> - Generate a string of `len` random lowercase hexadecimal digit characters.
-- `choose(len, alphabet)`<sup>master</sup> - Generate a random string of length `len` consisting of characters found in `alphabet`.
-  The alphabet must not contain repeated characters.
+- `choose(n, alphabet)`<sup>master</sup> - Generate a string of `n` randomly
+  selected characters from `alphabet`, which may not contain repeated
+  characters. For example, `choose('64', HEX)` will generate a random
+  64-character lowercase hex string.
 
 #### Semantic Versions
 
--- a/README.md
+++ b/README.md
@@ -1492,6 +1492,12 @@ which will halt execution.
 
 [BLAKE3]: https://github.com/BLAKE3-team/BLAKE3/
 
+#### Random
+
+- `pick(len)`<sup>master</sup> - Generate a string of `len` random lowercase hexadecimal digit characters.
+- `pick(len, alphabet)`<sup>master</sup> - Generate a random string of length `len` consisting of characters found in `alphabet`.
+  The alphabet must not contain repeated characters.
+
 #### Semantic Versions
 
 - `semver_matches(version, requirement)`<sup>1.16.0</sup> - Check whether a
--- a/README.md
+++ b/README.md
@@ -1494,8 +1494,8 @@ which will halt execution.
 
 #### Random
 
-- `pick(len)`<sup>master</sup> - Generate a string of `len` random lowercase hexadecimal digit characters.
-- `pick(len, alphabet)`<sup>master</sup> - Generate a random string of length `len` consisting of characters found in `alphabet`.
+- `choose(len)`<sup>master</sup> - Generate a string of `len` random lowercase hexadecimal digit characters.
+- `choose(len, alphabet)`<sup>master</sup> - Generate a random string of length `len` consisting of characters found in `alphabet`.
   The alphabet must not contain repeated characters.
 
 #### Semantic Versions
--- a/src/function.rs
+++ b/src/function.rs
@@ -181,13 +181,7 @@ fn choose(_evaluator: &Evaluator, n: &str, alphabet: &str) -> Result<String, Str
 
   let mut rng = thread_rng();
 
-  let mut s = String::with_capacity(n);
-
-  for _ in 0..n {
-    s.push(*alphabet.choose(&mut rng).unwrap());
-  }
-
-  Ok(s)
+  Ok((0..n).map(|_| alphabet.choose(&mut rng).unwrap()).collect())
 }
 
 fn clean(_evaluator: &Evaluator, path: &str) -> Result<String, String> {
--- a/src/function.rs
+++ b/src/function.rs
@@ -10,8 +10,6 @@ use {
   Function::*,
 };
 
-const DEFAULT_RANDOM_ALPHABET: &str = "0123456789abcdef";
-
 pub(crate) enum Function {
   Nullary(fn(&Evaluator) -> Result<String, String>),
   Unary(fn(&Evaluator, &str) -> Result<String, String>),
@@ -31,6 +29,7 @@ pub(crate) fn get(name: &str) -> Option<Function> {
     "cache_directory" => Nullary(|_| dir("cache", dirs::cache_dir)),
     "canonicalize" => Unary(canonicalize),
     "capitalize" => Unary(capitalize),
+    "choose" => UnaryOpt(choose),
     "clean" => Unary(clean),
     "config_directory" => Nullary(|_| dir("config", dirs::config_dir)),
     "config_local_directory" => Nullary(|_| dir("local config", dirs::config_local_dir)),
@@ -60,7 +59,6 @@ pub(crate) fn get(name: &str) -> Option<Function> {
     "os_family" => Nullary(os_family),
     "parent_directory" => Unary(parent_directory),
     "path_exists" => Unary(path_exists),
-    "pick" => UnaryOpt(pick),
     "prepend" => Binary(prepend),
     "quote" => Unary(quote),
     "replace" => Ternary(replace),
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -1492,6 +1492,12 @@ which will halt execution.
 
 [BLAKE3]: https://github.com/BLAKE3-team/BLAKE3/
 
+#### Random
+
+- `pick(len)`<sup>master</sup> - Generate a string of `len` random lowercase hexadecimal digit characters.
+- `pick(len, alphabet)`<sup>master</sup> - Generate a random string of length `len` consisting of characters found in `alphabet`.
```
t3hmrman: If you're into it, you could add some examples of the use to [the `examples` folder](https://github.com/casey/just/tree/master/examples) -- maybe under the existing kitchen sink ones?

src/function.rs
```diff
@@ -4,7 +4,9 @@ use {
     ToKebabCase, ToLowerCamelCase, ToShoutyKebabCase, ToShoutySnakeCase, ToSnakeCase, ToTitleCase,
     ToUpperCamelCase,
   },
+  rand::prelude::*,
```
t3hmrman: I don't know how people feel about * imports, but if we could avoid it that would be great!

src/function.rs
```diff
@@ -372,6 +375,30 @@ fn path_exists(evaluator: &Evaluator, path: &str) -> Result<String, String> {
   )
 }
 
+fn pick(_evaluator: &Evaluator, count: &str, alphabet: Option<&str>) -> Result<String, String> {
+  let alphabet = alphabet.unwrap_or("0123456789abcdef");
```
t3hmrman: let's move this out to a `const` at the top -- the default alphabet should probably be called out (and possibly reusable) somewhere.

src/function.rs
```diff
@@ -372,6 +375,30 @@ fn path_exists(evaluator: &Evaluator, path: &str) -> Result<String, String> {
...
+  if alphabet.is_empty() {
+    return Err(String::from("no characters in alphabet"));
+  }
+  let mut alphabet_unique = HashSet::<char>::with_capacity(alphabet.len());
```
t3hmrman: I think you can do a `HashSet::from_iter` here! It should be something like:

```rust
HashSet::<char>::from_iter(alphabet.chars())
```

I wonder if it's worth returning an error when we know we'll get the repeated characters (I think ya'll talked about this as well, so I'm definitely fine either way)

src/function.rs
```diff
@@ -372,6 +375,30 @@ fn path_exists(evaluator: &Evaluator, path: &str) -> Result<String, String> {
...
+  }
+  let alphabet = alphabet_unique.drain().collect::<Vec<char>>();
+  let len: usize = match count.parse::<usize>() {
+    Ok(l) => l,
+    _ => return Err(format!("Could not parse `{count}` as a positive integer")),
+  };
+  let mut rng = thread_rng();
+  let mut r = String::with_capacity(len);
+  while r.len() < len {
+    r.push(*alphabet.choose(&mut rng).unwrap());
```
t3hmrman: NIT: you might want to consider checking out the `IteratorRandom` trait in `rand::seq`, I think it might have an example that could make this more concise:

https://docs.rs/rand/latest/rand/seq/trait.IteratorRandom.html#method.choose_multiple_fill

or

https://docs.rs/rand/latest/rand/seq/trait.SliceRandom.html#tymethod.choose_multiple

Fortunately/unfortunately part of Rust is defininely digging in docs (glad they're there!) and finding the 👨‍🍳 💋 APIs to use -- there usually is a nice one for most things people want to do (in well written crates).

src/function.rs
```diff
@@ -372,6 +375,30 @@ fn path_exists(evaluator: &Evaluator, path: &str) -> Result<String, String> {
...
+  let mut alphabet_unique = HashSet::<char>::with_capacity(alphabet.len());
+  for c in alphabet.chars() {
+    if !alphabet_unique.insert(c) {
+      return Err(format!("alphabet contains repeated character {c}"));
+    }
+  }
+  let alphabet = alphabet_unique.drain().collect::<Vec<char>>();
+  let len: usize = match count.parse::<usize>() {
+    Ok(l) => l,
+    _ => return Err(format!("Could not parse `{count}` as a positive integer")),
```
t3hmrman:
```suggestion
    Err(e) => return Err(format!("failed to parse `{count}` as a positive integer: {e}")),
```

Super glad you have the count that failed to parse here -- messages like this are missing the thing that failed to parse way too often in the wild!

t3hmrman:
Hey @laniakea64 thanks for the quick turnaround on this! One more thing that would be awesome are some tests! There's not much you can test (especially as this is real random and not pseudo random), but at the very least the string formation/length might be worth testing!

I wonder if you can easily test entropy (I assume GH runners have enough and this won't be a problem)

src/function.rs
neunenak: That's normally true, but `rand::prelude::*` is designed to be imported with a wildcard like this.

t3hmrman: Yeah -- preludes are usually optional though, I think it's a stylistic/feelings based choice. While the *quick start* section of `rand` and the `rand` book use the prelude import, that's not exclusively how they use it (ex in the ["types of generators" section](https://rust-random.github.io/book/guide-gen.html))

laniakea64: I decided to change it in case the unused imports might have a negative performance impact.

src/function.rs
laniakea64: The reason not to take this type of approach was so that the error message could say which character was repeated.  Is there a way to have both without losing efficiency?

src/function.rs
laniakea64: The docs of that method say it chooses without replacement & I couldn't find a function that chooses with replacement?

src/function.rs
t3hmrman: Ah so this kind of touches on the second point -- do we actually need to tell people which character was repeated? If we say *repeated characters will be de-duplicated* or something to that effect in the doc, I don't think people would mind -- usually `"aabcdef..."` is a bug anyway.

laniakea64:
> Hey @laniakea64 thanks for the quick turnaround on this!

You're welcome, thanks for your review!

> One more thing that would be awesome are some tests! There's not much you can test (especially as this is real random and not pseudo random), but at the very least the string formation/length might be worth testing!

Done.
(I have no actual idea what I'm doing in this area, I just tried to follow the pattern of how existing functions tests were written and this change passed `cargo clippy --all-targets --all-features` and `cargo test` locally.)

> I wonder if you can easily test entropy (I assume GH runners have enough and this won't be a problem)

I have no idea how to even start there, sorry :fearful:

src/function.rs
t3hmrman: Oh sorry you're absolutely right -- it does it without replacement which is not what we want -- there also wasn't anything more convenient than just calling `choose` repeatedly in one form or another.

src/function.rs
laniakea64: Per https://github.com/casey/just/issues/1991#issuecomment-2118523605 and https://github.com/casey/just/issues/1991#issuecomment-2118471256 the characters won't be de-duplicated.

As long as duplicated characters are an error, it would seem helpful to say which character is duplicated, especially since the alphabet could be generated from an arbitrary `just` expression and/or evaluated backtick command output.

t3hmrman: Yup, makes sense to me!

README.md
casey: I think it's fine not to. I don't really require adding examples for every new function. An example can be added to the readme though.

src/function.rs
```diff
@@ -372,6 +377,34 @@ fn path_exists(evaluator: &Evaluator, path: &str) -> Result<String, String> {
   )
 }
 
+fn pick(_evaluator: &Evaluator, count: &str, alphabet: Option<&str>) -> Result<String, String> {
+  let alphabet = alphabet.unwrap_or(DEFAULT_RANDOM_ALPHABET);
```
casey: I prefer not to create constants that are only used in one place. This can just be `let alphabet = alphabet.unwrap_or("0123456789abcdef");`

src/function.rs
```diff
@@ -372,6 +377,34 @@ fn path_exists(evaluator: &Evaluator, path: &str) -> Result<String, String> {
...
+  if alphabet.is_empty() {
+    return Err(String::from("no characters in alphabet"));
```
casey:
```suggestion
    return Err("alphabet is empty".into());
```

src/function.rs
```diff
@@ -372,6 +377,34 @@ fn path_exists(evaluator: &Evaluator, path: &str) -> Result<String, String> {
...
+    Ok(l) => l,
+    Err(e) => {
+      return Err(format!(
+        "failed to parse `{count}` as a positive integer: {e}"
+      ));
+    }
+  };
+  let mut rng = thread_rng();
+  let mut r = String::with_capacity(len);
+  while r.len() < len {
```
casey:
```suggestion
  for _ in 0..len {
```

casey (changes requested):
Looks good! I think we should this function `choose`. I was poking around, and that seems to be the most common name.

src/function.rs
laniakea64: reverted per https://github.com/casey/just/pull/2049#discussion_r1605682328

casey:
LGTM!

I just merged #2054, which adds some predefined constants, including `HEX` which are the lowercase hexidecimal digits. Because `HEX` is now available, I made `choose` require `alphabet`, since it's very short, and maybe a little less confusing to write `choose('10', HEX)` instead of `choose('10')` where hex is implied. I made a couple of other small changes to the function, including using `map_err` to produce the error, and `(0..n).map(…).collect()` to produce the string, to avoid the for loop.

I actually think that, in the general case, pre-allocating the string length isn't useful. The length of a string is in bytes, but characters may be multiple bytes when encoded in UTF-8.

Issue #1955, opened 2024-03-11
Variable substitution in doc comment


halostatue:
It would be useful to have variable substitution in a task doc comment:

```just
# Run the primary CI test (PostgreSQL {{ PROD_POSTGRESQL }})
ci-test-primary: createdb test dropdb
```

This did not work, as it only shows the literal `{{ PROD_POSTGRESQL }}` and does not substitute the variable value (which is set). This would make the output of `just --list` somewhat more dynamic.

casey:
I think the easiest way to do this is to add a `doc` attribute, which allows using an expression as the doc comment:

```just
[doc("Run the primary CI test (PostgreSQL " + PROD_POSTGRESQL + ")")]
ci-test-primary: createdb test dropdb
```

I think this would be relatively easy. When we eventually have interpolation in strings, it could be written as:

```just
[doc(f"Run the primary CI test (PostgreSQL {{ PROD_POSTGRESQL }})")]
ci-test-primary: createdb test dropdb
```

casey:
Added in #3516. The `[doc]` attribute now accepts const expressions.

PR #2050, opened 2024-05-18
`doc` attribute for docstrings

neunenak:
cf. https://github.com/casey/just/issues/1955 , it would be convenient if there was a way to specify a recipe documentation string with an attribute. Also this would make possible to provide a docstring that is different from actual comments in the justfile.

### DIFF
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -453,8 +453,8 @@ impl<'src, D> Recipe<'src, D> {
 
   pub(crate) fn doc(&self) -> Option<&str> {
     for attribute in &self.attributes {
-      if let Attribute::Doc(comment) = attribute {
-        return comment.as_ref().map(|s| s.cooked.as_ref());
+      if let Attribute::Doc(doc) = attribute {
+        return doc.as_ref().map(|s| s.cooked.as_ref());
       }
     }
     self.doc
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -450,6 +450,15 @@ impl<'src, D> Recipe<'src, D> {
       })
       .collect()
   }
+
+  pub(crate) fn docstring(&self) -> Option<&str> {
+    for attribute in &self.attributes {
+      if let Attribute::Doc { docstring } = attribute {
+        return docstring.as_ref().map(|s| s.cooked.as_ref());
+      }
+    }
+    self.doc
+  }
 }
 
 impl<'src, D: Display> ColorDisplay for Recipe<'src, D> {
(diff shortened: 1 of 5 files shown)


casey:
I think this needs to be rebased on master, there are a bunch of extra commits in here.

neunenak:
> I think this needs to be rebased on master, there are a bunch of extra commits in here.

The new code in this PR is on top of https://github.com/casey/just/pull/1842 since that changes how attribute parsing works. I'm not sure if there's a way to tell github to omit commits from another PR in the pull request UI. In any case, https://github.com/casey/just/pull/2050/commits/eda55cc584c3eea8fa3a64d15e402a99d45655bd has the code that is actually new in this PR; I'm fine with not trying to get this merged until 1842 is merged.

src/recipe.rs
```diff
@@ -439,6 +439,15 @@ impl<'src, D> Recipe<'src, D> {
     }
     groups
   }
+
+  pub(crate) fn docstring(&self) -> Option<&str> {
+    for attribute in &self.attributes {
+      if let Attribute::Doc { docstring } = attribute {
+        return docstring.as_ref().map(|s| s.as_str());
```
chrass: .map(|s| s.as_str()) => .map(String::as_str())

casey:
I'm going to cut a release tomorrow, which will be a monster one, since the features have piled up. I think this would be sweet, if you want to get it in. No worries if you don't have time though!

neunenak:
> I'm going to cut a release tomorrow, which will be a monster one, since the features have piled up. I think this would be sweet, if you want to get it in. No worries if you don't have time though!

Yeah, it'd be great to get this in this release, although what I really care about is recipe groups so if this has to wait, not the end of the world.

casey:
Dope!

kranurag7:
@neunenak Thanks for working on this, I'm looking forward to using this one. 

```just
_default:
  @just --list

[doc('this is a long line
which gives usage instruction for 
test recipe')]
test:
  echo "testing"
```

When I run `just` in this directory I get the following output: 
```bash
$ just
Available recipes:
    test # this is a long line
which gives usage instruction for
test recipe
```

It would be great (in terms of readability) if can indent the extra lines as per indentation of the first line. 

```bash
$ just
Available recipes:
    test # this is a long line
           which gives usage instruction for
           test recipe
```
Is there a way I can do this now?

simonmichael:
Is this really needed, when you can also just leave a blank line between comment and recipe ?

Issue #1982, opened 2024-03-29
Add url encoding functions


nyurik:
I was trying to use `curl` from just:

```just
upload credentials file:
   curl '-u{{credentials}}' -T {{quote(file)}} 'https://example.com/foo/bar/{{file_name(file)}}'
```

and this worked fine until the filename had a `#` character - which broke the above because i was not URL-encoding it.  I think just could use a url encoding function to handle such cases

casey:
Seems reasonable to me! Maybe `url_encode(STRING)`. This would make a great first PR for anyone who wants to take a crack at it.

laniakea64:
"URL encoding" can have several different meanings, e.g.
- percent-encoding some characters, like [Javascript `encodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI)
- percent-encoding most characters, notably including characters that have meaning in a URL, like [Javascript `encodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)
- the encoding that replaces some whitespace characters with `+` and does percent-encoding in some other cases, like seen in search engine search URLs
- base64
- ...

Guessing from the example case, you're seeking `encodeURIComponent`-like encoding?

nyurik:
agree, url encoding has a number of different usecases, so it might be useful to provide more than one function.   For my usecase, `encodeURIComponent` is enough. I do not know how common are other usecases - I haven't encountered them (yet)

laniakea64:
Now that https://github.com/casey/just/pull/2052 has been merged and is available in latest `just` release, should this issue be closed?

casey:
Yup, thanks for noticing!

PR #2052, opened 2024-05-18
Add percent encoding function

laniakea64:
https://github.com/casey/just/issues/1982

This function encodes more characters than [Javascript `encodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) because the Examples section of that page mentions some cases that require additional percent-encoding, and in my experience extra percent-encoding like this doesn't hurt anything.

Not happy about the `static` in light of https://github.com/casey/just/pull/2049#discussion_r1605682328 but don't know any way around it :frowning_face:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -536,6 +536,7 @@ dependencies = [
  "libc",
  "log",
  "num_cpus",
+ "percent-encoding",
  "pretty_assertions",
  "regex",
  "semver",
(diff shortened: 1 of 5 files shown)


casey:
My issue with static is just that it requires navigating to the definition of the static to find out what it actually contains, not that it's static per se. I moved the static inside the function, so it's not an issue here.

Regarding which characters to actually percent encode, I think it might be useful to provide functions which exactly match `encodeURI` and `encodeURIComponent`, that way, users aren't surprised by which characters are encoded.

We could also then later add a function which percent encodes even more things, if that winds up being useful. I'd rather percent encode more things based on specific feedback and concrete use-cases.

What do you think?

laniakea64:
Changed to match `encodeURIComponent` behavior.  Per https://github.com/casey/just/issues/1982#issuecomment-2026794722 `encodeURIComponent` would fill the original request.  And should also work for how I might use it.

I don't recall ever using `encodeURI`.

casey:
LGTM! I renamed it to `encode_uri_component`, so that we have the ability to add a more flexible `percent_encode` function later on, perhaps one which takes as input a string containing a custom list of characters to escape.

PR #2058, opened 2024-05-19
Add bash completion for snap

albertodonato:
Also switches the snap to core22

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,4 +1,4 @@
-base: core20
+base: core22
 confinement: classic
 contact: casey@rodarmor.com
 description: Just is a handy way to save and run project-specific commands.
@@ -15,6 +15,7 @@ website: https://just.systems
 apps:
   just:
     command: bin/just
+    completer: just.bash
 
 parts:
   just:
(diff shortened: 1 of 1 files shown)


casey:
I don't really have the ability to test changes to the snapcraft.yaml file, since I don't have an ubuntu machine at the moment, so just to confirm, you tested this and it works?

albertodonato:
Yes, I tested with a locally built snap and it works for me.

casey:
Good enough for me! Thank you!

albertodonato:
Out of curiosity, are you currently publishing the snap manually? AFAICS the last build in the store is quite old

casey:
I am. Or at least I was 😅 I really need to publish a new version.

albertodonato:
FWIW it's possible to set up automation (through the snapcraft site) for the edge channel so that the snap is published on changes to the master branch.

casey:
I tried publishing a new version, but it was pretty annoying, so I gave up and instead removed instructions for installing via the snap in the readme #2070 😅 I think that `just` may actually now be available in Ubuntu 24.04 using `apt`.

zekefast:
May be it could be useful to tests changes like this.

There is a project called [distobox](https://github.com/89luca89/distrobox) which allows to spin up different distributions very quickly. I found it very useful for myself. May be you will too.

PR #2060, opened 2024-05-20
Improve `shell()` documentation

laniakea64:
- ~~Document that some shells will assign the first argument to `$0` and thus it won't be in `$@`, and add an example demonstrating this~~
- Fix bugs in example code
- Expand the examples to showcase `shell()`'s power more
- Clarify one comment

### DIFF
--- a/README.md
+++ b/README.md
@@ -1342,7 +1342,7 @@ file.
 
 #### External Commands
 
-- `shell(command, args...)` returns the standard output of shell script
+- `shell(command, args...)`<sup>master</sup> returns the standard output of shell script
   `command` with zero or more positional arguments `args`. The shell used to
   interpret `command` is the same shell that is used to evaluate recipe lines,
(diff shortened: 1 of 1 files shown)


casey:
Ahh, good catch. When running linewise recipes, we actually supply the command as the first argument, so that positional arguments appear in the expected variables. We should consider copying this behavior for the `shell()` function. It would be odd if:

```
x := shell('echo $1', 'bar')

foo bar:
   echo $1
```

`x` and the output of `just foo bar' were different.

casey:
I just merged #2061 which passes the command being run as the first argument to `shell()`. I think the other changes here are good though, if you want to rebase.

laniakea64:
> I think the other changes here are good though, if you want to rebase.

Done.  Also added which `just` version `shell()` was introduced.

casey:
Merged!

PR #2064, opened 2024-05-20
Compute default recipe on `Justfile`

neunenak:
If a `Justfile` value stores its own root path, it can compute the default recipe as required, which makes the analyzer simpler and also makes it easier to change what the rules are for determining the default recipe in the future.


casey:
This doesn't personally seem like a big win to me. It allows getting rid of `keyed::serialize_option`, but it removes `first` from the serialization of `Justfile`, which breaks backwards compatibility with the JSON dump format.

Issue #2078, opened 2024-05-23
Add timestamps to each command printed


nogweii:
Using `just` in CI, it would be pretty handy to have a bunch of timestamps on each printed line of the recipe to see how long each step takes. Possibly by opting into it with a new `--timestamps` flag?

casey:
That seems reasonable to me!

starthal:
Closed by #2084 ?

casey:
Whoops, yup! Thanks for noticing!

PR #2084, opened 2024-05-25
Print timestamps when --timestamps is passed

neunenak:
Print timestamps when the --timestamps flag is passed

https://github.com/casey/just/issues/2078


neunenak:
Thought I'd take a crack at https://github.com/casey/just/issues/2078 . 

```just


recipe:
    echo "ONE"
    sleep 10
    echo "TWO"
```

```
$ just --timestamps
[2024-05-25T07:41:28.376232233+00:00] echo "ONE"
ONE
[2024-05-25T07:41:28.377825276+00:00] sleep 10
[2024-05-25T07:41:38.380184613+00:00] echo "TWO"
TWO
```

A couple of questions:

* this feature doesn't seem all that useful unless there's a way to configure the format/time zone/etc. of the printed timestamp. The above is just printing the default `chrono` ISO8601 formatted timestamp. I'm not sure what the best way to configure this is - another option `--timestamp-format <format-string>` that gets passed directly into `chrono`? Have `--timestamp` accept an optional format string argument? something else?

* this could take the short argument `-t`, but I'm not sure if it's worth burning that single character on this as opposed to some other more useful option that could start with t

casey:
> * this feature doesn't seem all that useful unless there's a way to configure the format/time zone/etc. of the printed timestamp. The above is just printing the default `chrono` ISO8601 formatted timestamp. I'm not sure what the best way to configure this is - another option `--timestamp-format <format-string>` that gets passed directly into `chrono`? Have `--timestamp` accept an optional format string argument? something else?

Yah, agreed. The default formatting is super long. From the original issue, it sounds like the main thing that they wanted to know was how long each command took. We could consider using a default format that only prints the time, like `16:55`, which would allow figuring out how long things took, and letting them pass `--timestamp-format` if they want to customize it. I don't think `--timestamp` can have an optional format argument, otherwise `just --timestamp RECIPE` would swallow the recipe.

I'd add `--timestamp` first and do `--timestamp-format` in a follow-up PR.

> * this could take the short argument `-t`, but I'm not sure if it's worth burning that single character on this as opposed to some other more useful option that could start with t

I'm always very stingy about short arguments, so I'd leave it out and wait for someone to complain that they use it a lot.

`--timestamp` should also be settable with an env var, i.e., `JUST_TIMESTAMP`.

casey:
I removed subseconds from timestamps. My hunch is that this will be noisier for not much benefit. Timestamp output is human readable, so not covered by backwards compatibility guarantees, so we can add it back later if people want it.

Issue #1263, opened 2022-07-08
Allow unsetting environment variables


nickbabcock:
Just has support for setting environment variables:

```
export RUST_BACKTRACE := "1"

test:
    cargo test
```

But does not support unsetting environment variables

```
unset RUST_BACKTRACE

test:
    cargo test
```

In the ideal scenario, the environment variable has an equivalent value to being unset (0 in the case for `RUST_BACKTRACE`)

Unfortunately, some programs, crates, and/or libraries treat an unset environment variable different than even an empty string, so it would be helpful to be able to unset these at the global level instead of remembering to sprinkle in the manual method everywhere:

```
test:
    #!/bin/bash -x
    unset RUST_BACKTRACE
    cargo test
```

casey:
This definitely seems useful, and the `unset FOO` seems reasonable. The best way to implement this is probably with a new `Item` variant, `Item::Unset(String)`, which causes exported variables to be unset.

neunenak:
Should be closeable now with https://github.com/casey/just/pull/2098

PR #2098, opened 2024-05-26
Implement unexporting variables

neunenak:
https://github.com/casey/just/issues/1263


casey:
Just took a look at this, mostly looks good. I think we should call this `unexport`, since `unset` sounds like it has something to do with settings.

There are a few tests we should have:
- Duplicate `unexport` statements should be an error. If people want duplicate unsets, for whatever reason, maybe because they're importing them from modules, we can relax this, but can't change things in the opposite direction, i.e., make duplicate an error when it hadn't been before.
- Having an `export` and `unexport` for the same variable should be an error. Similarly, this is easy to relax in the future, but we couldn't later make it an error if it wasn't one before.
- Recipes can export parameters with `foo $VAR:` syntax. I'm inclined to not make this an error, and we should have a test which checks that if you unexport a variable at the top level, but export it in a function parameter, it gets set.

casey (changes requested):
Doing an actual review so it shows up as requested changes in my PR lists.

casey (approved):
Dope, LGTM!

Did some very minor stuff. For example, you can do `${VAR:-unset}`, to allow figuring out if a variable is unset without triggering an error, which is useful for tests.

Issue #2093, opened 2024-05-26
`dotenv-path` causes error if the specified envfile does not exist


FelixZY:
**Steps to reproduce:**
1. Create this `justfile` in an empty directory:
    ```just
    set dotenv-load := true
    set dotenv-path := ".env"

    sayhi:
        @echo "Hello world"
    ```
2. execute `just sayhi`

**Expected**

> Hello World

**Actual**

```
error: Failed to load environment file: No such file or directory (os error 2)
```

**Additional context**
* just version 1.27.0 (cargo)
* setting `dotenv-load` to `false` still reproduces the issue
* [Documentation for `dotenv-path`](https://github.com/casey/just#table-of-settings) states (my emphasis):
  > Load a .env file from a custom path, **if present**."

casey:
I just tested this, and it looks like `dotenv-path` has always failed if the environment file isn't present. Whether or not this is the best behavior is up for debate. On the one hand, if you have a truly optional `.env` file, then it would be nice to run if it isn't found. On the other hand, if the `.env` file is required for recipes to run correctly, you would want an error.

For now I'm going to update the docs, and leave this issue open to collect feedback.

FelixZY:
In my case, the env file is required for most actions. However, I have some additional requirements (specifically, docker compose secrets) which makes setting up the env file manually a bit of a pain. For this reason, I have a special task to help the user generate their env file, which is placed in a special location. 

If `dotenv-path` did not care about the presence of the env file, as originally documented, my users could start using `just` from a fresh clone. Now though, there is a manual step required first in creating an empty env file. 

Perhaps you're right that `dotenv-path` and `dotenv-filename` should not define the behaviour for the case of a missing env file. However, I really think there should be an option to support cases where `just` itself is used to generate the envfile (using `just` to interact with [`vercel env`](https://vercel.com/docs/cli/env) is yet another example of this)! Perhaps something like `dotenv-optional` or `dotenv-fallback: ignore/fail`

FelixZY:
Also, forcing the presence of an env file seems a bit weird in general. The env file is supposed to offer one *alternative* way of defining environment variables. It does not necessarily have to be *the* way.

Consider a CI environment such as GitHub actions where environment variables are pre-defined. If `just` throws an error simply because the env file does not exist, I'm now forced to add an extra step of setting up an env file, despite my environment having been correctly configured from the start.

casey:
[...]

PR #2099, opened 2024-05-26
Clarify that `dotenv-path`-given env file is required

casey:


README.md
```diff
@@ -862,7 +862,9 @@ bar
 If `dotenv-load`, `dotenv-filename` or `dotenv-path` is set, `just` will load
 environment variables from a file.
 
-If `dotenv-path` is set, `just` will look for a file at the given path.
+If `dotenv-path` is set, `just` will look for a file at the given path. It is
+an error if a dotenv file is not found at `dotenv-path`, but not an error if a
+dotenv file is not found with `dotenv-filename`.
```
FelixZY, line 867: What is the motivation behind the difference in logic?

casey: This is just documenting the existing behavior. I'm not sure what the original motivation was.

FelixZY: Personally, I find it confusing that `dotenv-path` and `dotenv-filename` have this kind of behavioural difference. I think it would be better if both parameters had the same behaviour.

---

In a more general context, I find the presence of both `dotenv-filename` and `dotenv-path` a bit confusing. Isn't `dotenv-filename` basically a more limited version of `dotenv-path` (ignoring this specific difference in behaviour)? Perhaps it would make sense to deprecate `dotenv-filename` (removing it outright would probably break some justfiles)?

FelixZY:
For future readers: see also #2093 which prompted this PR

PR #2104, opened 2024-05-28
Remove no longer needed bash completion replacement

### LINKED ISSUE #2094: 1.27.0 broke completion when using alias
Hi,

I am used to alias `just` to a single letter (`k` in my case).

To have recipe completion, I have added these lines to my ~/.bashrc:

```bash
source <(just --completions bash)
complete -F _just -o bashdefault -o default k
```

So `k <tab>` would give the same result as `just <tab>`.

However, since the update to 1.27.0, `k <tab>` displays a list of files in the current dir, not a list of available recipes as was previously the case.

I can't figure out where this difference comes from, because it seems that the only change has been the addition of the `--groups` option in `just.bash`.

Am I missing something?

Please note the issue doesn't seem to be related to installation procedure, as I have tried both `pacman -S just` and `curl --proto...`

Many thanks,
L.

laniakea64:
Fixes https://github.com/casey/just/issues/2094

### DIFF
--- a/completions/just.bash
+++ b/completions/just.bash
@@ -29,7 +29,7 @@ _just() {
     done
 
     case "${cmd}" in
-        "$1")
+        just)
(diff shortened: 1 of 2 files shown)


casey:
Nice, LGTM!

PR #2106, opened 2024-05-29
Timestamp format

neunenak:


neunenak:
This PR is failing the test `config::tests::show_arguments` because the clap-generated Usage: string insists on adding `--timestamp-format`, and I'm not sure why

casey:
`.arg("--timestamp-format FORMAT")` passes the single arg `"--timestamp-format FORMAT"`, it needs to be `.arg("--timestamp-format").arg("FORMAT")` or `.args(["--timestamp-format", "FORMAT")`. I just pushed a commit fixing it.

neunenak:
> `.arg("--timestamp-format FORMAT")` passes the single arg `"--timestamp-format FORMAT"`, it needs to be `.arg("--timestamp-format").arg("FORMAT")` or `.args(["--timestamp-format", "FORMAT")`. I just pushed a commit fixing it.

Ah thanks for fixing that

casey:
- I changed the flag to `--timestamp`. I noticed in the examples in the readme, you used `--timestamp`, which I think would be a common error, and `--timestamp` is shorter anyways.
- I removed the `--ts` alias for now. If it's a commonly used flag, I'd rather just give it a short flag, like `-t` or `-T`.
- `TIMESTAMP_FORMAT` has a default value, so it can just be unwrapped.

casey:
Oh yeah, also made it local time, since I figure that's more useful, and if it's local time, they can at least configure it by changing the time zone of their system, whereas if it's UTC, they can't configure it.

Issue #929, opened 2021-07-29
Modules stabilization tracking issue


casey:
This issue tracks stabilization of `mod` statements.

There are a bunch of features which aren't yet implemented, but I think they can all be implemented in a backwards compatible fashion:

- Recipes defined in one module can be called from another
- Variables defined in one module can be referenced from another
- Allow inline modules that are declared and defined in a single file.
- Allow submodules to opt-in to using parent module settings
- Allow submodules to opt-in to loading `.env` files
- #2119
- Make non-compilation errors refer to modules source file
- Make `just --show` work with modules

I want the feature to be well used and well tested before stabilizing it.

- [x] #1901
- [x] #1848
- [x] #2105 
- [x] #2107
- [x] #1866

matthewtgilbride:
Hi @casey, curious if this project is inspired to any extent by [mmake](https://github.com/tj/mmake)?  For me, the most painful thing that traditional make lacks is support for [remote includes](https://github.com/tj/mmake#remote-includes).  

For example, we have built out a set of common makefiles at my job for common projects (e.g. ExpressJS/Node apps, Spring Boot apps, AWS Lambda with SAM CLI, etc.).  We also have a core makefile that exports handy commands to print messages in certain colors, require certain environment variables, etc.

The most annoying part of this setup is that we have to ask people to copy/paste files into their repos.  We even have an `update-makefiles` command so that once they do so, they can consume updates without friction.

I just wanted to throw this out there.  I know there is a lot of thought being put into a module system in general, but allowing for remote modules would be _awesome_.

Thanks for all the hard work on a really cool tool :)

casey:
> Hi @casey, curious if this project is inspired to any extent by [mmake](https://github.com/tj/mmake)?

I didn't know about `mmake` but it seems to be in a similar vein, although `just` pre-dates it by a couple years. In fact, `just` also used to be a wrapper-script around `make`. I've been meaning to add a list of alternative command runners to the readme for a while, so I just opened https://github.com/casey/just/pull/1008 which mentions `mmake`.

> For me, the most painful thing that traditional make lacks is support for [remote includes](https://github.com/tj/mmake#remote-includes).

That's a super interesting feature. We don't have modules, but if we ever do, I wouldn't be opposed to adding something like that.

[...]

PR #2113, opened 2024-05-30
Allow listing recipes in submodules with `--list-submodules`

casey:
@crdx Do you prefer a nested view, like this:

```
Available recipes:
  foo
  bar

  abc:
    xyz
    wer

  hjk:
    yui
    asd
```

Or a flattened view, like this:

```
Available recipes:
  foo
  bar
  abc::xyz
  abc::wer
  hjk::yui
  hjk::asd
```

### DIFF
--- a/completions/just.bash
+++ b/completions/just.bash
@@ -30,7 +30,7 @@ _just() {
 
     case "${cmd}" in
         just)
-            opts="-n -f -q -u -v -d -c -e -l -s -E -g -h -V --check --chooser --color --command-color --yes --dry-run --dump-format --highlight --list-heading --list-prefix --no-aliases --no-deps --no-dotenv --no-highlight --justfile --quiet --set --shell --shell-arg --shell-command --clear-shell-args --unsorted --unstable --verbose --working-directory --changelog --choose --command --completions --dump --edit --evaluate --fmt --init --list --groups --man --show --summary --variables --dotenv-filename --dotenv-path --global-justfile --timestamp --timestamp-format --help --version [ARGUMENTS]..."
+            opts="-n -f -q -u -v -d -c -e -l -s -E -g -h -V --check --chooser --color --command-color --yes --dry-run --dump-format --highlight --list-heading --list-prefix --list-submodules --no-aliases --no-deps --no-dotenv --no-highlight --justfile --quiet --set --shell --shell-arg --shell-command --clear-shell-args --unsorted --unstable --verbose --working-directory --changelog --choose --command --completions --dump --edit --evaluate --fmt --init --list --groups --man --show --summary --variables --dotenv-filename --dotenv-path --global-justfile --timestamp --timestamp-format --help --version [ARGUMENTS]..."
                 if [[ ${cur} == -* ]] ; then
                     COMPREPLY=( $(compgen -W "${opts}" -- "${cur}") )
                     return 0
--- a/completions/just.elvish
+++ b/completions/just.elvish
@@ -46,6 +46,7 @@ set edit:completion:arg-completer[just] = {|@words|
             cand -n 'Print what just would do without doing it'
             cand --dry-run 'Print what just would do without doing it'
             cand --highlight 'Highlight echoed recipe lines in bold'
+            cand --list-submodules 'List recipes in submodules'
             cand --no-aliases 'Don''t show aliases in list'
             cand --no-deps 'Don''t run recipe dependencies'
             cand --no-dotenv 'Don''t load `.env` file'
--- a/completions/just.fish
+++ b/completions/just.fish
@@ -57,6 +57,7 @@ complete -c just -l check -d 'Run `--fmt` in \'check\' mode. Exits with 0 if jus
 complete -c just -l yes -d 'Automatically confirm all recipes.'
 complete -c just -s n -l dry-run -d 'Print what just would do without doing it'
 complete -c just -l highlight -d 'Highlight echoed recipe lines in bold'
+complete -c just -l list-submodules -d 'List recipes in submodules'
 complete -c just -l no-aliases -d 'Don\'t show aliases in list'
 complete -c just -l no-deps -d 'Don\'t run recipe dependencies'
 complete -c just -l no-dotenv -d 'Don\'t load `.env` file'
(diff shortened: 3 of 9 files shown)


valscion:
If I may add my preference? I'm in favor of the nested view as I'm not that comfortable showing `::` anywhere 😅 

The way we've used the "module-like" just scripts so far has relied on the separator being a space character and I'd like to avoid seeing the `::` in any output 😅 

I can see us using `--list-submodules` in some inner module, maybe already with the `just devices` case I pointed in https://github.com/casey/just/issues/929#issuecomment-2132780998  — but not at the top-level `just` case.

The `::` separator feels like some Rust style preference to me and not something I've seen in any other command line tool we use.

The old way of using "module-like" just scripts has also relied only on whitespace separated modules so we don't have any usage of e.g. `just devices::android::open-url` in the past but `just devices android open-url` instead.

I do agree that it's slightly easier to read the output when the list contains of `abc::xyz` as it's shorter — but I don't think there would be too many submodules to list.

---

One more question: Should this have some sort of depth limit option as well? What if there are submodules on 4 levels — should all of them be shown or would it be possible to limit the depth somewhere? I don't know if that would be useful to have but I bet it would be asked at some point in the future 😅

casey:
The issue with spaces in a flattened view is that it wouldn't be possible to tell what was a submodule/recipe and what was an argument:

```
Available recipes:
  foo
  bar
  abc xyz
  abc wer
  hjk yui
  hjk asd
```

`abc` could be a recipe with an argument named `xyz`. The `::`-separated notation is also necessary in few places. I chose it because `::` is commonly used to separate module paths, e.g., in Rust, C++, and I think others.

The output of `--summary` uses spaces to separate recipe names, and completion scripts split on spaces, so in order to include recipes in submodules, they must be `::`-separated. Because of this, `::` must also be accepted on the command line, so that a recipe from the output of `--summary` can be passed to `just`.

When, eventually, you can reference variables and recipes in other modules, you'll have to specify module paths using `::`. This is just because the grammar would otherwise be ambiguous between submodule recipes and arguments:

```justfile
mod foo

bar: foo::baz
```

So the command-line interfaces will all accept `::` as an alternative to spaces, and when module paths eventually appear in `justfiles` themselves, they'll be space-separated. I do agree that spaces are nicer!

I'll probably add some sort of option to list submodules nested, as in this PR, and flattened, in another PR, using `::`.

As far as a depth limit, I'd be happy to add that in the future, feel free to make an issue if/when you want it!

valscion:
Yeah to clarify my opinion — I don't have any objections of using `::` _inside_ Justfiles themselves. I do think it makes a lot of sense to use `::` inside Justfiles.

I'm mostly against showing that in the output of any `just` commands when ran on the command line. (At least those which are intended to be used regularly and not for mostly debugging purposes)

I also do agree that the flattened view as in your example would be confusing and would not want that either.

It is indeed a tricky case! 

> The output of `--summary` uses spaces to separate recipe names, and completion scripts split on spaces, so in order to include recipes in submodules, they must be `::`-separated. Because of this, `::` must also be accepted on the command line, so that a recipe from the output of `--summary` can be passed to `just`.

I'm not sure I fully follow this train of thought, so let me try to understand what you mean with this example:

```
Available recipes:
  foo
  bar

  abc:
    xyz
    wer

  hjk:
    yui
    asd
```

I would image `just <TAB>` would show in autocompletion:

```
foo
bar
abc
hjk
```

And then I'd do `just abc <TAB>` and see:

```
xyz
wer
```

But it sounded like what I'd have instead is this when running `just <TAB>`?

```
foo
bar
abc::xyz
abc::wer
hjk::yui
hjk::asd
```

Do correct me if I'm wrong! I'm mostly worried that the tab autocompletions would get useless for us if all the submodule recipes would also be shown in there with the `::` separating them instead of needing "diving deeper" like we now have since https://github.com/casey/just/pull/2108

crdx:
@casey The original nested output is my personal preference. I wouldn't use the flattened view.

I don't use the concept of a default recipe, i.e. I don't have `just` on its own run anything. Instead, the first recipe is always:

```
[private]
help:
    just --list --unsorted
```

Because I prefer to see what's available and pick from the list rather than have a default. So, I can easily add a new flag `--nested` (or whatever) to this recipe everywhere.

@valscion I believe that the output of `--summary` is correct to show fully-qualified recipes. Technically a submodule recipe is still a recipe that can be run from the top-level context so completions should show it. And it needs to be `::`-separated because of completion word-splitting characters.

The current shell completions are generated by applying simple replacements to the clap-generated default completions:

https://github.com/casey/just/blob/f5bb82dea3edde07b5aca4303e2d73a300f45d5e/src/completions.rs

I wouldn't personally like to try to implement more intelligent submodule recipe completion handling like this. This is the point where hand-crafted completion scripts would be better.

valscion:
> Technically a submodule recipe is still a recipe that can be run from the top-level context so completions should show it.

Yeah I do agree — but this "technically correct" vs. what I'm in favor of as "intuitive behavior" are different.

For us, it would make the autocompletion cumbersome and the first `tab` hit after entering `just <TAB>` would be useless as we'd see hundreds of potential autocompletions. At that point, we would not see the rest of the commands available at all.

For example, the `aws` CLI (which is huge!) also doesn't show "submodules" but instead lists the outermost modules first:

image

Then I can tab to just the `aws s3` scripts:

image

This is how I'd expect "top level modules" to first work and then submodules showing only after I've explicitly tabbed to show the submodules.

casey:
> The original nested output is my personal preference. I wouldn't use the flattened view.

Okay, nice. I'm going to merge this, if someone wants the flat view, they can open an issue.

The completion scripts are extremely hacky. `just` uses clap to parse arguments, and clap generates the basic completion scripts. We then modify them to make them complete recipe names with the output of `just --summary`, which includes `::`-separated recipe paths in modules.

I would prefer to use space-separated recipe paths, and I would prefer to complete them one at a time, but this just isn't possible with the current approach, since the completion scripts have no idea which arguments are present on the command line so far, so they can't incrementally complete recipes and modules as the user descends into modules.

Fixing this is blocked on https://github.com/clap-rs/clap/issues/1232, which is a proposed improvement to claps's completion engine. I suppose someone could try modifying the completion scripts, but that gets messy fast.

valscion:
> I would prefer to use space-separated recipe paths, and I would prefer to complete them one at a time, but this just isn't possible with the current approach, since the completion scripts have no idea which arguments are present on the command line so far, so they can't incrementally complete recipes and modules as the user descends into modules.
> 
> Fixing this is blocked on https://github.com/clap-rs/clap/issues/1232, which is a proposed improvement to claps's completion engine. I suppose someone could try modifying the completion scripts, but that gets messy fast.

Oh wow what a rabbit hole — I understand now that this isn't such an easy feat 😅

PR #2117, opened 2024-06-01
add color display for default recipe

cache-missing:


casey:
I'm not sure about this. I do think that marking the default recipe somehow is useful, but I feel like the purple highlight is pretty inscrutable. Maybe create an issue to collect feedback on the best way to mark the default recipe?

cache-missing:
Of course it useful , i always put my frequency used recipe as default , and when i have a long time not work on a project , i always forget what is the  default  recipe , so i want identify default recipe use highlighting

PR #2118, opened 2024-06-01
add new function indent and add new token type uint

cache-missing:


laniakea64:
Why?  This is already easy to do with greater flexibility with [`replace_regex()`](https://just.systems/man/en/chapter_32.html#string-manipulation).  For example, to indent with two spaces: 
```
replace_regex(..., "(?m)^", "  ")
```

uint token type seems awkward and out of place.  Everything being a string is a feature of `just`, and `just` shouldn't panic on error in a justfile.  It would be better to take all string arguments, fallibly parsing the string argument as an integer and giving a helpful error message if that fails, like [`choose()` does](https://github.com/casey/just/blob/f2201d8684bdf9fd6dfe7a3fb18e481feb4f4cfe/src/function.rs#L187-L204).

casey:
I think I agree with @laniakea64's assessment here. This can be easily and more flexibly done with a regex replace, and introducing an integer token type just for this function feels unnecessary.

PR #2127, opened 2024-06-05
Remove duplicated section in cn readme.

potterxu:
Remove duplicated section 1.6.6
image

it is already included in section 1.6.4 (which aligns with the en readme)
image


casey:
Nice, thank you!

Issue #1985, opened 2024-03-29
Bug: when recipe repeats recipe inside it, it only runs the first time


whyboris:
Example `justfile`:

```sh
lol: hi hello hi

@hi:
  echo "1"

@hello:
  echo "2"
```

Expected output from `just lol`

```
1
2
1
```

Actual output from `just lol`

```
1
2
```

I think this is a bug because I had a command that did cleanup between various other commands and I spent 2 hours trying to figure out why cleanup wasn't working 😢

casey:
This is actually as intended, although I can see why it would be confusing. Just will only run normal recipes once, even if they appear more than once on the command line, or more than once in recipe dependencies. Just also doesn't necessarily run recipes in the same order as they're passed on the command line. If you do `just a b`, and `a` is a dependency of `b`, it will run `b` first and then `a`.

whyboris:
I suggest adding a quick note in the relevant section:

https://github.com/casey/just?tab=readme-ov-file#features

> You can also use dependencies to run multiple recipes by default:

```md
_Note:_ just will not run a recipe twice
```

The model I had in my head was "recipe", meaning a step-by-step instruction of what `just` will do. I'm puzzled why anyone would _expect_ `just` to _ignore_ an instruction.

I am not insisting on implementing changes - I super ❤️ love `just` ❤️ -- I just think that even though my use case is unusual, there may be others who will have cleanup steps in their script.

My example: https://github.com/whyboris/utilitarianism.net/blob/main/justfile
I build a website in three steps: search, PDF, and production - so I had to duplicate a cleanup step :trollface:

casey:
I think that's reasonable, I'll reopen this issue to track adding a note to the docs.

colemannerd:
@casey I think this issue is now closed.

quebin31:
I just stumbled upon this behavior with a use case where I wanted to run multiple recipes, but needed at least one "clean" in between each of them, unfortunately only the first invocation ran. This got me thinking, this shouldn't be so hard to support, right? How do you like the idea of adding a new recipe attribute like `[no-cache]` or `[always-run]` that would allow a recipe to skip this behavior and always run whenever it's called?

PR #2134, opened 2024-06-06
Updated comment to clarify that just won't execute multiple commands

colemannerd:
resolves issue #1985 

Since this is the first time I'm contributing, I am happy to fix changes!


casey:
Thanks for the PR!

I think this is actually already documented in the readme [here](https://github.com/casey/just/?tab=readme-ov-file#dependencies).

Issue #2086, opened 2024-05-25
Add `[ATTRIBUTE: VALUE]` shorthand


casey:
`[ATTRIBUTE: VALUE]` would be the same as `[ATTRIBUTE(VALUE)]`, but avoids the parens.

casey:
This has been done!

PR #2136, opened 2024-06-07
Add [attribute: value] shorthand

neunenak:
https://github.com/casey/just/issues/2086

Adds a shorthand syntax `[attribute: value]` for attributes with a single argument, that is equivalent to `[attribute(value)]`.

### DIFF
--- a/src/attribute.rs
+++ b/src/attribute.rs
@@ -42,7 +42,7 @@ impl AttributeDiscriminant {
 impl<'src> Attribute<'src> {
   pub(crate) fn new(
     name: Name<'src>,
-    mut arguments: Vec<StringLiteral<'src>>,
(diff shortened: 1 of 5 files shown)


casey (changes requested):
I like the `[attribute: value]` shorthand, but I don't think we should parse multiple attribute values until we actually have an attribute that takes multiple values, so Attribute:new should be changed back to take an `Option<StringLiteral>`

casey (changes requested):
One more thing: Can you add a separate `[attribute: value]` test instead of changing the existing group test? I think we can have a single test which tests the short syntax, and leave the other tests as they are.

casey (approved):
LGTM!

PR #2138, opened 2024-06-08
Some evaluator simplifications

neunenak:
This PR modifies some of the methods on `Evaluator` that accept a large number of arguments to instead accept a `RecipeContext`, which already contains most of those arguments, which allows the call sites to be more concise. 

It also deduplicates some of the code around running recipe dependencies, and replaces a mutable variable update with a fold expression.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,6 +1,7 @@
 use super::*;
 
-pub(crate) struct RecipeContext<'src: 'run, 'run> {
+#[derive(Copy, Clone)]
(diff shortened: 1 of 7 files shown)


casey:
Nice, this is a good refactor. I took it a step further by just having the `RecipeContext` (now `ExecutionContext`, because it's used all over the place) be a field on the evaluator. This required some tweaking, in particular, there was a scope that was created outside a call to the evaluator that is now created inside the evaluator.

Issue #1930, opened 2024-02-28
Can a recipe know if it has been invoked directly or as a depedency of another?


nogweii:
Per the question, if not, I think it would be a useful function to add.

For what I'm thinking, I'd like to do the following:

```just
run: install
  ./start_server.py

install:
  poetry install {{ if is_dep() { '--quiet' } else { '--verbose' } }}
```

That is, if someone were to run `just install` manually, they'd get the full output from `poetry install`. If a user runs `just run` instead, poetry would have reduced output.

casey:
This isn't currently possible, but seems pretty reasonable to me. I think the implementation would be pretty simple, in case anyone wants to take a crack at it.

laniakea64:
In the mean time, here's a possible workaround? -
```
run: (install '--quiet')
  ./start_server.py

install flag='--verbose':
  poetry install {{ flag }}
```

neunenak:
With https://github.com/casey/just/pull/2139 this issue can be closed

PR #2139, opened 2024-06-08
is_dependency function

neunenak:
https://github.com/casey/just/issues/1930


casey:
I hate functions named `is_` whatever, but `dependee` seems like a crazy choice that nobody would understand, so `is_dependency` it is!

PR #2169, opened 2024-06-18
Remove dependency on cradle

nc7s:
...as an effort of minimizing dependencies in Debian. Since the tests are meant to pass, I just `unwrap()`'d it.

`JustRun` is just an improvised name, as well as its placement. Do advice if you just don't like that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -307,15 +307,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cradle"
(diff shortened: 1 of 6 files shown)


casey:
Good call!

I think it's fine to just call `Command` instead of adding a trait, since it's only three places, it's not really a huge amount of extra code. I pushed a diff to the PR which removes `JustRun` and uses `Command` directly.

PR #2170, opened 2024-06-19
Ignore errors from env_logger init.

EnigmaCurry:
I want to use `just` as a rust library from my existing rust application, which already uses env_logger:

```
use env_logger::Env;

fn main() {
    env_logger::init_from_env(
        Env::default()
            .filter_or("LOG_LEVEL", "warn")
            .write_style_or("LOG_STYLE", "always"),
    );
    set_env_default("DATABASE_URL");

    // This fails with env_logger double initialization error:
    if let Err(code) = just::run() {
        std::process::exit(code);
    }
}
```

If I run `just::run()` after I have already initialized `env_logger`, I get this error:

```
Builder::init should not be called after logger initialized: SetLoggerError(())
```

It looks like the correct fix is to use env_logger's [try_init](https://docs.rs/env_logger/latest/env_logger/fn.try_init.html#) instead, and so that's what I've done, which fixes the above code to have no more errors.

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -11,7 +11,8 @@ pub fn run() -> Result<(), i32> {
       .filter("JUST_LOG")
       .write_style("JUST_LOG_STYLE"),
   )
-  .init();
+  .try_init()
+  .ok();
 
   let app = Config::app();
 


casey:
Seems reasonable to me!

casey:
Also, although `just` isn't really designed to be used as a library, I'm not opposed to doing simple things to make using it as a library easier, like making `run()` take command line arguments, so they can be passed from a source other than the binary's command line arguments.

PR #2172, opened 2024-06-19
Adds `just::run_with_args` to provide arguments from Rust code

EnigmaCurry:
This adds a new method to run just commands passed from a Vec<&str>.

I found [this blog post](https://www.kirillvasiltsov.com/writing/optional-arguments-in-rust/) where I copied this pattern for optional arguments, I didn't want to mess with the original arity of the run() method, so I made a second method for passing args, and a third to consolidate the common logic, but maybe there's a better way?

This example runs the `foo` target:

```
cargo run --example run-with-args
 ## (stdout): bar
```

I don't quite know how to test this yet either, because I'm not forking the process, I don't think I can capture stdout like the other tests do.

### DIFF
--- a/examples/run-with-args.rs
+++ b/examples/run-with-args.rs
@@ -1,0 +1,8 @@
+// Example showing how to run just commands from Rust code:
+
+fn main() {
+  let args = vec!["-f", "kitchen-sink.just", "foo"];
+  if let Err(code) = just::run_with_args(args) {
+    std::process::exit(code);
+  }
+}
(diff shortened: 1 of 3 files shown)


casey:
Thanks for the PR!

I think it's fine to just have a single `run()` which takes arguments, and pass in `std::env::args_os()` from main. Also, this is small enough that I don't think it needs to be tested. I merged #2173 which does this.

A couple things to keep in mind when using just as a library:

- Since `just` is primarily a CLI tool, the library interface doesn't have any kind of semantic versioning guarantees, and may wind up breaking or changing.

- Also, `run()` will not return if the command line arguments can't be parsed, since currently we call `get_matches_from()` which will exit on error. This could be changed without too much work, and make `run()` return an error instead of exiting when the args can't be parsed.

I added docs to this effect in #2174.

EnigmaCurry:
Thanks!

Issue #2194, opened 2024-06-26
Suggestion: Doc comments for modules


Spatenheinz:
With the Justfiles (in v. 1.29.1):
```
# Justfile
# recipes for foo'ing
mod foo

# foo.just
bar:
  echo "foobar"
```

calling `just --list` gives:

```
Available recipes:
    foo ...
```

It would be great if modules has doc comments similar to recipes.

```
Available recipes:
    foo ...  # recipes for foo'ing
```

casey:
I think this would be great. The relevant code is [here](https://github.com/casey/just/blob/570d3058cfc8381e03d79e93cf857c510a494f09/src/parser.rs#L301). There's already a function for getting the last doc comment, which is used to get recipe doc comments. All that's needed is to add a comment field to `Item::Module`, populate it by calling `pop_doc_comment` when parsing, and it through to the parsed `Justfile`, and display it in `--list`.

Spatenheinz:
I will get to it then :D

laniakea64:
Comment style doc for modules works for me, but `[doc]` attribute style doesn't work? :confused: 

```
[doc("Foo.")]
mod j
```

```
$ just --unstable --list
error: Expected '*', ':', '$', '=', identifier, or '+', but found end of line
 ——▶ justfile:2:6
  │
2 │ mod j
  │      ^
```

casey:
The `[doc(…)]` syntax hasn't been implemented yet, I opened #2207 to track it.

PR #2199, opened 2024-06-27
Add doc for modules for `--list`

Spatenheinz:
PR for #2194.

### DIFF
--- a/README.md
+++ b/README.md
@@ -604,8 +604,8 @@ Examples
 --------
 
 A variety of example `justfile`s can be found in the
-[examples directory](https://github.com/casey/just/tree/master/examples).
-
+[examples directory](https://github.com/casey/just/tree/master/examples) and on
+[GitHub](https://github.com/search?q=path%3A**%2Fjustfile&type=code).
 Features
 --------
 
--- a/README.md
+++ b/README.md
@@ -603,9 +603,10 @@ testing… all tests passed!
 Examples
 --------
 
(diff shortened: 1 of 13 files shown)


casey (changes requested):
Nice! Two things:

- This should have a test which shows that doc comments are printed, and that they're aligned.
- Since `Subcommand::format_doc` is only used inside of `Subcommand::list_module`, I would put it inside that function.

Spatenheinz:
Ok I am a bit unsure why it shows other commits than mine as well. Maybe it is because I have rebased my fork in the meanwhile?

I think it should follow your requirements.
I also changed `/bin/echo` to `/usr/bin/env echo` in `tests/functions` as some of them where failing on for me on NixOS. using `/usr/bin/env` should be more universal. I just replaced all occurences of it for consistency. Let me know if you disagree.

Spatenheinz:
Just noticed! I overwrote some of my changes while trying to fix the git repo.
Will have to reapply them.

casey (approved):
Looks good to me! I made some minor changes, mainly alphabetizing argument lists (when they get long enough, I just sort the argument by name instead of having some semantic order) and un-hiding the doc from the JSON serialization, and fixing the JSON tests.

The extra commits are expected, if you rebase onto a new version of master, you'll see those extra commits, but I ultimately always squash merge, so they get squashed away when they get merged to master.

PR #2211, opened 2024-06-30
Use clap ValueParser for command-line configuration parsing

neunenak:
Some of the configuration currently done with explicit function calls could be done more concisely with clap ValueParser


casey:
LGTM!

casey:
I also converted CommandColor into a ValueEnum that can then be converted to an ansi_term::Color, since that seems less error prone.

PR #2214, opened 2024-07-01
Make `install` an optional dependency and add 32-bit arm targets

### LINKED ISSUE #1284: Install script does not work on Raspberry Pi Zero
Running the script (the first time) returns the following error:
```sh
pi@raspberrypi:~ $ curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to DEST
install: error: Could not determine target from output of `uname -m`-`uname -s`, please use `--target`: armv6l-Linux
```

Running it again or with sudo won't return anything:
```sh
pi@raspberrypi:~ $ curl --proto '=https' --tlsv1.2 -sS^Chttps://just.systems/install.sh | bash -s -- --to DEST
pi@raspberrypi:~ $ sudo curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to DEST
pi@raspberrypi:~ $ 
```

OS/Arch:
```
pi@raspberrypi:~ $ uname -a
Linux raspberrypi 5.10.63+ #1459 Wed Oct 6 16:40:27 BST 2021 armv6l GNU/Linux
```

CramBL:
Resolves #1284 

Believe it or not, some Linux embedded device will not have `install` installed. This is not uncommon when they have a custom distribution build with Buildroot or Yocto, such devices typically have `cp` and `chmod` though, so when `install` is not available, there's now a fall back to cp and chmod. 

Then I added 32-bit arm targets, that includes old Raspberry Pi models but also various embedded linux devices such as Xilinx Zynq family.

### DIFF
--- a/www/install.sh
+++ b/www/install.sh
@@ -142,6 +142,8 @@ if [ -z "${target-}" ]; then
 
   case $uname_target in
     aarch64-Linux)     target=aarch64-unknown-linux-musl;;
+    armv7l-Linux)      target=armv7-unknown-linux-musleabihf;;
+    armv6l-Linux)      target=arm-unknown-linux-musleabihf;;
     arm64-Darwin)      target=aarch64-apple-darwin;;
     x86_64-Darwin)     target=x86_64-apple-darwin;;
     x86_64-Linux)      target=x86_64-unknown-linux-musl;;
--- a/www/install.sh
+++ b/www/install.sh
@@ -129,12 +129,12 @@ if [ -z "${target-}" ]; then
   uname_target="$(uname -m)-$kernel"
 
   case $uname_target in
-    aarch64-Linux)     target=aarch64-unknown-linux-musl;;
-    armv7l-Linux)      target=armv7-unknown-linux-musleabihf;;
-    armv6l-Linux)      target=arm-unknown-linux-musleabihf;;
-    arm64-Darwin)      target=aarch64-apple-darwin;;
-    x86_64-Darwin)     target=x86_64-apple-darwin;;
-    x86_64-Linux)      target=x86_64-unknown-linux-musl;;
+    aarch64-Linux) target=aarch64-unknown-linux-musl;;
+    arm64-Darwin) target=aarch64-apple-darwin;;
+    armv6l-Linux) target=arm-unknown-linux-musleabihf;;
+    armv7l-Linux) target=armv7-unknown-linux-musleabihf;;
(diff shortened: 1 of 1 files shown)


CramBL:
It is tested on a Zynq-7000 which has this output for `uname -a`:
```
Linux hostname 5.15.19-xilinx-v2022.1 #1 SMP PREEMPT Wed Aug 24 06:33:16 UTC 2022 armv7l GNU/Linux
```
This device does not have `install`.

CramBL:
Now also tested on Raspberry PI 0 which has this `uname -a` output

```shell
Linux hostname 6.1.21+ #1642 Mon Apr  3 17:19:14 BST 2023 armv6l GNU/Linux
```

www/install.sh
```diff
@@ -51,6 +51,19 @@ need() {
   fi
 }
 
+need_one_of() {
```
casey: This can be reverted.

www/install.sh
```diff
@@ -62,6 +75,17 @@ download() {
   fi
 }
 
+install_just() {
```
casey: This can be reverted.

www/install.sh
```diff
@@ -93,14 +117,10 @@ while test $# -gt 0; do
   shift
 done
 
-command -v curl > /dev/null 2>&1 ||
-  command -v wget > /dev/null 2>&1 ||
-  err "need wget or curl (command not found)"
-
-need install
+need_one_of curl wget
```
casey: Doing `need unzip` or `need tar` in the `$target` case statement seems fine to me, instead of adding a `need_one_of` function.

www/install.sh
```diff
@@ -171,7 +193,7 @@ if [ -e "$dest/just" ] && [ "$force" = false ]; then
   err "\`$dest/just\` already exists"
 else
   mkdir -p "$dest"
-  install -m 755 "$td/just" "$dest"
+  install_just "$td/just" "$dest"
```
casey: I don't think there's any particular advantage to using `install` when it's available, so let's just use `cp` and `chmod` directly.

casey (changes requested):
Sounds good to me. I think we should just always use `cp` and `chmod` on unix, since there's no particularly advantage to using `install`. See comments for more details.

CramBL:
Requested changes implemented

casey (approved):
Nice, LGTM!

PR #2234, opened 2024-07-07
readme: Use nix-shell instead of nix-env

Pandapip1:
`nix-shell` is generally safer than `nix-env`


casey:
I think we actually want nix-env here. The rest of the commands are to install system wide, and, correct me if I'm wrong, nix-shell only opens a new shell with the command, and doesn't persist on reboot.

Issue #2207, opened 2024-06-29
Allow `[doc(…)]` attribute on modules


casey:
Currently, modules can have doc comments, but the `[doc(…)]` attribute syntax doesn't work.

laniakea64:
Thanks @neunenak for #2247 :+1: 

`[doc(...)]` now works, but [`[doc]` to suppress documentation comment](https://just.systems/man/en/chapter_28.html) has no effect? -
```
# not a doc comment
[doc]
mod foo
```
```
$ just --list
Available recipes:
    foo ... # not a doc comment
```

casey:
Good catch! Fixed in #2254.

PR #2247, opened 2024-07-14
doc annotation on modules

neunenak:
https://github.com/casey/just/issues/2207

### DIFF
--- a/src/compile_error.rs
+++ b/src/compile_error.rs
@@ -32,13 +32,15 @@ impl Display for CompileError<'_> {
     use CompileErrorKind::*;
 
     match &*self.kind {
-      AliasInvalidAttribute { alias, attribute } => {
-        write!(
-          f,
-          "Alias `{alias}` has invalid attribute `{}`",
-          attribute.name(),
-        )
-      }
+      InvalidAttribute {
+        item_name,
+        item_kind,
+        attribute,
+      } => write!(
+        f,
+        "{item_kind} `{item_name}` has invalid attribute `{}`",
+        attribute.name(),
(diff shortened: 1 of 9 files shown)


src/compile_error.rs
```diff
@@ -32,13 +32,15 @@ impl Display for CompileError<'_> {
...
 
     match &*self.kind {
-      AliasInvalidAttribute { alias, attribute } => {
-        write!(
-          f,
-          "Alias `{alias}` has invalid attribute `{}`",
-          attribute.name(),
-        )
-      }
+      InvalidAttribute {
```
casey: The arms of this match statement should be sorted by variant name.

src/parser.rs
```diff
@@ -430,6 +410,17 @@ impl<'run, 'src> Parser<'run, 'src> {
           Some(Keyword::Alias) if self.next_are(&[Identifier, Identifier, ColonEquals]) => {
             items.push(Item::Alias(self.parse_alias(attributes)?));
           }
+          Some(Keyword::Mod)
+            if self.next_are(&[Identifier, Identifier, Comment])
```
casey: This condition is so heinous that we should deduplicate it. Maybe `Parser::next_is_module`?

casey (changes requested):
Mostly looks good! See comments.

casey (approved):
Nice, LGTM! I refactored attribute parsing to just parse attributes before the item, then parse the item, and if the item didn't use the attributes, then return a new CompileErrorKind::ExtraneousAttribute error.

Issue #2248, opened 2024-07-14
Parser support for attributes with multiple arguments


drmacdon:
Currently the Grammar for attribute is,
```
attribute     : NAME ( '(' string ')' )?
```
Which is limiting if you would like an attribute to have more than one argument. ie. to support a script attribute (#1479, #2192) it would benefit from having multiple arguments in addition to the command. ie.
```
[script("bash", "-e")]
[script("python", "-W", "all)]
[script("pwsh", "-NonInteractive")]
[script("nu", "--log-level", "debug")]
```

The error returned for too many arguments could also be better handled if support for multiple arguments was present. [confirm.rs:confirm_recipe_with_prompt_too_many_args](https://github.com/casey/just/blob/master/tests/confirm.rs#L133).

casey:
I think this should be added if and when we add an attribute which needs it.

PR #2249, opened 2024-07-14
Added grammer support to allow for multi-argument attributes

drmacdon:
Parser support for attributes with multiple arguments #2248


casey:
This should be done if and when we have an attribute that supports multiple arguments. Otherwise, it just complicates the codebase for no benefit.

Issue #2243, opened 2024-07-13
Submodules should be groupable (or have their own group)


FelixZY:
I've started grouping commands because that makes them easier to find and read. I'm also using submodules to group certain commands specific to e.g. database commands. However, when listing these using

```just
# Show this help text
[group("General")]
help:
    @just --list --unsorted --justfile {{ justfile() }}
```

the submodules are simply appended to the grouped output, making it appear as if they are part of the last declared group:

```
Available recipes:
    [General]
    help                # Show this help text

    [Code Utils]
    format              # Reformat all source files
    lint                # Lint source files

    [Environment]
    chenv new_env       # CHange ENVironment
    genenv              # Generate a .env file based on secrets/{environment}. Set the `ENVIRONMENT` envvar to change environment.
    current_environment # Print the environment of the variables in the current .env file
    app ...             # Application commands
    auth ...            # Authentication module commands
    db ...              # Database management commands
```

At a minimum, I would expect the submodules to be grouped separately:

```
    [Environment]
    chenv new_env       # CHange ENVironment
    genenv              # Generate a .env file based on secrets/{environment}. Set the `ENVIRONMENT` envvar to change environment.
    current_environment # Print the environment of the variables in the current .env file

    app ...             # Application commands
    auth ...            # Authentication module commands
    db ...              # Database management commands
```

At best, I would like to be able to set custom groups for my submodules:

```just
# Application commands
[group("General")]
mod app "src/app/.app.justfile"

# Authentication module commands
[group("Subcommands")]
mod auth "src/auth/.auth.justfile"

# Database management commands
[group("Subcommands")]
mod db "src/db/.db.justfile"
```

```
Available recipes:
    [General]
    help                # Show this help text
    app ...             # Application commands

    [Subcommands]
    auth ...            # Authentication module commands
    db ...              # Database management commands

    [Code Utils]
    format              # Reformat all source files
    lint                # Lint source files

    [Environment]
    chenv new_env       # CHange ENVironment
    genenv              # Generate a .env file based on secrets/{environment}. Set the `ENVIRONMENT` envvar to change environment.
    current_environment # Print the environment of the variables in the current .env file
```

casey:
Thanks for trying out modules!

I think these would be good changes. I added a space before submodules in `--list` when groups are present in #2244.

Putting modules into groups is a larger change, since it involves modifying the parser to accept attributes on modules, but it seems like a good idea, if anyone wants to tackle it.

FelixZY:
[...]

PR #2263, opened 2024-07-19
feat: allow groups on submodules

jmwoliver:
Hello! I've been working on a fix for https://github.com/casey/just/issues/2243.

I added the whole `attributes` object instead of just passing around the group like how the `doc` is. My rationale for this was to make it easier to have access to anything stored in the `attributes` in the future, let me know if that is the wrong path though.

This is my first time contributing to `just`, I'm open to any feedback!

### DIFF
--- a/src/justfile.rs
+++ b/src/justfile.rs
@@ -413,7 +413,6 @@ impl<'src> Justfile<'src> {
 
     for submodule in self.modules.values() {
       for group in submodule.groups() {
-        // submodule.name.unwrap() is safe here because any non-root justfile will have a name set
         groups.push((&[], submodule.name.unwrap().offset, group.to_string()));
       }
     }
--- a/src/justfile.rs
+++ b/src/justfile.rs
@@ -18,6 +18,7 @@ pub(crate) struct Justfile<'src> {
   pub(crate) default: Option<Rc<Recipe<'src>>>,
   #[serde(skip)]
   pub(crate) loaded: Vec<PathBuf>,
+  pub(crate) groups: Vec<String>,
   pub(crate) modules: Table<'src, Justfile<'src>>,
   #[serde(skip)]
   pub(crate) name: Option<Name<'src>>,
@@ -395,14 +396,25 @@ impl<'src> Justfile<'src> {
     recipes
   }
 
+  pub(crate) fn groups(&self) -> &[String] {
+    &self.groups
+  }
+
   pub(crate) fn public_groups(&self, config: &Config) -> Vec<String> {
     let mut groups = Vec::new();
 
     for recipe in self.recipes.values() {
       if recipe.is_public() {
         for group in recipe.groups() {
-          groups.push((&recipe.import_offsets, recipe.name.offset, group));
+          groups.push((recipe.import_offsets.as_slice(), recipe.name.offset, group));
         }
+      }
+    }
+
+    for submodule in self.modules.values() {
+      for group in submodule.groups() {
+        // submodule.name.unwrap() is safe here because any non-root justfile will have a name set
+        groups.push((&[], submodule.name.unwrap().offset, group.to_string()));
       }
     }
 
--- a/tests/modules.rs
+++ b/tests/modules.rs
@@ -772,6 +772,204 @@ fn doc_attribute_on_module() {
 }
 
 #[test]
+fn group_attribute_on_module() {
+  Test::new()
+    .write("foo.just", "")
+    .write("bar.just", "")
+    .write("zee.just", "")
+    .justfile(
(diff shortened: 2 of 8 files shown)


casey (changes requested):
Nice! I think I would avoid putting storing all attributes on a `Justfile`, for a couple reasons. One abstract reason is that I try not to make changes in anticipation future needs. There's a good chance those future needs may manifest, or they might manifest differently than expected. And one concrete reasons is that attributes serialize in a somewhat odd way, so they will be difficult to parse using `jq`, if you use `--dump --dump-format json` to dump a justfile, whereas a `groups: Vec<String>` field will be easy to parse, so I think I would separate out groups and just store the group names on `Justfile`.

jmwoliver:
@casey Thanks for the feedback! Great points, I'll get this updated to just pass in the group names instead.

jmwoliver:
@casey I switched to passing around `groups: Vec<String>`, hopefully this looks better!

jmwoliver:
Fixed linter errors and re-pushed.

src/justfile.rs
```diff
@@ -395,6 +396,10 @@ impl<'src> Justfile<'src> {
     recipes
   }
 
+  pub(crate) fn groups(&self) -> Vec<String> {
```
casey: This can return `&[String]`, so that we can avoid a clone of the whole vec.

src/justfile.rs
```diff
@@ -406,6 +411,13 @@ impl<'src> Justfile<'src> {
       }
     }
 
+    let empty_vec: Vec<usize> = Vec::new();
+    for submodule in self.modules.values() {
+      for group in submodule.groups() {
+        groups.push((&empty_vec, 0, group));
```
casey: This can be:
```suggestion
        groups.push((&[], 0, group));
```

casey: With this line:

```rust
groups.push((&recipe.import_offsets, recipe.name.offset, group));
```

Changed to:

```rust
groups.push((recipe.import_offsets.as_slice(), recipe.name.offset, group));
```

casey: The contents of `group` is used for sorting, so to make submodule groups sort correctly, think instead of `0`, use `submodule.name.unwrap().offset`. `unwrap()` is okay here, because any non-root justfile will have `name` set.

tests/modules.rs
```diff
@@ -771,6 +771,88 @@ fn doc_attribute_on_module() {
     .run();
 }
 
+#[test]
+fn group_attribute_on_module() {
```
casey, line 775: We should have a test for the ordering of submodules in groups if `--unsorted` is used.

My first thought was that submodules should be ordered based on their position in the `justfile`. However, I'm now not sure if that's correct. I think it's natural to have submodules all declared together at the top or bottom of a `justfile`, and natural to list submodules at the end of each group.

So maybe my comment above is off, and we should, instead of using `submodule.name.unwrap().offset` in the ordering key, use `usize::MAX` instead.

casey (changes requested):
Nice, left some initial comments! To check if the PR will pass CI, run `just ci`, which runs clippy and errors on warnings.

tests/modules.rs
jmwoliver: Hm, I tested setting `submodule.name.unwrap().offset` vs. `usize::MAX`, and both have the same behavior. Both implementations output the submodules at the end of a group because of how we output recipes and then submodules in `subcommand.rs`:

```rust
for (i, group) in ordered.into_iter().enumerate() {
  for recipe in recipes {...}
  for (i, submodule) in submodules.iter().enumerate() {...}
}
```

If we wanted the ability to order based on the submodule's position in the `justfile`, then we would need to interleave the recipes and submodules. But since it sounds like we want them at the end of a group, I'll go with `usize::MAX`. Let me know if you disagree.

src/justfile.rs
jmwoliver: Updated.

src/justfile.rs
jmwoliver: Updated.

tests/modules.rs
jmwoliver: I added tests for `--unsorted` with and without `--list-submodules`. I ended up going with `submodule.name.unwrap().offset` because I think that better guarantees the submodules are ordered how they appear in the `justfile` at the bottom of a group.

jmwoliver:
@casey Thanks for the all the great feedback and the `just ci` tip. I've updated everything with your recommendation, let me know how it looks!

casey (approved):
Nice, looks good to me!

Issue #2024, opened 2024-05-07
nix multi-line shebang is not handled by just


oux:
Environment:
* nix 2.20.1 or later
* just 1.25.2

justfile:
```
hello:
  #!/usr/bin/env nix
  #! nix shell  nixpkgs#hello nixpkgs#cowsay --command bash
  hello | cowsay
```
Error:
![image](https://github.com/casey/just/assets/237991/a6d59947-9382-485b-a6f6-a642b708e59e)

Same shebang on a simple shell script file:
![image](https://github.com/casey/just/assets/237991/09b6f704-9a43-41db-acc5-8777937c472d)

Nix shebang documentation: 
   https://nixos.org/manual/nix/stable/command-ref/new-cli/nix.html#shebang-interpreter

casey:
I think the reason this doesn't work is that when just extracts the body of a shebang recipe into a file to execute, `just` places the shabang line first, and then inserts blank lines before inserting the contents of the recipe, so that line numbers in error messages correspond to line numbers of the recipe.

So the way to fix this is to see if the next line is a shebang and put it at the top of the file.

So for example, this:

```just
hello:
  #!/usr/bin/env nix
  #! nix shell  nixpkgs#hello nixpkgs#cowsay --command bash
  hello | cowsay
```

Is currently translated to this:

```sh
#!/usr/bin/env nix

#! nix shell  nixpkgs#hello nixpkgs#cowsay --command bash
hello | cowsay
```

When it should be translated to this:

```sh
#!/usr/bin/env nix
#! nix shell  nixpkgs#hello nixpkgs#cowsay --command bash

hello | cowsay
```

casey:
Fixed in #2276! All shebang lines are now kept together in the top of the resulting script.

Enzime:
Tested this, #2276 fixes this for me 🎉

casey:
Dope!

PR #2073, opened 2024-05-22
nix shebang multiline functionality

chrass:
Nix shebang multiline  #2024 
runs a map on a vector to collect lines that start with #!

PR #2276, opened 2024-07-26
Allow multiline shebangs in scripts

vkstrm:
Fix for issue #2024, allowing multiline shebangs, most importantly for use with Nix scripts, I don't know if they are relevant anywhere else.

Manual testing with Nix works for me, and I think error line numbers are retained correctly. The test added is a bit fake (just reads it like a shebang and shell like a comment), but a realistic integration test would require adding Nix to the dependencies which maybe isn't a good idea.

First time contributing, thank you for any feedback!


casey (changes requested):
Thanks for the PR! It looks like this code will trigger no matter where in the recipe body the shebang line is. Instead, it should only trigger for the first contiguous shebang lines, and then treat shebang lines after the first non-shebang line normally.

vkstrm:
f20f493f53c441c3e6b7de90cc22a603a50aa95b A suggestion how to do that. Looks a bit more messy, but won't look at shebangs after the ones following the first line. The middle iteration is annoying since it still needs to handle a non-shebang line so that it is added correctly. The last one is cleaner though.

laniakea64:
> Looks a bit more messy, 

FWIW there is also https://github.com/casey/just/pull/2073 which started to attempt a different approach to implement multi-line shebang functionality.  If avoiding messy-looking code is important, might that other PR suggest some way to improve this?

vkstrm:
> > Looks a bit more messy,
> 
> FWIW there is also #2073 which started to attempt a different approach to implement multi-line shebang functionality. If avoiding messy-looking code is important, might that other PR suggest some way to improve this?

I did look at that first but gave it a go myself since it seems it wasn't being worked on. I guess it depends on how you look at it if that solutions is more appropriate, if the Shebang struct is interested in more than just the actual real shebang? I can see how collecting them there and then using them in script() could be done. If you think that is better I can try.

laniakea64:
I'm not `just` dev so not sure what's better.  If I were making this patch less messy-looking I'd make the iterator `.peekable()` and change your middle iteration to something like
```rust
        // Any shebang line that follows the first should also be at the top
        while let Some((line, evaluated)) = iter.by_ref().peek() {
          if !line.is_shebang() {
            break;
          }

          script.push_str(evaluated);
          script.push('\n');
          n += 1;
          iter.next();
        }
```
This avoids duplicating the line-number-matching code and is more concise.

Tried this locally and it passes `cargo test` here.

vkstrm:
Thank you for the suggestion! I tried it with peekable, it helped a bit.
Also moved everything out of the Command arm because it didn't do anything different anymore.
2112a3e0e00f1a0c99fb37f8538695cef795ace7

casey (approved):
LGTM! I simplified it slightly to count the number of shebang lines ahead of time, to avoid needing a peekable iterator, and then just skipping however many shebang lines we processed later.

PR #2280, opened 2024-07-27
grammar: complete missing elements and small fixes

poliorcetics:
Small fixes I saw

Also, why are settings specified by name but not builtin functions and attributes ?


casey:
Thanks for the PR! Settings actually are part of the grammar, since the parser handles parsing setting values depending on which setting it is. Functions and attributes are all parsed the same, independent of which setting or attribute it is, so they're not part of the grammar.

Issue #2273, opened 2024-07-24
Set default working dir for the whole justfile


nyurik:
Frequently I need to place justfile in a different location from where the default working dir should be. Would it be possible to have some global `set` command to modify the working directory, either for the current justfile, or possibly for all of the included ones as well (unless they override it?). The multiple justfiles could in theory be solved with modules, so this is a lower priority.

casey:
I think this would be pretty useful, and not super complicated. Currently, settings are not shared between modules, so adding some kind of setting sharing would be something for another PR. `set working-directory := …` would be pretty easy to add. Marking as good first issue in case someone wants to take a crack at it.

nyurik:
I hacked together an implementation  in #2283, but it is still missing the actual workdir setting. May need some guidance here.

casey:
Implemented in #2283.

PR #2283, opened 2024-07-29
Add `workdir` setting

nyurik:
Implement #2273 - an ability to set workdir for the entire justfile


casey:
Nice! Check out `Recipe::working_directory`. If `self.change_directory()` is true, first use the `workdir` setting, and if it isn't present, Use the existing logic.

nyurik:
thx @casey, seems to work, ready for review

casey:
Nice! I did some refactoring, and renamed the setting `working-directory` for clarity, since i try to avoid too many abbreviations. We could eventually also allow `working-dir` as an alias.

This also made me realize that there's a bug! backticks do not execute in the correct directory if they're in a submodule. I'll fix that in another PR.

nyurik:
thx!  Are you sure you want a long `working-dir` when you already have a short `tempdir` ?

casey:
I think so, just because I commonly hear `tempdir`, but people usually say `working directory` or `working dir`, and `workdir` feels a little less common.

nyurik:
:shrug: your call. I searched github for code matches:
* [workdir](https://github.com/search?q=%22workdir%22&type=code) - 3.5M
* [working directory](https://github.com/search?q=%22working+directory%22&type=code) - 2.7M

P.S. `working directory` mostly seem to show up in documentation, but `workdir` tend to be a lot in the actual code

PR #2288, opened 2024-08-01
add working-directory docs

nyurik:


casey:
Good call on adding some docs. Can you change it so that the existing example isn't changed, and add a new example with `set working-directory`?

casey (changes requested):
See comment.

casey (approved):
LGTM

Issue #2195, opened 2024-06-26
Suggestion: Listing of groups


Spatenheinz:
In v.1.29.1, I can do something like `just --list <submodule_name>` and see the recipes for a given submodule.
It would be great if I could do something along the lines of `just --list <group>` to see all the available recipes for a certain group.
The actual syntax should indeed be considered as simply stating a group could name-clash with submodule names.

I am unsure if the interface is stable, but I personally enjoy a git similar approch with `just --group <group_name>`, but to unify a common interface then I guess also submodules should be `just --submodule <submodule_name>`.

casey:
`just --group <group_name>` is an option, although I think it would be unfortunate if we added another subcommand for listing recipes, and didn't reuse `--list`, since they would both be such similar commands. Some way of disambiguating would likely be awkward though, for example a prefix: `just --list group:NAME` / `just --list @NAME`.

I think it would probably be fine to just use `just --list`, and if there is a conflict between a group and a submodule, display one or the other, probably defaulting to the submodule. If users don't want that behavior, then it's easy enough to rename either the submodule or the group so that it's unambiguous.

wiktor-k:
Excellent idea! I'd like to mark tasks which need to be run in CI via `[group('ci')]` and it'd be really useful to programmatically get all tasks that belong to a given group.

Edit: found a workaround: `just --dump --dump-format=json` and then manually checking for groups. (just in case anyone finds this comment via search).

dougthor42:
+1 to this request.

For our use case, we have a bunch of commands and want to only show the "main" group by default.

> found a workaround: ` --dump --dump-format=json`

And to really make it nifty, some `jq` magic will print things out:

```makefile
_default:
    @just --dump --dump-format=json | jq '.recipes | to_entries[] | select(.value.attributes | any(.group == "main")) | .key'

[group("main")]
ci:

[group("main")]
[group("python")]
format-python:

[group("python")]
test-python:
```

```console
$ just
"ci"
"format-python"
```

I admit, I used gemini to write that `jq` query, so who knows if it's actually any good 🤣

dougthor42:
Here's a slightly better workaround implementation. I'm still working on the blue comments haha - seems `column` doesn't like terminal color codes.

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/59c3f444-9d88-4f30-b5b4-9cd2a7d7b69f" />

```justfile
_default:
    @scripts/_just_list.sh
```

```bash
#!/bin/bash
[...]

PR #2298, opened 2024-08-02
Allow listing of groups (also recursively) 

Spatenheinz:
This commit is in reference to: https://github.com/casey/just/issues/2195

It took me some time to implement this feature, with getting to understand the layout, trying to parameterize it to remove code duplication and going on vacation. 

With this commit it should be possible to list all occurences of a group. If a submodule and a group has the same name the submodule takes precedence.

It was a nightmare to rebase this. I hope it is working as intended still (at least the tests should pass :D)

Let me know if you like it or not, and if you would like anything changed.

### DIFF
--- a/src/analyzer.rs
+++ b/src/analyzer.rs
@@ -227,7 +227,6 @@ impl<'src> Analyzer<'src> {
       loaded: loaded.into(),
       modules,
       name,
-      doc,
       recipes,
       settings,
       source: root.into(),
--- a/src/analyzer.rs
+++ b/src/analyzer.rs
@@ -227,6 +227,7 @@ impl<'src> Analyzer<'src> {
       loaded: loaded.into(),
       modules,
       name,
+      doc,
       recipes,
       settings,
       source: root.into(),
--- a/src/error.rs
+++ b/src/error.rs
@@ -172,6 +172,9 @@ pub(crate) enum Error<'src> {
   UnknownSubmodule {
     path: String,
   },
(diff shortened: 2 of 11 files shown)


casey (changes requested):
I don't think it's a good idea to overload `--list NAME` to mean either "list submodule with `NAME`" or "list group with NAME". Submodules and groups may have the same name, so this is ambiguous.

This also looks, I think, more complicated than it needs to be.

My suggestion is to add a `--group` flag which takes the name of a group. If `--group is passed, then the results of list should be filtered to only show recipes which match the given group name. I think the existing list code can be largely unchange, aside from the addition of a few:

```rust
if let Some(group) = &config.group {
  if !recipe.groups().contains(group) {
    continue;
  }
}
```

Spatenheinz:
I can look into making it less confusing (although most of what i have done is to not repeat too much code between listing of a group and of a submodule), but looking at it again it can probably be done in a better way.

Regarding the `--group`, either I have completely misunderstood you or you shifted your opinion since I opened https://github.com/casey/just/issues/2195 :

https://github.com/casey/just/issues/2195#issuecomment-2192821668

casey:
Yah I think looking at the implementation, and how much simpler `just --list --group FOO` would be made me change me ind.

casey:
Sorry to flip flop!

Spatenheinz:
its all good. I will try and see to it.

casey:
It is now possible to do `just --list --group foo`.

PR #2300, opened 2024-08-04
Implement the private attribute for assignments

### LINKED ISSUE #2299: Allow defining private variables
Often, I need to define variables that are built from a smaller set of fundamental settings/variables. It would be useful to allow `[private]` to work on variables such that they can't be defined by the user of the script. Alternatively, the `_` prefix could also be used such that `just --variables` avoids listing them.

Happy to work on a PR if this is something that will be accepted

adsnaider:
This PR fixes https://github.com/casey/just/issues/2299. When a variable is specified as `[private]` or is named with `_*` the variable will be hidden from `just --variables` and forbidden from being overridden. I purposefully didn't change the behavior of the evaluator on private variables since it's likely useful to be able to view the values of private variables for debugging & introspection.

### DIFF
--- a/src/analyzer.rs
+++ b/src/analyzer.rs
@@ -159,11 +159,9 @@ impl<'src> Analyzer<'src> {
         return Err(assignment.name.token.error(DuplicateVariable { variable }));
       }
 
-      if self
-        .assignments
-        .get(variable)
-        .map_or(true, |original| assignment.depth <= original.depth)
-      {
+      if self.assignments.get(variable).map_or(true, |original| {
+        assignment.file_depth <= original.file_depth
+      }) {
         self.assignments.insert(assignment.clone());
       }
 
--- a/src/binding.rs
+++ b/src/binding.rs
@@ -9,8 +9,16 @@ pub(crate) struct Binding<'src, V = String> {
   pub(crate) export: bool,
   /// Binding name
   pub(crate) name: Name<'src>,
+  /// Whether this binding is private to the script
+  pub(crate) private: bool,
   /// Binding value
   pub(crate) value: V,
+}
+
+impl<V> Binding<'_, V> {
+  pub fn is_public(&self) -> bool {
+    !self.private
+  }
 }
 
 impl<'src, V> Keyed<'src> for Binding<'src, V> {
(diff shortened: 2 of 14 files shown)


casey (changes requested):
The `[private]` attribute on recipes doesn't prevent running private them, only listing them, and I'm inclined to think the `[private]` variable attribute should similar, i.e., only prevent listing them, but still allow overriding them.

casey:
I would also filter them out of `--evaluate`. We can think about some kind of `--all` flag in a follow-up PR, which un-hides them. (`--all` should probably also allow listing all recipes, including private recipes)

adsnaider:
I believe this is ready again. Haven't implemented the `--all` flag but the other changes should now be working properly and as you point out that flag can be added later

casey (changes requested):
Sorry I missed this the first time! I would remove the `allow-private-variables` setting, since I think an `--all` flag, which can be added later, is more useful and general.

adsnaider:
Ah sure, I assumed that would be a backwards-incompatible change

laniakea64:
> Ah sure, I assumed that would be a backwards-incompatible change

Using the `[private]` attribute on variables is backwards-compatible, since it's an error to do that in current `just`, so it won't change behavior of any working justfiles.  So no need for a setting for that.

But wouldn't making all underscore-prefixed variables private be backwards-incompatible, and therefore _that_ aspect _would_ need an opt-in setting?

adsnaider:
@laniakea64 Yeah, that's what I thought. I could take out the `_` prefix for private variables :shrug:

casey:
It could be argued either way, but I think making variables private if they start with `_` is probably fine. It won't actually break any justfiles, just remove those variables from the output of `just --evaluate` and `just --variables`, which are, I think, mostly for human consumption.

adsnaider:
> Sorry I missed this the first time! I would remove the `allow-private-variables` setting, since I think an `--all` flag, which can be added later, is more useful and general.

Removed the setting @casey

casey (approved):
LGTM!

casey:
Available in 1.35.0, just released: https://github.com/casey/just/releases/tag/1.35.0

PR #2312, opened 2024-08-13
Add Python package wrapper (pyproject.toml) and wheel action

### LINKED ISSUE #2311: [feature request] Installation via pip?
This is a bit of an odd request, but I work in some environments where installing external dependences is difficult (e.g. it would complicate the dev setup process) except that installation of Python packages is considered acceptable.

Would you consider a PR that (a) creates a Python package that just bundles `just` (via maturin) as a CLI script, and (b) a Github actions workflow that cuts a new PyPI release whenever `just` is updated?

That way, you can install `just` via `pip install just` 

(though in practice we might have to choose a different name, `just` is unfortunately taken on PyPI)

No worries if out of scope - I would just maintain my own such wrapper if that's the case but I thought it would appropriate to ask.

rachtsingh:
Fixes https://github.com/casey/just/issues/2311

Should hopefully not be a big maintenance burden, but will make it easier for me to recommend `just` as the Python make replacement.

To do (by maintainer):
- decide on package name (I picked `justcli` since `just` was taken on PyPI, but no preference)
- create a PyPI account and set the `PYPI_API_TOKEN` in Github to allow uploads

While looking at PyPI for a good name I found this package that wraps `just`: [pybin](https://github.com/justin-yan/pybin), but it looks like it is releasing a (slightly) older version of just, i.e. 1.26.0.


casey:
It's a bit convoluted, but in `workflows/ci/release`, there's a test for whether or not the tag matches `[[:digit:]]+[.][[:digit:]]+[.][[:digit:]]`, i.e., whether or not it's a version tag `x.y.z` or some other tag. If it's not a version tag, the release job runs, but it doesn't publish the built artifacts. Can you add this to the maturin job? Otherwise, there's no way to run the job, for debugging, without also attempting to upload those wheels. (Unless there's some other way to do this with the existing workflow.)

dalito:
This is interesting. It would also make just installable via pipx, or not? (and pipx is by default available in gh-actions)

casey:
I'm not sure if it would allow you to install them directly. You would need the wheels, which would only exist after an actions run, so they would be hard to grab.

casey:
See comments in https://github.com/casey/just/pull/2361, but I'm leaning towards not adding this. I fear it will get out of date or the build will break, and I'm honestly pretty unlikely to test it, so I would suggest maintaining this out-of-tree, and I can link to it in the readme. For the package name, I would suggest rust-just, but justcli is fine too.

PR #2313, opened 2024-08-20
Separates cross-platform managers into separate table

cskeeters:
As suggested in #1971.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,45 @@ most Windows users.)
 
 ### Packages
 
+Cross-platform package managers:
+
+<table>
+  <thead>
+    <tr>
+      <th>Package Manager</th>
+      <th>Package</th>
+      <th>Command</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td><a href=https://asdf-vm.com>asdf</a></td>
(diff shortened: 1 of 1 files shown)


casey:
Looking at this, I think maybe it's better if we just put all the entries for `Various` at the top of the table, instead of splitting it into two , and maybe instead of `Various` call it `Cross-platform`. What do you think?

cskeeters:
That's what I did.

casey:
My suggestion is to keep a single table, but to put the cross-platform options at the top of that single table.

cskeeters:
Oh I see.  Personally, I like the sparate table.  It's like two Hashmaps that have different keys.  It tells the user you have to check both maps; one for the package manager and one for the OS.  If you got rid of the OS column and ordered everything by package manager, that would be one thing, but then you'd loose some of the information you have about OS support.  That said, I think you should do what makes sense to you.

casey:
Just based on personal preference, I think I like a single unified table best, mostly because I don't like too deeply nested headings.

casey:
I wound up doing this in #2385, and I think you're right, multiples tables are probably best.

PR #2318, opened 2024-08-25
Dump individual recipes by name

### LINKED ISSUE #737: Dump individual recipe contents
Opening regarding @vmiheer's comment in #701:

> On other hand, is there way to keep the generated shebang script for later inspection? I see a file created in /tmp/justJOLujt/maestro but it gets deleted when the just execution stops. Maybe printing the path at the beginning so that one can put sleep 100 to inspect the content would be useful.

This is definitely possible, but an easier way of doing it might just be to add the ability to just print the generated shebang script to stdout. Would that work?

neunenak:
This commit modifies the `--dump` option to optionally take the name of a recipe as an argument, and only dump the contents of that single recipe. Resolves https://github.com/casey/just/issues/737

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -10,6 +10,15 @@ pub(crate) struct Ast<'src> {
   pub(crate) working_directory: PathBuf,
 }
 
+impl<'src> Ast<'src> {
+  pub(crate) fn get_recipe_item(&self, name: &str) -> Option<&Item<'src>> {
+    self
+      .items
+      .iter()
+      .find(|item| matches!(item, Item::Recipe(recipe) if recipe.name.lexeme() == name))
+  }
+}
+
 impl<'src> Display for Ast<'src> {
   fn fmt(&self, f: &mut Formatter) -> fmt::Result {
     let mut iter = self.items.iter().peekable();
--- a/src/config.rs
+++ b/src/config.rs
@@ -71,8 +71,7 @@ mod cmd {
     VARIABLES,
   ];
 
-  pub(crate) const ARGLESS: &[&str] =
-    &[CHANGELOG, DUMP, EDIT, FORMAT, INIT, MAN, SUMMARY, VARIABLES];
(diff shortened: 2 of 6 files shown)


casey (changes requested):
Nice! A couple thoughts:

- Like `--show` this should probably take a `ModulePath` so submodule recipes can be dumped, and ideally, it should be possible to dump a submodule if the name resolves to a module and not a recipe.
- This should probably reuse the `UnknownRecipe` error.

neunenak:
> Nice! A couple thoughts:
> 
>     * Like `--show` this should probably take a `ModulePath` so submodule recipes can be dumped, and ideally, it should be possible to dump a submodule if the name resolves to a module and not a recipe.
> 
>     * This should probably reuse the `UnknownRecipe` error.

Hm, is `--dump <recipe>` actually doing anything different than `--show <recipe>`? Should it? I guess there's a difference if you pass in `--dump-format json`, but maybe it's confusing to have two commands that mostly do the same thing.

casey:
Whoops 😅 Yah it looks like show already shows a recipe, and dump shows the entire justfile. Looking at #737, it looks like what's being requested is a way to print the *evaluated* recipe, i.e., including interpolations, so you can see what is going to be executed.

neunenak:
So maybe it makes sense if `--dump` always printed out the pre-evaluation version of a recipe (or the whole justfile); and `--show` printed out a post-evaluation version. This would involve making `--dump <recipe>` do the same thing that `--show <recipe>` currently does, and then changing the behavior of `--show <recipe>` to something new. 

Or to avoid breaking changes there could be another flag (`--display`?) that does the post-evaluation behavior, and then maybe it's fine if `--show <recipe>` really is just an alias for `--dump <recipe>`.

casey:
I think it's probably not worth breaking changes. I would keep `--show` for recipes and `--dump` for justfiles, and add a flag that changes the behavior of `--show`.

Issue #610, opened 2020-03-15
Refactor current config into `Config`, `Options`, and `Subcommand`


casey:
Move the fields, aside from `subcommand`, of `Config` into a struct called `Options`.

Then move all the subcommand execution functions into `subcommand.rs`, with the main entry point being `Subcommand::run` with signature `fn run(self, options: Options)`.

PR #2320, opened 2024-08-26
Remove Config::run

neunenak:
Minor refactor that removes a layer of indirection and one function named `run`. Arguably partially addresses https://github.com/casey/just/issues/610 , although I'm not sure if this issue is still relevant anyway.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -748,11 +748,6 @@ impl Config {
       Err(Error::UnstableFeature { unstable_feature })
     }
   }
-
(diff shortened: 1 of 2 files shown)


casey:
This seems reasonable. I'm not even sure if #610 is worth it, I think maybe it was motivated by the desire to have a struct which contains options, but doesn't contain the subcommand, but I think this is really just aesthetic, so I'm just gonna close it.

PR #2323, opened 2024-08-27
Generate .sha256sum files for release artifacts

twm:
I am *very much* not confident that `sha256sum` is available on the Windows runner images, so unsure that this will work in its current form (hence the draft status).

If `sha256sum` is not available, I could easily do the work in Python, which [I see](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md) is available on the runners.

Another possibility: rather than uploading the artifacts one-by-one to the release (which allows partial failure), I could restructure the build to temporarily [store the artifacts](https://github.com/twisted/twisted/blob/d0db2ee5592bff64f5c8f23a1265180eae022d38/.github/workflows/test.yaml#L314-L319) and then [download them all](https://github.com/twisted/twisted/blob/d0db2ee5592bff64f5c8f23a1265180eae022d38/.github/workflows/test.yaml#L346-L351) before generating the release. That'd happen in a Linux environment where `sha256sum` is available.

Please let me know your preference!

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -99,7 +99,9 @@ jobs:
       if: ${{ startsWith(github.ref, 'refs/tags/') }}
       with:
         draft: false
-        files: ${{ steps.package.outputs.archive }}
+        files: |
+          ${{ steps.package.outputs.archive }}
+          ${{ steps.package.outputs.archive-checksum }}
         prerelease: ${{ steps.ref-type.outputs.value != 'release' }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -101,7 +101,7 @@ jobs:
         draft: false
         files: |
           ${{ steps.package.outputs.archive }}
-          ${{ steps.package.outputs.archive-checksum }}
+          ${{ steps.package.outputs.archive }}.sha256sum
         prerelease: ${{ steps.ref-type.outputs.value != 'release' }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/bin/package
+++ b/bin/package
@@ -45,13 +45,12 @@ case $OS in
     ARCHIVE=$DIST/just-$VERSION-$TARGET.tar.gz
     tar czf $ARCHIVE *
     echo "archive=$ARCHIVE" >> $GITHUB_OUTPUT
+    shasum -a 256 $ARCHIVE > $ARCHIVE.sha256sum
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -99,7 +99,9 @@ jobs:
       if: ${{ startsWith(github.ref, 'refs/tags/') }}
       with:
         draft: false
-        files: ${{ steps.package.outputs.archive }}
+        files: |
+          ${{ steps.package.outputs.archive }}
+          ${{ steps.package.outputs.archive-checksum }}
```
casey: Can you avoid needing to create another output variable by doing this:

```suggestion
          ${{ steps.package.outputs.archive }}.sha256sum
```

casey (changes requested):
Nice! It's probably best not to make the different jobs depend on each other, since it might complicate debugging if something breaks.

On macOS you'll have to use `shasum -a 256`, which I think is also available on Linux, so you can use that on all non-windows machines.

If `shasum -a 256` isn't available on Windows (I have no idea, but it's probably worth a shot, who knows what they have installed on the runners.) you can use `certutil -hashfile FILE SHA256`. (As much as I would like to pretend that I somehow knew this, Claude told me.)

So maybe change it to use `shasum -a 256` on all platforms, and I'll kick off a job by pushing a test branch to GitHub.

Also, see the comment about possibly avoiding needing to create another output variable.

twm:
> On macOS you'll have to use `shasum -a 256`, which I think is also available on Linux, so you can use that on all non-windows machines.

It looks like that's a Perl thing, and Perl is installed on the Windows runners. I'll standardize on it.

twm:
Hopefully this works! :crossed_fingers:

twm:
So `shasum` for Mac and Linux and `sha256sum` for Windows. :sweat_smile:

casey (approved):
I tried `certutil`, however although it calculates SHA2 hashes, it doesn't actually output them as `HASH\tFILE`.

Humorously, I originally saw that sha256sum was failing on macOS, and didn't pay attention to whether it worked on Windows. It actually *is* installed on Windows, via Git Bash. So I switched it back to `shasum -a 256` on Linux and Mac, and `sha256sum` on Windows.

casey:
Just a heads up for anyone using the shasum files: In #2417 I changed the release workflow to generate a single `SHA256SUM` file, instead of a `.shas56sum` file for each artifact. So for upcoming releases, you'll need to download the `SHA256SUM` file, and run `shasum -a 256 -c SHA256SUM --ignore-missing`.

twm:
@casey Thanks for the heads-up! And thanks again for adding this in the first place.

PR #2335, opened 2024-08-31
Slight refactors in subcommand and search modules

neunenak:
* adds some documentation comments for various methods on `Search`
* passes `Compilation` into several `Subcommand` handler methods to reduce cluttering up the outer scope of `Subcommand::execute`
* makes some code slightly more concise in a few areas


casey:
I wasn't a huge fan of the refactor to `init` to use variables to store the `justfile` and `working_directory`, but otherwise looks good, made some tweaks to the docs, and passing in a `Compilation` is definitely an improvement.

PR #2336, opened 2024-09-01
Simplify `Subcommand::run`

neunenak:
The `Subcommand::run` function is a bit messy because it needs to handle the case where fallback is enabled and a missing recipe triggers searching for a new justfile further up the directory tree. This commit attempts to simplify the loop within `Subcommand::run` and make the logic generally easier to understand.


casey:
Won't this lead to double compiling the first justfile, since it's already compiled before we call run?

neunenak:
> Won't this lead to double compiling the first justfile, since it's already compiled before we call run?

Whoops, yeah. Added a commit that fixes this.

casey:
Nice, this is a great simplification!

PR #2351, opened 2024-09-07
Add groups to repo justfile

neunenak:
This commit adds the recipes in the repo's own `justfile` to several recipe groups.

### DIFF
--- a/justfile
+++ b/justfile
@@ -8,54 +8,69 @@ log := "warn"
 
 export JUST_LOG := log
 
+[group: 'dev']
 watch +args='test':
   cargo watch --clear --exec '{{ args }}'
 
+[group: 'test']
 test:
   cargo test --all
 
+[group: 'check']
 ci: forbid test build-book clippy
   cargo fmt --all -- --check
   cargo update --locked --package just
 
+[group: 'check']
 fuzz:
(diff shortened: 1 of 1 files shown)


neunenak:
I thought it would be nice if the just project's own `justfile` exhibited groups and had more documentation of the recipes. I picked some group names that I thought made sense, don't feel very strongly about the exact names or categorization scheme.

casey:
Let's just do groups first, since it's less to nitpick.

What do you think of these groups? I'd like to keep the names short. I also like single quotes, since they don't require hitting the shift key to type. (I also just merged #2367, which nuked a bunch of old recipes, so some of the following are gone.)

```
[group: 'dev']
watch
push
pr
done
install-dev-deps
replace
convert-integration-test

[group: 'test']
test
filter
test-completions
test-bash-completions
test-release-workflow

[group: 'check']
ci
lint
fuzz
shellcheck
check
outdated
clippy
forbid
build-book

[group: 'misc']
run
build
fmt
install
sloc

[group: 'doc']
man
view-man
render-readme
watch-readme

[group: 'release']
update-changelog
update-contributors
publish
readme-version-notes

[group: 'demo']
test-quine
quine
polyglot
pwd
```

neunenak:
yeah those groups seem good to me

Issue #2350, opened 2024-09-06
Bug in redefinition compile error reporting


marcaddeo:
While working on #2344 I noticed that the error reporting for redefinition compile errors was not properly swapping types in the messaging when the redefinition is between two types.

Examples:

#### Aliases

```just
foo:
  echo foo

alias foo := bar

bar:
  echo bar
```

```
error: Alias `foo` defined on line 1 is redefined as a recipe on line 4
 ——▶ just-test:4:7
  │
4 │ alias foo := bar
  │       ^^^
```

#### Modules

```just
foo:
  echo foo

mod foo "foo.just"
```

```
error: Module `foo` defined on line 1 is redefined as a recipe on line 4
 ——▶ just-test:4:5
  │
4 │ mod foo "foo.just"
  │     ^^^
```

Should I pull the relevant code changes from #2344 and put a new PR in addressing the bug?

casey:
Thanks for the report!

> Should I pull the relevant code changes from https://github.com/casey/just/pull/2344 and put a new PR in addressing the bug?

Yes, that would be awesome!

casey:
Fixed by #2353.

PR #2353, opened 2024-09-07
Fix bug in redefinition compile error reporting

marcaddeo:
This fixes a bug that caused the error reporting for redefinition compile errors to be reported "backwards" when the original and redefinition were differing types.

See #2350 for additional information.

### DIFF
--- a/src/analyzer.rs
+++ b/src/analyzer.rs
@@ -383,7 +383,7 @@ mod tests {
     line: 2,
     column: 6,
     width: 3,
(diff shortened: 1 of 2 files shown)


casey:
Nice! Can you add a test which demonstrates the fix?

marcaddeo:
> Nice! Can you add a test which demonstrates the fix?

Test added 👍🏻

casey:
Nice, merged!

PR #2360, opened 2024-09-10
Allow escaping arbitrary characters

### LINKED ISSUE #2359: Allow escaping arbitrary characters in double-quoted strings
Currently `just` strings delimited with `" "` or `""" """` only support escaping a limited set of characters - https://just.systems/man/en/chapter_30.html

Most other languages have a syntax for escaping arbitrary Unicode characters.  Having such syntax in `just` would make some justfiles easier to work with:
 - Not all valid Unicode characters are printable.  And even within the set of printable characters, some text editors don't always do a good job of displaying some printable Unicode characters.  Representing these characters as sequence of ASCII characters is not subject to such problems.
 - If you know the escape sequence of a character you want to use, having escape syntax means you can just type the character rather than needing to find the character in a character map and copy+paste.

Additionally, such syntax would be helpful if the recipe attribute suggested in https://github.com/casey/just/issues/647#issuecomment-2329480131 gets implemented.

Thinking about `just`'s backwards-compatibility guarantee, maybe the best syntax would be copying [Rust's `\u{...}`](https://doc.rust-lang.org/reference/tokens.html#character-escapes), since it allows a flexible and unambiguous number of hex digits?

laniakea64:
Closes https://github.com/casey/just/issues/2359

Not sure the error reporting is ideal, and not sure if the tests are covering enough cases?

### DIFF
--- a/src/compile_error.rs
+++ b/src/compile_error.rs
@@ -181,7 +181,6 @@ impl Display for CompileError<'_> {
         "{item_kind} `{item_name}` has invalid attribute `{}`",
         attribute.name(),
       ),
-      InvalidCharacter { hex } => write!(f, "`{hex}` does not represent a valid character"),
       InvalidEscapeSequence { character } => write!(
         f,
         "`\\{}` is not a valid escape sequence",
@@ -193,10 +192,7 @@ impl Display for CompileError<'_> {
           _ => character.escape_default().collect(),
         }
       ),
-      InvalidHex { hex, error } => write!(f, "`{hex}` is not a valid hexadecimal number: {error}"),
-      InvalidUEscapeSequence { expected, found } => {
-        write!(f, "expected `{expected}` but found `{found}`")
-      }
+      InvalidUEscapeSequence { character } => write!(f, "expected `{{` but found `{character}`"),
       MismatchedClosingDelimiter {
         open,
         open_line,
@@ -252,10 +248,6 @@ impl Display for CompileError<'_> {
         f,
         "Non-default parameter `{parameter}` follows default parameter"
       ),
-      UEscapeSequenceTooLong { hex } => write!(
-        f,
-        "more than 6 hex digits in escape sequence starting with `\\u{{{hex}`"
-      ),
       UndefinedVariable { variable } => write!(f, "Variable `{variable}` not defined"),
       UnexpectedCharacter { expected } => write!(f, "Expected character `{expected}`"),
       UnexpectedClosingDelimiter { close } => {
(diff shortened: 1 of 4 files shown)


casey:
What do you think about introducing a `State` enum, which can be declared inside the parse string function, like so:

```
enum State {
  Initial, // initial state
  Backslash, // after `\`
  Unicode, // after `\u`, but not opening `{`
  UnicodeValue { value: String }, // after `\u{`
}
```

Then you would do:

```
let mut state = State::Initial;
while let Some(c) = chars.next() {
  match state {
  }
}

if state != State::Initial {
  // return error
}
```

I think this might make the code a little simpler.

Also, I think we should see if the error messages from `u32::from_str_radix` are good enough to use without needing to define our own error type. If they're reasonable, then we could accept any characters between `\u{` and `}`, and let `u32::from_str_radix` complain if they're not hex, or if they overflow a u32.

laniakea64:
Done.  It does simplify the code and looks better, thanks for the suggestions!

> we could accept any characters between `\u{` and `}`, and let `u32::from_str_radix` complain if ... they overflow a u32.

The syntax we're implementing only accepts up to 6 hex digits, while `u32` is up to 8 hex digits, so the code needs to keep its own length check.

casey:
Looking at the error message from `std::num::ParseIntError`, I think it's actually not great, so I'm coming around to doing more checking ourselves and not delegating the error message.

I think we should have these errors:

- `UnicodeEscapeEmpty` for `\u{}`
- `UnicodeEscapeCharacter { character: char }` for `\u{X}` where `X` is anything other than `0-9a-fA-F`
- `UnicodeEscapeLength` for when there are more than six characters inside an escape sequence
- `UnicodeEscapeRange` when the u32 is not in range for a char, the error message should mention that `10FFFF` is the maximum valid codepoint
- `UnicodeEscapeUnterminated` when we don't encounter `}` before the end of the string

This should allow us to just `.unwrap` when calling `from_str_radix`, since it will be guaranteed to be a valid hex string that's in range for a u32.

Sorry for waffling!

laniakea64:
> I think we should have these errors:

But when `just` encounters an invalid escape sequence, it points at the entire string, which means the errors need more information than that to pinpoint.  To make the exact suggested errors awesome, could `just` please point only at the malformed escape sequence? :slightly_smiling_face: 

In the mean time, pushed a commit trying to find a middle ground, since digging into why `just` points at the whole string is more than I have capacity for atm :frowning:

casey:
Compilation error message context is based on tokens, so since a string is one token, by default it points at the whole string. This could definitely be improved. The easiest solution would be to construct a new, fake token which only points at the sub-section of the string containing the invalid escape sequence, but that's probably best left to a follow-up PR.

casey:
Merged! This is great. Tweaked the code a little, plus error messages.

laniakea64:
> Tweaked the code a little, plus error messages.

Nice, I like the tweaks, thanks!  Shows a few Rust syntaxes I'm not sufficiently familiar with.

PR #2366, opened 2024-09-13
note that you can override dotenv-path at runtime

willie:
Being able to override `-E/--dotenv-path` at runtime is super useful and should be part of documentation.


casey:
Nice, merged!

willie:
Nice revision, even clearer!

PR #2375, opened 2024-09-18
Remove AliasShadowsRecipe compilation error

neunenak:
`AliasShadowsRecipe` doesn't actually ever get created, because the check for redefinitions that emits the more-general `Redefinition` compiler error catches cases where an alias has the same name as an existing recipe instead. So it is redundant and can be removed.


casey:
Nice, good catch!

Issue #1641, opened 2023-07-11
[Feature] Support globbing with includes directive


th3fallen:
It would be amazing to be able to do `!include path/to/my/subjusts/*.just` and move on with life vs including each seperately.

IrrationalPi:
I would love this feature also. I used it a lot with make to divide makefiles into like functionality.

Here is a workaround. The `update` recipe will scan a folder for all *.just files, format them, create a temporary Justfile with all the !include directives, then dump them to a single Justfile. I have found !include to be too buggy for daily use in the primary Justfile.

Initial Justfile:
`!include ./justfiles/self.just`

./justfiles/self.just:
```
justfile_dir := "./justfiles"

update:
    #!/usr/bin/env bash
    set -euxo pipefail
    cp Justfile Justfile.bak
    echo "# DO NOT EDIT. Auto-generated. Update with 'just update'." > Justfile.includes
    for file in {{ justfile_dir }}/*.just; do
        echo "!include ${file}" >> Justfile.includes
        just --justfile "${file}" --fmt --unstable || true
    done
    just --unstable --justfile Justfile.includes --dump > Justfile
    rm Justfile.includes
```

After making any changes to a *.just file in your watched folder, run:
`just update`

casey:
This seems reasonable. `include` has been stabilized as `import`, so this wouldn't technically be backwards compatible, but I think that `*` in paths is so vanishingly rare that we can go ahead and do it.

casey:
Closing in favor of #1885, it's later but has more discussion.

PR #2376, opened 2024-09-19
Support globbing in imports

neunenak:
Implements shell-style globs in import statements using the Rust `glob` crate. Addresses https://github.com/casey/just/issues/1641


src/compiler.rs
```diff
@@ -75,14 +75,39 @@ impl Compiler {
               .join(Self::expand_tilde(&relative.cooked)?)
               .lexiclean();
 
-            if import.is_file() {
+            let import_base_str = import.to_string_lossy();
```
casey: `just` is expected to work with non-unicode paths, so we can't do a lossy conversation here. I'd suggest checking out the `globset` crate, which I believe handles non-unicode paths.

src/compiler.rs
```diff
@@ -75,14 +75,39 @@ impl Compiler {
...
+                case_sensitive: true,
+                require_literal_separator: false,
+                require_literal_leading_dot: false,
+              };
+              let import_paths = glob::glob_with(&import_base_str, glob_options)
+                .map_err(|error| Error::ImportGlob { error, path: *path })?;
+              for import in import_paths {
+                let Ok(import) = import else { continue };
+
+                if import.is_file() {
```
casey, line 123: I think a glob matching a directory should be an error. Just because we can backwards-compatibly make it not an error, but not backwards-compatibly turn it into an error.

casey (changes requested):
See comments. The main one is that we have to handle non-unicode paths, so we should look at a different `glob` crate.

src/compiler.rs
neunenak: `globset` unfortunately doesn't have a method to iterate over a file system entry and select only paths that match the glob, like the iterator that `glob_with` returns with the `glob` crate. Looking at how `glob_with` is implemented in `glob`, it seems like this code is a bit tricky to write and I'd rather avoid it (or if I did write that code, I think it might work out to just implementing path-globbing functionality from scratch, which is also undesirable). 

There's also this crate: https://crates.io/crates/wax , which is utf-8 only but claims to reasonably handle sane non-utf-8 text internally, and also seems to have the file system iterator I want. It also hasn't been updated in a bit and claims to be experimental, so I'm kind of iffy about this crate.

Maybe I'm missing something obvious that would make this easier to do.

casey: Ahh, that's lame. A concern about `wax`, any syntax that `wax` supports becomes part of the public-facing interface of `just`, and so can't be added and removed without a backwards-incompatible change. `wax` supports some pretty exotic syntax, like repetitions `<a*/:0,>` and  alternatives `{a?c,x?z,foo}`, which I doubt would be incredibly useful for use in import paths.

neunenak: Actually it looks like even globset expects to be given the glob as a string: https://docs.rs/globset/latest/globset/struct.GlobBuilder.html#method.new . Also see this 2019 comment from the maintainer of `globset` on `glob`: https://github.com/rust-lang/glob/issues/78#issuecomment-483733581 and this `globset` issue: https://github.com/BurntSushi/ripgrep/issues/1250 . 

I'm starting to think that there's no principled way to support globbing on non-utf8 filesystems without basically submitting a patch to some upstream glob project to do it, but maybe few enough people are using these tools with non-utf8 globs that no one has previously run into it?

The `to_string_lossy` function turns any non-utf8 sequence into the unicode replacement character �. I wonder if there is some other function that could be called to try to convert a `PathBuf` into a meaningful Rust utf-8 string (maybe doing something smart with guessing the original encoding). The quick solution would be to move this conversion into the `has_glob` block, which would mean that people using glob imports and file paths with non-utf8 names would run into issues but at least no existing just behavior would break.

casey: It's a bit lame, but we could just throw an error if you try to glob with a path which can't be converted to unicode. Non-unicode paths are very rare and basically just crazy, so not supporting them is probably fine, especially if they're only unsupported if you use a pretty niche feature.

I'm tempted to land this feature as unstable initially, both because of the unicode path issue, but also because there's some choice about which glob syntaxes we support, and it might be good to make compare them all and make a principled choice about what to support.

neunenak:
Alright, I gated glob imports behind unstable, went back to using the `glob` crate, and explicitly error out if conversion to a utf-8 native rust string doesn't work.

src/analyzer.rs
```diff
@@ -52,9 +52,9 @@ impl<'run, 'src> Analyzer<'run, 'src> {
             self.assignments.push(assignment);
           }
           Item::Comment(_) => (),
-          Item::Import { absolute, .. } => {
-            if let Some(absolute) = absolute {
-              stack.push(asts.get(absolute).unwrap());
+          Item::Import { absolute_paths, .. } => {
+            for p in absolute_paths {
```
```suggestion
            for path in absolute_paths {
```

src/unstable_feature.rs
```diff
@@ -3,6 +3,7 @@ use super::*;
 #[derive(Copy, Clone, Debug, PartialEq, Ord, Eq, PartialOrd)]
 pub(crate) enum UnstableFeature {
   FormatSubcommand,
+  GlobImport,
```
casey, line 7:
```suggestion
  ImportPathGlob,
```

src/unstable_feature.rs
```diff
@@ -11,6 +12,7 @@ impl Display for UnstableFeature {
   fn fmt(&self, f: &mut Formatter) -> fmt::Result {
     match self {
       Self::FormatSubcommand => write!(f, "The `--fmt` command is currently unstable."),
+      Self::GlobImport => write!(f, "Glob imports are currently unstable"),
```
casey, line 22:
```suggestion
      Self::ImportPathGlob => write!(f, "Globs in import paths are currently unstable."),
```

src/parser.rs
```diff
@@ -365,7 +365,7 @@ impl<'run, 'src> Parser<'run, 'src> {
             let optional = self.accepted(QuestionMark)?;
             let (path, relative) = self.parse_string_literal_token()?;
             items.push(Item::Import {
-              absolute: None,
+              absolute_paths: vec![],
```
```suggestion
              absolute_paths: Vec::new(),
```

(No real reason, I just like avoiding macros.)

src/item.rs
```diff
@@ -7,7 +7,7 @@ pub(crate) enum Item<'src> {
   Assignment(Assignment<'src>),
   Comment(&'src str),
   Import {
-    absolute: Option<PathBuf>,
+    absolute_paths: Vec<PathBuf>,
```
casey: I would actually just keep this called `absolute`. It's a vec of paths, so we know that they're paths and there's potentially more than one of them.

src/compiler.rs
```diff
@@ -75,14 +86,65 @@ impl Compiler {
...
+                return Err(Error::UnstableFeature {
+                  unstable_feature: UnstableFeature::GlobImport,
+                });
+              }
+
+              let import_base_str =
+                import
+                  .as_os_str()
+                  .to_str()
+                  .ok_or_else(|| Error::ImportGlob {
```
casey:
```suggestion
                  .ok_or_else(|| Error::ImportGlobUnicode {
```

I'd like to use distinct error types instead of having a string message field.

src/compiler.rs
```diff
@@ -75,14 +86,65 @@ impl Compiler {
...
+                    path: *path,
+                  })?;
+              let glob_options = glob::MatchOptions {
+                case_sensitive: true,
+                require_literal_separator: false,
+                require_literal_leading_dot: false,
+              };
+              let import_paths =
+                glob::glob_with(import_base_str, glob_options).map_err(|error| {
+                  Error::ImportGlob {
```
casey:
```suggestion
                  Error::ImportGlobPattern {
```

I believe this top level glob error only happens if the pattern is bad.

src/compiler.rs
```diff
@@ -75,14 +86,65 @@ impl Compiler {
...
+                  if current.file_path.contains(&import) {
+                    return Err(Error::CircularImport {
+                      current: current.path,
+                      import,
+                    });
+                  }
+                  absolute_paths.push(import.clone());
+                  stack.push(current.import(import, path.offset));
+                } else if import.is_dir() {
+                  return Err(Error::ImportGlob {
```
casey:
```suggestion
                  return Err(Error::ImportGlobDirectory {
```

src/compiler.rs
```diff
@@ -75,14 +86,65 @@ impl Compiler {
               .join(Self::expand_tilde(&relative.cooked)?)
               .lexiclean();
 
-            if import.is_file() {
+            let has_glob = import.as_os_str().as_encoded_bytes().contains(&b'*');
```
casey, line 93: This is tricky because glob supports more meta-characters than `*`. It supports `?`, `*`, `[`, `!`, and `]`. I suppose we should check for all of them.

src/compiler.rs
```diff
@@ -75,14 +86,65 @@ impl Compiler {
...
+                import
+                  .as_os_str()
+                  .to_str()
+                  .ok_or_else(|| Error::ImportGlob {
+                    error: "non-UTF-8 glob import".to_string(),
+                    path: *path,
+                  })?;
+              let glob_options = glob::MatchOptions {
+                case_sensitive: true,
+                require_literal_separator: false,
```
casey: I believe this should be `true`. If `false`, then `/` can be matched by `*`, which is the `bash` default and `zsh` default. `**` can be used to match an arbitrary directory path, including `*`.

src/compiler.rs
```diff
@@ -75,14 +86,65 @@ impl Compiler {
...
+              if !(unstable || ast_unstable) {
+                return Err(Error::UnstableFeature {
+                  unstable_feature: UnstableFeature::GlobImport,
+                });
+              }
+
+              let import_base_str =
+                import
+                  .as_os_str()
+                  .to_str()
```
casey: We can do `to_str` on the path without first converting to an `OsStr`.

src/compiler.rs
```diff
@@ -75,14 +86,65 @@ impl Compiler {
...
+              };
+              let import_paths =
+                glob::glob_with(import_base_str, glob_options).map_err(|error| {
+                  Error::ImportGlob {
+                    error: error.to_string(),
+                    path: *path,
+                  }
+                })?;
+              for import in import_paths {
+                let Ok(import) = import else { continue };
```
casey: This swallows I/O errors. If we encounter an I/O error we should fail.

src/compiler.rs
```diff
@@ -75,14 +86,65 @@ impl Compiler {
...
+                  return Err(Error::ImportGlob {
+                    error: format!(
+                      "a glob import cannot match a directory; matched `{}`",
+                      import.display()
+                    ),
+                    path: *path,
+                  });
+                }
+              }
+            } else if import.is_file() {
```
casey, line 139: Is this circular import test duplicated with the one above?

casey (changes requested):
Did some initial review, see comments.

src/compiler.rs
neunenak: Can we assume that none of those characters will show up in a legitimate justfile? I think that's safe to assume for `*` but maybe not for characters like `[` and `]`.

src/item.rs
neunenak: My argument for changing it is that "absolute" is a pretty general word out of context, so it's hard to guess what it might mean when you see it as a random property access somewhere in code. I don't feel super strongly about it however.

src/compiler.rs
neunenak: What if there was a slightly different version of the `import` keyword that signaled that shell globs might be used, or an argument to it? `import glob "my/path/*.just"` or something like that?

src/compiler.rs
neunenak: It is, but I think that trying to de-duplicate those lines of code would add more complexity than it's worth, since the logic here is just constructing the `CircularImport` error if necessary and then manipulating some local variables.

casey:
I keep coming back to this. I have these worries:

- I worry about delegating glob parsing to a dependency we don't control, and thus may not understand the syntax.

- I worry about the backwards compatibility impact on existing justfiles. Since although it is crazy to use `*/?/**` in a path name, it does happen, and maybe especially on windows or GUI systems. This might also be extra confusing because the globbing character wouldn't appear in the path literal in the `justfile`.

- I worry about the restriction that it doesn't work with non-unicode paths.

I wonder if maybe it would be better if there was some syntax to enable the glob at the import, and it was that syntax which was experimental, instead of a setting which turns it on for all imports

Something like:

```just
glob import '*/mod.just'
```

woutervh:
looking forward to see something like this merged.

casey:
@neunenak What are you thinking about this? I think we should do this, and I think the combination of both making it unstable, and opting into explicitly for each import (syntax tbd, but `glob import "foo"` or `import glob "foo"` both seem fine), make me much more comfortable about dealing with the shortcomings later.

woutervh:
@neunenak Still working on this?

woutervh:
:(

woutervh:
> I worry about the backwards compatibility impact on existing justfiles. Since although it is crazy to use */?/** in a path name, it does happen, and maybe especially on windows or GUI systems. This might also be extra confusing because the globbing character wouldn't appear in the path literal in the justfile.
> I worry about the restriction that it doesn't work with non-unicode paths.


This is my simple use-case
I have dozens of toplevel justfiles that load project-specific  justfiles from a .just-subdirectory


her is an example toplevel file:
```
# Just is a crossplatform task-runner, similar to make.
# And justfiles are equivalent to makefiles.
#
# Official docs:
#  - https://just.systems/man/en
#  - installation: https://just.systems/man/en/pre-built-binaries.html
#
# Usage:
#   > just  # configured to display all available tasks
#   > just --help
#   > just <taskname>
#
# Notes:
#  - Comments immediately preceding a recipe will appear in just --list

# load environment variables from .env file
set dotenv-filename := ".env"
set dotenv-load := true

# search for justfiles in parent directories
set fallback

# set tempdir := "var/tmp"

# enable experimental features
set unstable

# set shell to powershell on Windows
set shell := ["bash", "-uc"]


# waiting for glob-support, see
# - https://github.com/casey/just/issues/1885
# - https://github.com/casey/just/pull/2376
import '.just/bandit.justfile'
import '.just/dir-structure.justfile'
import '.just/dotenv.justfile'
import '.just/ipython.justfile'
import '.just/just.justfile'
import '.just/mypy.justfile'
import '.just/pre-commit.justfile'
import '.just/project.justfile'
import '.just/pylint.justfile'
import '.just/pyroma.justfile'
import '.just/pytest.justfile'
import '.just/readthedocs.justfile'
import '.just/ruff.justfile'
import '.just/safety.justfile'
import '.just/sshx.justfile'
import '.just/ty.justfile'
import '.just/ubuntu.justfile'
import '.just/uv.justfile'


# Display all canonical tasks (default recipe)
list:
    @ just --list --unsorted --no-aliases
```

neunenak:
> @neunenak Still working on this?

ngl I've totally let this PR slip, and I've got other stuff going on, but yeah I would like to try to get this merged at some point. I would have to refresh myself with respect to what the bottlenecks were to getting this PR merged earlier, and it's been open long enough that I'd probably have to change a decent amount of stuff to get it mergeable. OTOH it's also been open so long that AI has gotten meaningfully better at coding since I opened it, so there's a positive I suppose :) (We live in an exciting time in human history).

Anyway I'll see what I can do here, definitely feel free to ping me in the future about this or any other issue I have open. Or if someone else wants to take this code and run with it, feel free.

woutervh:
@neunenak , @casey 

Can we focus on the most simple use-case to get started?
No unicode in the path. Error message is fine
No special pathnames like */?/**

casey:
I think this PR isn't likely to be merged any time soon. One of the main issues, which is kind of dealbreaker, is that the existing glob crates do not have an API that allows matching against non-unicode paths, which Just supports.

PR #2377, opened 2024-09-19
chore: remove redundant words in README.md

jingchanglu:


laniakea64:
Those repetitions are intentional and not redundant.  The first removal changes the meaning to be incorrect, and the second removal makes incorrect grammar.

To avoid immediate repetition of words, the first instance could be written as "Positional arguments may also be enabled on a per-recipe basis", and the second instance could be written as "Look for an appropriate file in `tests` to put the test in".  But I think the existing wording in the README is clear as-is.

casey:
Yah, these are actually correct, if a little awkward.

jingchanglu:
> Yah, these are actually correct, if a little awkward.

Thanks. Sorry for the interruption.

PR #2392, opened 2024-09-27
Reduce code duplication in suggestion methods

neunenak:
The existing `suggest_recipe` and `suggest_variable` methods have a decent chunk of duplicate code. This commit factors that out into a separate method.


casey:
I changed it to use the `min_by_key` method on iterators to avoid collecting into a vec. I'm also kind of a comment hater, unless it's on "thar be dragons" type code, so I removed it.

PR #2393, opened 2024-09-27
Invertible attributes

### LINKED ISSUE #1895: Allow inverted cfg attributes, like `[not(macos)]`

neunenak:
This PR implements a `not()` syntax for certain OS-target attributes.

Resolves https://github.com/casey/just/issues/1895

### DIFF
--- a/src/attribute.rs
+++ b/src/attribute.rs
@@ -14,17 +14,17 @@ pub(crate) enum Attribute<'src> {
   ExitMessage,
   Extension(StringLiteral<'src>),
   Group(StringLiteral<'src>),
-  Linux,
-  Macos,
+  Linux { enabled: bool },
+  Macos { enabled: bool },
   NoCd,
   NoExitMessage,
   NoQuiet,
-  Openbsd,
+  Openbsd { enabled: bool },
   PositionalArguments,
   Private,
   Script(Option<Interpreter<'src>>),
-  Unix,
-  Windows,
+  Unix { enabled: bool },
+  Windows { enabled: bool },
   WorkingDirectory(StringLiteral<'src>),
 }
 
@@ -53,6 +53,7 @@ impl<'src> Attribute<'src> {
   pub(crate) fn new(
     name: Name<'src>,
     arguments: Vec<StringLiteral<'src>>,
+    enabled: bool,
   ) -> CompileResult<'src, Self> {
     let discriminant = name
       .lexeme()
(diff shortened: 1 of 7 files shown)


neunenak:

src/attribute_set.rs
```diff
@@ -12,6 +18,28 @@ impl<'src> AttributeSet<'src> {
...
+        | Attribute::Openbsd { inverted }
+        | Attribute::Unix { inverted }
+        | Attribute::Windows { inverted } => {
+          if *inverted {
+            InvertedStatus::Inverted
+          } else {
+            InvertedStatus::Normal
+          }
+        }
+        _ => return None,
```
casey: I think this function should panic if we call it with a non-invertible attribute.

src/recipe.rs
```diff
@@ -116,19 +116,81 @@ impl<'src, D> Recipe<'src, D> {
...
+    let disabled = Systems {
+      linux: matches!(linux, Some(InvertedStatus::Inverted)),
+      macos: matches!(macos, Some(InvertedStatus::Inverted)),
+      openbsd: matches!(openbsd, Some(InvertedStatus::Inverted)),
+      unix: matches!(unix, Some(InvertedStatus::Inverted)),
+      windows: matches!(windows, Some(InvertedStatus::Inverted)),
+    };
+
+    if cfg!(target_os = "linux") {
+      return !(disabled.linux || disabled.unix)
```
casey: Are these `unix` checks redundant with the ones below? I notice that the other Unix OS's don't have the same checks.

src/recipe.rs
```diff
@@ -116,19 +116,81 @@ impl<'src, D> Recipe<'src, D> {
...
-    let windows = self.attributes.contains(AttributeDiscriminant::Windows);
-
-    (!windows && !linux && !macos && !openbsd && !unix)
-      || (cfg!(target_os = "linux") && (linux || unix))
-      || (cfg!(target_os = "macos") && (macos || unix))
-      || (cfg!(target_os = "openbsd") && (openbsd || unix))
-      || (cfg!(target_os = "windows") && windows)
-      || (cfg!(unix) && unix)
-      || (cfg!(windows) && windows)
+    use attribute_set::InvertedStatus;
```
casey: I'm not sure I like introducing inverted status only to turn it back to a bool with `matches!`. I think it would probably be better to just use a bool the whole way.

src/attribute.rs
```diff
@@ -13,17 +13,17 @@ pub(crate) enum Attribute<'src> {
   Doc(Option<StringLiteral<'src>>),
   Extension(StringLiteral<'src>),
   Group(StringLiteral<'src>),
-  Linux,
-  Macos,
+  Linux { inverted: bool },
```
casey: I think we should probably use:

```suggestion
  Linux { enabled: bool },
```

And invert the meaning of the boolean, to make the downstream code make more sense.

src/attribute.rs
```diff
@@ -75,29 +76,38 @@ impl<'src> Attribute<'src> {
...
-      AttributeDiscriminant::Macos => Self::Macos,
-      AttributeDiscriminant::NoCd => Self::NoCd,
-      AttributeDiscriminant::NoExitMessage => Self::NoExitMessage,
-      AttributeDiscriminant::NoQuiet => Self::NoQuiet,
-      AttributeDiscriminant::Openbsd => Self::Openbsd,
-      AttributeDiscriminant::PositionalArguments => Self::PositionalArguments,
-      AttributeDiscriminant::Private => Self::Private,
-      AttributeDiscriminant::Script => Self::Script({
+    Ok(match (inverted, discriminant) {
+      (inverted, AttributeDiscriminant::Linux) => Self::Linux { inverted },
```
casey:
```suggestion
      (inverted, AttributeDiscriminant::Linux) => Self::Linux { enabled: !inverted },
```

src/parser.rs
```diff
@@ -1135,7 +1135,18 @@ impl<'run, 'src> Parser<'run, 'src> {
       token.get_or_insert(bracket);
 
       loop {
-        let name = self.parse_name()?;
+        let (name, inverted) = {
```
casey, line 1194: I think this is probably cleaner than using mutable variables:

```suggestion
        let (name, inverted) = {
          let name = self.parse_name()?;
          if name.lexeme() == "not" {
            self.expect(ParenL)?;
            let name = self.parse_name()?;
            (name, true)
          } else {
            (name, false)
          }
        }
```

casey: And I think we should actually, since we don't have any invertable variables which aren't system variables, change the argument to `Attribute::new` to be !inverted, and within `Attribute::new` it's called `enabled`.

src/recipe.rs
casey: I think this maybe all gets easier if we have a `System` enum which represents the current system:

```
enum System {
  Windows,
  Linux,
  Unix,
  ...
}

impl System {
  fn current() -> &'static [System] {
    if cfg(target_os = "linux") {
       &[Self::Linux, Self::Unix]
    }
  }
}
```

And then we can handle all these checks in the same way.

casey (changes requested):
See some initial comments!

src/recipe.rs
```diff
@@ -16,6 +16,76 @@ fn error_from_signal(recipe: &str, line_number: Option<usize>, exit_status: Exit
...
+    if cfg!(target_os = "macos") {
+      return MacOS;
+    }
+    if cfg!(target_os = "windows") || cfg!(windows) {
+      return Windows;
+    }
+    if cfg!(unix) {
+      return Unix;
+    }
+    panic!("No recognized system");
```
casey: I don't think this can panic. We should have an `Unrecognized` variant.

src/recipe.rs
```diff
@@ -16,6 +16,76 @@ fn error_from_signal(recipe: &str, line_number: Option<usize>, exit_status: Exit
...
+  }
+
+  fn enabled(self, enabled: SystemMap, disabled: SystemMap) -> bool {
```
casey: I think we need some kind of more extensive testing for this. It looks right to me, but it strikes me as very fragile, and easy to get wrong when extending..

src/recipe.rs
```diff
@@ -16,6 +16,76 @@ fn error_from_signal(recipe: &str, line_number: Option<usize>, exit_status: Exit
...
+    match self {
+      System::Windows => {
+        !disabled.windows
+          && (enabled.windows
+            || !(enabled.macos || enabled.linux || enabled.openbsd || enabled.unix))
+      }
+      System::MacOS => {
```
casey: On MacOS, and other unix systems, a recipe with `[not(unix)]` will be enabled. This seems wrong to me.

casey (changes requested):
See comments. This looks great, I like the structure, but we definitely need to do some kind of more extensive testing.

src/recipe.rs
neunenak: Yeah, makes senses - I'm not not entirely happy with the fragility of this code either, but I don't have any ideas off the top of my head for how to make it more robust. Do you have any ideas for how the tests should be architected?

casey: One idea:

```
#[test]
fn foo() {
  fn case(attributes: &str, active: bool) {
    Test::new().justfile(format!("[{attributes}]\nfoo:"))
    // assert that foo is or isn't active, depending on attributes.
  }

  case("macos",  cfg!(target_os = "darwin"));
}
```

One idea for improving the structure would be to try to factor out this pattern:

```
      System::Windows => {
        !disabled.windows
          && (enabled.windows
            || !(enabled.macos || enabled.linux || enabled.openbsd || enabled.unix))
      }
```


Maybe with maps:

```
        !disabled[self] && (enabled[self] || !self.others().iter().any(|other| system.enabled()))
```

casey:

offbyone:
> Closing, since this has been sitting for a while.

If someone else picked it up would you re-open? I've also got a desire for this.

casey:
@offbyone Pull requests are disabled, see #3227. I know it's a coveted feature though, so it's on my radar.

offbyone:
Legit! Given there's a start on it, and it's on your radar, I'll ... "just" wait for it.

Thanks!

PR #2395, opened 2024-09-27
Tidy up error message of interrupt signal for default 'fzf' chooser

taufik-rama:
Hello :wave: 

I'm kind of just starting to use `just` & I think I like it. While I was just trying things out I see that this tool can do a "chooser" mode where the user is guided from interactive CLI, which I think is kind of nice

I have an improvement idea regarding some error message thrown during `chooser` mode that uses the default `fzf` command. Basically right now when I do this:

```justfile
default:
    just --choose

ping:
    echo pong
```

It'll take me to the expected CLI menu
image

But when I ^C or ^D it it'll throw some (unexpected) error

```shell
just --choose
error: Chooser `fzf --multi --preview 'just --unstable --color always --justfile "/some/path/justfile" --show {}'` failed: exit status: 130
error: Recipe `default` failed on line 2 with exit code 130
```

I think since it is expected that exit code 130 to be [`ExitInterrupt`](https://github.com/junegunn/fzf/blob/30a8ef28cdc1d23cf6956f57ca05bd28db515014/src/constants.go#L74), we can just use it to have better error message

The new error message would be

```shell
just --choose
error: Command was terminated by signal 130
error: Recipe `default` failed on line 2 with exit code 1
```

IMO better error message would be something along the line of just `Command was interrupted` without any signal code, since that signal is `fzf`-specific, but current ones convey the meaning just fine too

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -154,7 +154,7 @@ pub(crate) enum Error<'src> {
     recipe: &'src str,
   },
   Signal {
-    recipe: &'src str,
+    recipe: Option<&'src str>,
     line_number: Option<usize>,
     signal: i32,
   },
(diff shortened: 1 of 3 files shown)


src/error.rs
```diff
@@ -154,7 +154,7 @@ pub(crate) enum Error<'src> {
     recipe: &'src str,
   },
   Signal {
-    recipe: &'src str,
+    recipe: Option<&'src str>,
```
taufik-rama, line 157: it's possible for the user to just do `just --chooser` without specifying the recipe, so I changed this into an `Option`

marcaddeo:
You can also put `[no-exit-message]` above your `default:` recipe to suppress that error.

taufik-rama:
TIL, though I don't know about whether it'll work for all usecases, since I don't see any problem with exit error messages itself, just that it could be tidied up

casey:
I don't think that special-casing this is worth it. It feels a bit hacky, for example it doesn't work it the user is using `fzf` as their chooser, but with a custom invocation. Also, I think you really don't want an error message at all if you ctrl-c `fzf`, since there's not really an error, per se, so I think `[no-exit-message]` is a better solution.

PR #2419, opened 2024-10-08
Add AttributeSet type

neunenak:
Replace the several places that directly use a `BTreeSet<Attribute>` with a wrapper data structure `AttributeSet`, that provides some convenience methods for working with attributes.


neunenak:
I think having a data structure like this will make it easier to implement https://github.com/casey/just/issues/1895 , and it's an independently-reviewable chunk of work so it doesn't have to be part of https://github.com/casey/just/pull/2393 .

src/analyzer.rs
```diff
@@ -69,17 +69,17 @@ impl<'run, 'src> Analyzer<'run, 'src> {
           } => {
             let mut doc_attr: Option<&str> = None;
             let mut groups = Vec::new();
-            for attribute in attributes {
+            attributes.ensure_valid_attributes(
```
casey, line 75: I think we should keep the error checking in the `else` below, otherwise the set of allowed attributes is in two places, here, where we call `ensure_valid_attributes`, and below, in the `if / else` statement.

src/attribute.rs
```diff
@@ -129,6 +129,73 @@ impl<'src> Display for Attribute<'src> {
   }
 }
 
+#[derive(Default, Debug, Clone, PartialEq)]
+pub(crate) struct AttributeSet<'src> {
+  inner: BTreeSet<Attribute<'src>>,
+}
+
+impl<'src> Serialize for AttributeSet<'src> {
```
casey: Can we avoid needing to write a custom a serialize implementation if we make the struct:

```
pub(crate) struct AttributeSet<'src>(BTreeSet<Attribute<'src>>);
```

I think in that case serde will serialize the struct as just the inner contents.

src/attribute.rs
```diff
@@ -129,6 +129,73 @@ impl<'src> Display for Attribute<'src> {
...
+}
+
+impl<'src> AttributeSet<'src> {
+  pub(crate) fn from_iter(iter: impl IntoIterator<Item = Attribute<'src>>) -> Self {
+    Self {
+      inner: iter.into_iter().collect(),
+    }
+  }
+
+  pub(crate) fn count(&self) -> usize {
```
casey: I think we should just call this `len`, since that's what native collections use.

src/attribute.rs
```diff
@@ -129,6 +129,73 @@ impl<'src> Display for Attribute<'src> {
...
+    self.inner.len()
+  }
+
+  pub(crate) fn contains(&self, target: AttributeDiscriminant) -> bool {
+    self.inner.iter().any(|attr| {
+      let discriminant: AttributeDiscriminant = attr.into();
```
casey: Let's add a `fn discriminant(&self) -> AttributeDiscriminant { self.into() }` function to `Attribute`, so we can avoid needing to use type inference to do the conversion.

src/attribute.rs
```diff
@@ -129,6 +129,73 @@ impl<'src> Display for Attribute<'src> {
...
+      let item_discriminant: AttributeDiscriminant = (*attr).into();
+      discriminant == item_discriminant
+    })
+  }
+
+  pub(crate) fn iter(&self) -> impl Iterator<Item = &Attribute<'src>> {
+    self.inner.iter()
+  }
+
+  pub(crate) fn private(&self) -> bool {
```
casey: I think we should avoid one-off helpers like this, unless they do something more complex, since doing `attributes.contains(AttributeDiscriminant::Private)` isn't so bad.

src/parser.rs
```diff
@@ -1128,7 +1115,8 @@ impl<'run, 'src> Parser<'run, 'src> {
     if attributes.is_empty() {
       Ok(None)
     } else {
-      Ok(Some((token.unwrap(), attributes.into_keys().collect())))
+      let attribute_set = AttributeSet::from_iter(attributes.into_keys());
```
casey: Can you do `collect` here instead of `from_iter`?

src/recipe.rs
```diff
@@ -386,13 +389,16 @@ impl<'src, D> Recipe<'src, D> {
...
     let mut path = tempdir.path().to_path_buf();
 
-    let extension = self.attributes.iter().find_map(|attribute| {
-      if let Attribute::Extension(extension) = attribute {
-        Some(extension.cooked.as_str())
-      } else {
-        None
-      }
-    });
+    let extension = self
```
casey: I think the original version of this is actually better, since we don't duplicate which attribute we're looking for.

src/recipe.rs
```diff
@@ -460,10 +466,8 @@ impl<'src, D> Recipe<'src, D> {
   }
 
   pub(crate) fn doc(&self) -> Option<&str> {
-    for attribute in &self.attributes {
-      if let Attribute::Doc(doc) = attribute {
-        return doc.as_ref().map(|s| s.cooked.as_ref());
-      }
+    if let Some(Attribute::Doc(doc)) = self.attributes.get(AttributeDiscriminant::Doc) {
```
casey: This is kind of a weird one. This `if let` will only match if *both* `self.attributes.get` returns `Some`, and the returned attribute is an `Attribute::Doc`, if the code is correct, of course, both or neither will be the case.

I think a more scrupulous correct version would be:

```
if let Some(attribute) = self.attributes.get(AttributeDiscriminant::Doc) {
  if let Some(Attribute::Doc(doc)) else {
    unreachable!();
  }
}
```

Although that's a little crazy, so I think maybe the original on the left is better.

src/recipe.rs
```diff
@@ -479,7 +483,7 @@ impl<'src, D: Display> ColorDisplay for Recipe<'src, D> {
       writeln!(f, "# {doc}")?;
     }
 
-    for attribute in &self.attributes {
+    for attribute in self.attributes.iter() {
```
casey: We should implement `IntoIterator` for `&AttributeSet` so we can avoid needing to call `.iter()`.

casey (changes requested):
Nice, this is a definite improvement. See comments!

src/analyzer.rs
neunenak: If we keep the error checking here, then we would have pretty similar error-generation code here and in the `ensure_valid_attributes` method. I added `ensure_valid_attributes` as a way of de-duplicating attribute error generation code. I modified the code below to add an `unreachable` and hopefully make it clearer what's going on.

src/parser.rs
neunenak: Not quite sure what code change you want to see here.

neunenak:
@casey were there any more tweaks you wanted to make to this PR? It looks like some subsequent code merges have caused conflict with this PR, I'll try to fix them unless you had something specific you were trying to do with your commits.

casey (approved):
Fixed the merge conflicts. I think this is fine. I have slight misgivings about needing to keep the sets of valid attributes in the ensure_valid_attributes call and match statement in sync, but I can't actually think of a way to improve it.

Issue #2272, opened 2024-07-24
`just --list` should highlight parts of the comment surrounded by backticks (`)


nyurik:
If a recipe uses backticks in the description comment, I think it would be helpful to the users to show that content in a different color:

```just
# run command with docker, e.g. `just docker bazel build //:mbgl-core`, or open docker shell with `just docker`
docker *ARGS:
    docker run --rm -it -v "$PWD:/app/" -v "$PWD/docker/.cache:/home/user/.cache" maplibre-native-image {{ARGS}}
```

When using `just --list`, the above should be shown something like this (probably with better color choices):

![image](https://github.com/user-attachments/assets/322fe9fc-e399-4da9-b242-bb994c7ec725)

casey:
This seems like a good idea to me!

Blacksmoke16:
I think this can be closed now given https://github.com/casey/just/pull/2423 was merged.

casey:
Good catch!

PR #2423, opened 2024-10-09
Highlight text within backticks when listing recipes

neunenak:
https://github.com/casey/just/issues/2272


casey:
Seems reasonable! Needs some tests though. Should definitely test edge cases, like an unclosed backtick, etc.

casey:
I could not wrap my head around what `split_inclusive` was doing, so I replaced it with a regex. This also avoids color codes sometimes when they aren't needed.

laniakea64:
Although this type of highlighting could be nice in theory, the chosen color makes it only worse.  The white backtick color stands out too much relative to...**everything else** in the output, to the point of making `just --list` output a struggle to read.  It's especially distracting in the auto-generated descriptions for aliases.

Was going to check whether changing the backtick highlight to bright blue (`0;94`) would resolve this while still standing out enough, but can't figure out how to specify bright blue in this part of `just` code?

casey:
It looks okay for me, but that's probably just my color theme. I can kinda see how it stands out too much See below. Can you post a screenshot of what it looks like for you?

Screenshot 2024-11-16 at 4 10 50 PM

Cyan might be a good choice, since it goes with blue and is bright. It's the same color used for parameters, but that seems fine, and sometimes you use backticks in doc comments to call out what parameters do:

Screenshot 2024-11-16 at 4 15 29 PM

casey:
Changed the color to cyan in #2469. Hopefully that looks good in more color schemes!

laniakea64:
> It looks okay for me, but that's probably just my color theme. 

In your color scheme it would be fine for me too.

> Can you post a screenshot of what it looks like for you?

Sure.

White -

Screen shot 2024-11-17 1

Cyan -

Screen shot 2024-11-17 2

The cyan is much better, thanks!

PR #2424, opened 2024-10-10
doc(cn): add document for Running tasks in parallel in CN

Jannchie:
I found that the Chinese documentation did not mention running tasks in parallel, I added this part.


casey:
Thanks! The Chinese documentation is sadly getting pretty out of date.

Issue #1323, opened 2022-08-29
Consider using ariadne for error messages


casey:
Error messages are a complex part of just which require quite a bit of tricky code. https://github.com/zesterer/ariadne is a nice looking crate which we should consider using for error messages, which might enable us to both rip out our existing code, and make error messages better at the same time.

PR #2426, opened 2024-10-10
Use ariadne error reporting library for compile errors

neunenak:
This commit starts introducing the `ariadne` rust error-reporting crate to `just`. In order to make individual commits a bit more tractable, this commit only introduces the new ariadne error formatting for compile errors, leaving other sorts of errors alone for the time being.

With the deliberately-broken justfile:

```just

[group: ]
test_recipe:
    echo "TEST"

```

Running `just` now results in an error that looks like:

image

This commit also updates the minimum rust version to 1.85, since the latest version of `ariadne` also requires that minimum version to compile; in the process of porting the tests, I also noticed that a couple of tests, namely `unknown_start_of_token_invisible_unicode` and `unknown_start_of_token_ascii_control_char`, had raw unprintable ascii literals in the source file, which was causing some problems with e.g. running ripgrep on the file containing them. So I changed them to use unicode escapes instead. Apart from this, all the tests should be exactly the same as they were previously, except for testing for the new error formatting.

Pertains to https://github.com/casey/just/issues/1323


neunenak:
I deliberately introduced a compile error, here's what that looks like currently:

```
just -l
error: Attribute `group` got 2 arguments but takes 1 argument
  ——▶ justfile:15:2
   │
15 │ [group('test', 'bogus extra argument')]
   │  ^^^^^
   ```
   
   and here's what that looks like with ariadne as it's currently set up in that PR:
   
   ```
   [01] Error: Attribute argument count mismatch
    ╭─[justfile:15:2]
    │
 15 │ [group('test', 'bogus extra argument')]
    │  ──┬──
    │    ╰──── Found 2 arguments
    │
    │ Note: `group` takes 1 argument
────╯
```

slightly nicer I suppose

casey:
Is the formatting off in that error message? It looks misaligned.

neunenak:
> Is the formatting off in that error message? It looks misaligned.

Huh, yeah, must've happened when I copy-pasted it from the terminal, it looks fine there. Screenshot:
ariadne

And the current error message in the same terminal:


current

neunenak:
Another error:

duplicate

casey:
Tentatively, I think this looks good. This will be a *massive* change though, because there are so many tests which check error messages.

I think the easiest way to do this would be to land a commit which switches to ariadne, but which doesn't change any error messages or try to add any additional highlights. I think we could write a script to automatically convert existing tests to the new format.

Then we could start improving the error messages one by one, by adding additional spans, etc.

Some questions:

- Can "Error:" be made lowercase?
- Can the error numbers (i.e. `[E01]`) be removed?
- Can the underline be made red, to indicate that it's an error? (For existing errors, the underlined token is always an error, so it should be red, and then if we add additional spans, like hints, those could be a different color)

neunenak:
> Tentatively, I think this looks good. This will be a _massive_ change though, because there are so many tests which check error messages.
> 
> I think the easiest way to do this would be to land a commit which switches to ariadne, but which doesn't change any error messages or try to add any additional highlights. I think we could write a script to automatically convert existing tests to the new format.
> 
> Then we could start improving the error messages one by one, by adding additional spans, etc.

This makes sense


> Some questions:
> 
>  Can "Error:" be made lowercase?

Yeah. You have to do this in the kinda-silly way of defining `ReportKind::Custom("error", Color::Red)`, and using that in place of `ReportKind::Error` when building the ariadne `Report`, but that works fine. 

>  Can the error numbers (i.e. `[E01]`) be removed?

Yeah, that's totally optional. I kind of like the idea of having a rustc-style authoritative list of error codes, but that's not worth trying to merge in the first pass in any case.

> Can the underline be made red, to indicate that it's an error? (For existing errors, the underlined token is always an error, so it should be red, and then if we add additional spans, like hints, those could be a different color)

Yeah there's an API for assigning pretty arbitrary colors to labels:

formatting

casey:
> Yeah. You have to do this in the kinda-silly way of defining `ReportKind::Custom("error", Color::Red)`, and using that in place of `ReportKind::Error` when building the ariadne `Report`, but that works fine.

Dope.

> Yeah, that's totally optional. I kind of like the idea of having a rustc-style authoritative list of error codes, but that's not worth trying to merge in the first pass in any case.

I like the idea of doing that too, although definitely later.

> > Can the underline be made red, to indicate that it's an error? (For existing errors, the underlined token is always an error, so it should be red, and then if we add additional spans, like hints, those could be a different color)
> 
> Yeah there's an API for assigning pretty arbitrary colors to labels:

Dope.

So yeah, I think this is a good change, although the search and replace for changing all the tests will be brutal.

casey:
An idea that might make this easier: Can we style the adriane errors so that they look the same as `just` errors, which would let us merge this with minimal changes to the tests? We could follow up with PRs which improved the style using adriane features, which would require changing the tests, but hopefully could be entirely automated. This PR is likely to keep accumulating conflicts which might make it hard to land.

neunenak:
> An idea that might make this easier: Can we style the adriane errors so that they look the same as `just` errors, which would let us merge this with minimal changes to the tests? We could follow up with PRs which improved the style using adriane features, which would require changing the tests, but hopefully could be entirely automated. This PR is likely to keep accumulating conflicts which might make it hard to land.

I'm not sure there's an easy way to make the error output look like just errors. I was having some trouble with tests not matching, which I realized was because ariadne was always emitting ansi color codes even in tests; so I added some functionality to pass color enable/disable to ariadne which solves that problem.  Now it's just a matter of updating the tests, which I was trying to find a way to use an AI code editor to do. If not I'll write a script for it or do it manually.

neunenak:
@casey in the year or so since I last looked at this, LLMs got good enough to make the code changes in a (relatively) non-tedious way, so this PR is ready for review.

casey (changes requested):
I just checked this out, and source locations are missing underlining:

Screenshot 2026-03-29 at 3 22 23 PM

There should be an indication of where the error occurred, and it shouldn't just be a color change, in case colors are disabled.

Also, `Error` is now uppercase, when it should be lowercsae.

Also, for the initial PR, I think we should remove special rendering for specific error messages. It's nice, but they make the PR harder to review, and I'd rather follow up with individual improvements to specific error messages.

casey:

PR #2430, opened 2024-10-15
Deprecate backticks

offby1:
Just a sentence suggesting people use `shell` instead of backticks.

Oh look, I just used backticks up there!! :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2034,6 +2034,8 @@ Use `just --groups --unsorted` to print groups in their justfile order.
 
 ### Command Evaluation Using Backticks
 
+Backticks are an older feature that have largely been obsoleted by the `shell` function (see [External Commands](#external-commands).); we recommend that you use `shell` in new justfiles.
+
 Backticks can be used to store the result of commands:
 
 ```just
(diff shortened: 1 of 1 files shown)


laniakea64:
Why deprecate?

Backticks are still useful for concisely expressing simple, fixed (i.e. not `just` expressions) commands.  Using `shell()` where a backtick can do the job doesn't address any potentially-unexpected behavior and doesn't add clarity.

Also, backticks are not equivalent to `shell()`:
```
backtick := `echo "$0" "$@"`
shell := shell('echo "$0" "$@"')
```
```
$ just --evaluate
backtick := "sh"
shell    := "echo "$0" "$@""
```

Mentioning `shell()` in this part of the README seems reasonable to me, but maybe better along the lines of "More flexible and more powerful command invocations can be achieved with the [`shell()` function](https://just.systems/man/en/functions.html#external-commands)" and written at the end of the section

casey:
Nice, merged! Yah, unlike in shell scripts, backticks in just don't have any weird behaviors, so no need to deprecated them. The note is nice though.

PR #2431, opened 2024-10-15
Emphasize that variable assignments can use functions.

offby1:
I didn't realize until today that I could use functions in variable assignments, probably because the doc for the latter only mentioned that they work in recipes.


casey:
Thanks for the PR! Yah, good idea. I also changed "just provides a few built-in functions" to "just provides many built-in functions", since they've really multiplied over the years 😅

PR #2437, opened 2024-10-21
Allow interchangable recipe redefinitions

### LINKED ISSUE #2436: Importing many times the same justfile.
Hi.

I am writing some multiple just fragments to handle some aspects of my build systems. Some of them, say `A.just` and `B.just` depend on another, `C.just`. Essentially, I'd like to write `import "A.just"` and/or `import "B.just"`. When I import only one of them, it works fine: A.just is importing C.just, or B.just is importing C.just. No conflict. When I import **both** of them, the tasks in `C.just` conflict as they are declared twice, once for each import.

Ideally, there should be some "caching" mechanism that would allow the C.just file to be detected to be the same, and thus imported only once. No ?

casey:
Fixes #2436.

@neunenak Does this seem reasonable to you? See the original issue, but basically, if you have the same recipe imported in multiple places in the same module, you'll get an error. This seems safe to allow, since they're exactly the same recipe. Allowing it is kind of like automatically wrapping recipes in C-style include guards.

This is a bit hacky, since it only works for recipes. The less hacky way to do it would be to just ignore an import if it's a file that's already been imported into this module, which would work for everything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3321,6 +3321,8 @@ import? 'foo/bar.just'
 
 Missing source files for optional imports do not produce an error.
 
+Importing the same source file multiple times is not an error.
+
 ### Modules<sup>1.19.0</sup>
 
 A `justfile` can declare modules using `mod` statements.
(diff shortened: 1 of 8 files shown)


neunenak:
Yeah, seems sensible enough

PR #2438, opened 2024-10-22
feat: Add working-directory attribute

### LINKED ISSUE #2082: Add `[cd(DIR)]` attribute
Allow setting the working directory of a recipe to an arbitrary directory. Good first issue!

bcheidemann:
Closes https://github.com/casey/just/issues/2082


src/recipe.rs
```diff
@@ -133,11 +133,24 @@ impl<'src, D> Recipe<'src, D> {
...
   fn working_directory<'a>(&'a self, context: &'a ExecutionContext) -> Option<PathBuf> {
-    if self.change_directory() {
-      Some(context.working_directory())
-    } else {
-      None
+    if !self.change_directory() {
+      return None;
     }
+
+    let working_dir = self
```
casey: I think this is better implemented as a for loop:

```rust
let working_directory = context.working_directory();

for attribute in self.attributes {
  if let Attribute::WorkingDirectory(dir) = attribute {
    return Some(working_directory.join(dir.cooked);
  }
}

Some(working_directory)
```

Also note the use of `cooked`, which is the contents of the string literal after processing escape sequences.

tests/working_directory.rs
```diff
@@ -331,3 +331,70 @@ file := shell('cat file.txt')
...
+fn attribute() {
+  Test::new()
+    .justfile(
+      r#"
+      [working-directory('bar')]
+      print1:
+        echo "$(basename "$PWD")"
+
+      [working-directory('bar')]
+      [working-directory('baz')]
```
casey: Duplicate working directory attributes should be an error.

casey (changes requested):
Reviewed, see comments!

src/recipe.rs
bcheidemann: > I think this is better implemented as a for loop

Yup, the suggested for loop reads better and is more concise. Just curious, is there any other reason you suggested this refactor?

> Also note the use of cooked, which is the contents of the string literal after processing escape sequences.

Good to know - thanks for the info!

casey (approved):
Sorry for taking so long to get this in! I made `[no-cd]` and `[working-directory]` conflict, so you can't have them both on a recipe. I also added an `expect_file` function to the test runner, which allowed the tests to be simplified, since you can `touch` a file and check that a file in the expected directory was created.

bcheidemann:
Thanks @casey! Looking forward to using this in my projects 😍

CoreyLivingston:
Nice, happy to have this.

PR #2439, opened 2024-10-23
Add `xdg_*_directory` functions

### LINKED ISSUE #2434: XDG directory functions do not respect XDG variables on macOS
Due to long standing difficulties with (a/the) maintainer of dirs-rs, they refuse to recognize XDG environment variables explicitly defined by the user on macOS: https://github.com/dirs-dev/directories-rs/issues/47. `config_directory()` will always go to `/Users/<username>/Library/Application Support` on macOS, even if you have set `XDG_CONFIG_HOME`. I would suggest using https://github.com/lunacookies/etcetera instead, though there may be other alternatives that also respect user preferences.

uncenter:
Closes #2434. Adds a new set of functions like the existing directory functions, but prefixed with `xdg_` and return true paths to XDG directories. Tried my best at updating the README as well.


uncenter:
I also don't want to bloat just, so we may want to look into using etcetera's `choose_native_strategy` to replicate `dirs`'s functionality (if it does fully match as it does now for backwards compat) and remove `dir` entirely.

casey:
Closing for now, see #2434.

Issue #2411, opened 2024-10-04
Allow chaining conditionals


JohnstonJ:
I wanted to write some code like this:

```
export VCPKG_DEFAULT_TRIPLET := if os_family() == "windows" && arch() == "x86_64" {
    "x64-windows-static"
} else if <snip> {
    <snip>
}
```

But apparently most common operators aren't implemented beyond what looks like string concatenation (judging from the error message).  For example, logical operators would be quite useful here.

```
error: Expected '{', '+', or '/', but found '&&'
 ——▶ justfile:1:61
  │
1 │ export VCPKG_DEFAULT_TRIPLET := if os_family() == "windows" && arch() == "x86_64" {
  │                                                             ^^
```

_This time_, I can work around this with a condition like `if os_family() + "-" + arch() == "windows-x86_64" {`, but I would imagine there are other scenarios out there where it doesn't work out as cleanly.

It seems that GNU make has similar limitations: https://stackoverflow.com/questions/6451477/makefile-ifeq-logical-and

But I don't know if those limitations from GNU make are worth keeping in just?  None of the proposed solutions in the above StackOverflow link look "better" to me than what the original poster was trying to do.

Some operators I could imagine being useful:
- Logical operators:
  - `&&`
  - `||`
  - `!` (unary NOT)
  - maybe an XOR operator
- Math operators: if the variable holds a number, then do some math on it. (`+`, `-`, `*`, `/`, etc).  But unless just gains a type system, IDK if that's something you'd want to do (else, `+` is ambiguous with strong concatenation)...

To me, the logical operators would be the most nice ones to have, but maybe keep options open for math operators in the future?

casey:
Thanks for opening this issue!

Yah, conditionals are a massive hack. They exist as a special case in the grammar of conditionals, not as expressions. My motivation for implementing them in this terrible, awful, no good way, was because I've wanted to avoid committing to the representation of `true` and `false` in `just`. If logical operations are just allowed in conditionals, you can't get the result of evaluating one, so we don't need to decide what they evaluate too.

This is tricky, because I can imagine that one might want `&&` and `||` to behave in ways which aren't exactly the same as a logical operator that only takes `true` and `false`.

For example, you might want:

```
"a" || "" -> "a"
"" || "b" -> "b"
"a" || "b" -> "a"
"" || "" -> ""

"a" && "" -> ""
"" && "b" -> ""
"a" && "b" -> "b"
"" || "" -> ""
```

[...]

Issue #2334, opened 2024-08-30
Add variable to have just require a program to be present on the system


nogweii:
As a command runner, it would be convenient to be able to specify the list of programs necessary for my Justfile to run. Something like the following example:

```
set required-programs := ["kubectl", "helm", "helmfile"]

bootstrap: && helm-charts
    kubectl apply -f foo.yaml

helm-charts:
    helmfile apply --file bar.yaml

status:
    helm ls -A
```

And if I didn't have those programs installed, just would error out before running a recipe with a helpful message saying that I need to install, say, `helm` & `helmfile`.

If this is added, I can imagine a need for `[require-optional]` (with a better name though) if people wanted to provide a recipe to install all the missing programs.

This is basically automatically doing #2109 and reporting an error. If that function is added, this could be solved by creating a recipe that does this. However, I still think it's worthwhile to include this feature.

casey:
You could do this with a recipe:

```
require-dependencies:
  helm --version || echo "please install helm"

status: require-dependencies
    helm ls -A
```

This is more verbose, since you have to add that recipe as a dependency, but more flexible, since you can add that recipe as a dependency only to those recipes which require the binaries your checking for.

I think actually the way to support this is #2178, which would allow defining a recipe to be run before any other recipe in the justfile, which would allow for checking any kind of precondition, including whether or not particular binaries are installed.

PR #2440, opened 2024-10-25
Add which function for finding executables in PATH

### LINKED ISSUE #2109: `executable_exists` function to check whether a certain program exists and is executable on the path
As an example I want my install scripts to use `nala` if available and fall back to `apt-get` if not.

0xzhzh:
Closes #2109.

I considered calling it `executable_exists` but that seemed limiting---`which` also returns the path of the executable, not just whether it exists. Then one can implement `executable_exists(name)` using `which(name) == ""`.

I wasn't sure whether to add more tests. Most of the interesting test cases are the responsibility of the `which` crate to get right (eg executable existing in multiple directories in PATH, executable existing in PATH but not having executable permissions, etc). At the end of the day this function is just a wrapper around `which::which`, so there is not much more to test here than what should already be tested by `which::which`.

Also not sure if documentation was added in the right place.


casey:
Thanks for the PR! I this is definitely useful.

One thought is that perhaps this function should actually fail, as in, produce an error and terminate execution, if the binary does not exist? It's very common that if a binary isn't available, you can't do anything useful, and so you just want to give up. In that case, you would be required to check that `which` wasn't producing the empty string before calling any command it returns.

Tests which are just testing the functionality of `which` aren't necessary. In general, I try to avoid testing dependencies, and just assume they work (assuming that they're relatively popular, seem well maintained, etc).

Some thoughts:

- Should we use `which` or `which_global`? Is the only difference that `which("foo/bar")` will consider the current directory, and `which("foo/bar")` will not? Is there only a difference when the path contains `/` or `./`?
- Should we use `which` or `which_re`? I can actually see `which_re` being very useful, since you could do things like `which("g?make")` to find make or gmake, and regular expression special characters are vanishingly uncommon in binary names, so it doesn't seem like it would be annoying if you didn't want regular expressions.
- If we make it error if it can't find the executable, should we give it another name? `require(BIN)`?
- Should we provide both versions, like `which(BIN)` and `require(BIN)`? They both seem useful, so maybe?

0xzhzh:
> One thought is that perhaps this function should actually fail, as in, produce an error and terminate execution, if the binary does not exist?

The use cases I have in mind don't involve throwing an error. For instance, the example from #2109 suggests using `nala` and falling back to `apt`; I was thinking of using `rg` and falling back to `grep`.

If the user wants to give up, they could write, for instance:

```justfile
git := if which("git") == "" { error("git is not installed") } else { which("git") }
```

An alternative is to allow `which(cmd)` to fail, and then have an alternate version like `which(cmd, fallback)`.

> Tests which are just testing the functionality of which aren't necessary. In general, I try to avoid testing dependencies, and just assume they work (assuming that they're relatively popular, seem well maintained, etc).

That makes sense. Should I remove the tests I currently have?

> Should we use `which` or `which_global`? Is the only difference that `which("foo/bar")` will consider the current directory, and `which("foo/bar")` will not? Is there only a difference when the path contains `/` or `./`?

I think that's right, though I haven't tested it myself to confirm. I chose to use `which` because it seems to have strictly more functionality than `which_global` (i.e., it can resolve relative paths), but I'm not entirely sure if that's necessary.

> Should we use `which` or `which_re`? I can actually see `which_re` being very useful, since you could do things like `which("g?make")` to find make or gmake, and regular expression special characters are vanishingly uncommon in binary names, so it doesn't seem like it would be annoying if you didn't want regular expressions.

Yeah `which_re` does seem very useful, although one notable exception is `g++` and `clang++`. Perhaps both should be provided.

> If we make it error if it can't find the executable, should we give it another name? `require(BIN)`?
>
> Should we provide both versions, like `which(BIN)` and `require(BIN)`? They both seem useful, so maybe?

I like the word choice of `require`, because it makes it clear what happens in the command is missing. But then again, `which` makes it clear what the return value is when the command is present; it's unclear what `require` should return. i.e.,

```justfile
x := require("ls")

@test:
    echo {{x}}  # what does this print?
```

I guess it could just return the full path like `which`, but I'm not sure how much additional value that adds, at the expense of adding to the built-in functions' real estate. Another argument for not adding `require` now: we can always add it later.

Design-wise, I'm starting to lean toward having `which(cmd)` and `which(cmd, fallback_string)` (or maybe even an arbitrary number of fallbacks `which(cmd, fallback_cmd1, fallback_cmd2, ..., fallback_string)`), because it seems to be a balance of being concise and expressive for a variety of use cases. But ultimately it's up to you, and I'm happy to adjust the PR according to whatever design you think makes the most sense.

casey:
> The use cases I have in mind don't involve throwing an error. For instance, the example from #2109 suggests using `nala` and falling back to `apt`; I was thinking of using `rg` and falling back to `grep`.

Gotcha, that's good to know. In that case I think returning the empty string is ideal. I'm actually thinking about adding Python-style `and` and `or`, spelled `&&` and `||` in `just`, since the grammar won't permit using an identifier as an operator, `LHS || RHS` returns `LHS` if it is non-empty, and `RHS` if it is, so with `||` you could do:

```
git := which('git') || error(…)
```

Or, if you have a fallback:

```
grep := which('rg') || which('grep') || error(…)
```

If we added require, which I agree we can add later:

```
grep := which('rg') || require('grep')
```

(require would behave like `error()` if a binary wasn't found, so execution would stop with an error message, and it would not return a value)

> That makes sense. Should I remove the tests I currently have?

I think the tests you have are reasonable, and we should have at least one test, not to test the dependency, but test the function implementation, i.e. that we're calling the right dependency.

> I think that's right, though I haven't tested it myself to confirm. I chose to use `which` because it seems to have strictly more functionality than `which_global` (i.e., it can resolve relative paths), but I'm not entirely sure if that's necessary.

I think this is good, and we should use `which`, since users just not pass relative paths if they don't want to consider the working directory.

> Yeah `which_re` does seem very useful, although one notable exception is `g++` and `clang++`. Perhaps both should be provided.

Ooo, good call. Yah, `g++` and `clang++` are common enough that we definitely shouldn't make `which_re` the default. We can add `which_re` later, if needed. (And the fallback with a hypothetical `||` operator makes it easy to do `which("gmake") || which("make)`, which is probably better than `which("g?make")` since the precedence is not ambiguous.)

> I guess it could just return the full path like `which`, but I'm not sure how much additional value that adds, at the expense of adding to the built-in functions' real estate. Another argument for not adding `require` now: we can always add it later.

Yah, I agree. And `which(…) || error(…)` isn't bad, and let's the user specify an error message.

> Design-wise, I'm starting to lean toward having `which(cmd)` and `which(cmd, fallback_string)` (or maybe even an arbitrary number of fallbacks `which(cmd, fallback_cmd1, fallback_cmd2, ..., fallback_string)`), because it seems to be a balance of being concise and expressive for a variety of use cases. But ultimately it's up to you, and I'm happy to adjust the PR according to whatever design you think makes the most sense.

I think we should just do `which(cmd)` for now. Fallbacks could be done with `which(…) || which(…)`, and we could add the multi-argument version later, since it would be backwards compatible.

README.md
```diff
@@ -1597,6 +1597,22 @@ $ just
   name `key`, returning `default` if it is not present.
 - `env(key)`<sup>1.15.0</sup> — Alias for `env_var(key)`.
 - `env(key, default)`<sup>1.15.0</sup> — Alias for `env_var_or_default(key, default)`.
+- `which(exe)`<sup>1.37.0</sup> — Retrieves the full path of `exe` according
```
casey: This should be `<sup>master</sup>`. When I make a release, I replace all references to `master` with the new version.

src/function.rs
```diff
@@ -667,6 +668,16 @@ fn uuid(_context: Context) -> FunctionResult {
   Ok(uuid::Uuid::new_v4().to_string())
 }
 
+fn which_exec(_context: Context, s: &str) -> FunctionResult {
+  let path = which::which(s).unwrap_or_default();
```
casey: This will swallow the error from `which`. But I checked out the `which` crate, and it looks like it will actually swallow I/O errors, which (haha) isn't ideal. I opened an error on the crate: https://github.com/harryfei/which-rs/issues/100

I don't think this is a show-stopper, since at the very worst, we can swap out the implementation later.

tests/lib.rs
```diff
@@ -110,6 +110,7 @@ mod timestamps;
 mod undefined_variables;
 mod unexport;
 mod unstable;
+mod which_exec;
```
casey:
```suggestion
mod which;
```

casey (changes requested):
Random comments.

casey:
I took a look at the `which` crate, and I think we should actually just write our own. The `which` crate makes a number of choices that I'm not entirely comfortable with, like swallowing I/O errors, appending windows executable extensions, reading an environment variable called `PATHEXT` to control whether or not an extension is added, and the like. I would really prefer surface I/O errors, and not do anything fancy with extensions. Let me know if you are up for this!

tests/lib.rs
0xzhzh: Unfortunately this doesn't work because of the name conflict with the `which` crate, which is imported and used in this test suite:

https://github.com/casey/just/blob/4f31853835afcd8d511fccf7b2b6680f6a628446/tests/lib.rs#L28

0xzhzh:
> I took a look at the which crate, and I think we should actually just write our own. ... I would really prefer surface I/O errors, and not do anything fancy with extensions. Let me know if you are up for this!

Yeah, I agree. I'm happy to give it a shot, though I'm not familiar with how executables work on Windows, nor what the conventions are, so I'll start based on what the `which` crate does and then ask for your advice.

> I'm actually thinking about adding Python-style `and` and `or`, spelled `&&` and `||` in just, since the grammar won't permit using an identifier as an operator...

Yeah I saw you had mentioned that in another issue somewhere, and I really like that proposal. Besides, it would be nice to clarify what the semantics of conditional expressions and "Booleans" are---introducing those operators would act as a forcing function for that clarification.

0xzhzh:
I just pushed a draft implementation, but I haven't written any new tests (for which coverage is probably more important now). This implementation doesn't use the `which` crate anymore, but delegates to [`is_executable`](https://crates.io/crates/is_executable) to determine whether a path refers to an executable file. Looking forward to your thoughts and feedback.

Please note that `is_executable` _does_ still [swallow I/O errors](https://github.com/fitzgen/is_executable/blob/1b6dfa126abb19b4fac07a869c7bb6d0198c12ed/src/lib.rs#L40-L41). However, I think this actually makes some sense, because I don't expect `which` to complain if I have `PATH=/bin1:/bin2` and `/bin1/cmd` is an invalid path. In fact, neither `sh` nor `which` (on my system) seem to care whether `/bin1` is an unreadable path or a broken symlink, which I checked with the script folded in the folded details block.

<details>

The following script runs `/tmp/<path>/bin/cmd` as expected, without reporting any errors.

```sh
#!/usr/bin/env bash

set -e

tmpd="$(mktemp -d)"
pushd "$tmpd" >/dev/null

# Add an ordinary directory with an ordinary executable to PATH
mkdir bin
printf '#!/bin/sh\necho "this is expected"\n' > bin/cmd
chmod 755 bin/cmd
NEWPATH="$(pwd)/bin"

# An unreadable empty directory to PATH
mkdir unreadable-dir
chmod 000 unreadable-dir
NEWPATH="$(pwd)/unreadable-dir:$NEWPATH"

# Add a broken symlink to PATH
ln -s nowhere broken-dir
NEWPATH="$(pwd)/broken-dir:$NEWPATH"

# Add a directory with an unreadable file to PATH
mkdir unreadable
printf '#!/bin/sh\necho "this is unexpected"\n' > unreadable/cmd
chmod 000 unreadable/cmd
NEWPATH="$(pwd)/unreadable:$NEWPATH"

# Add a directory with a broken symlink to PATH
mkdir broken
ln -s nowhere broken/cmd
NEWPATH="$(pwd)/broken:$NEWPATH"

sh="$(which sh)"
which="$(sh -c 'which which')"

printf "Executing \`cmd'...\n\t"
PATH="$NEWPATH" "$sh" -c "cmd"
# executes ./bin/cmd

printf "Running \`which cmd'...\n\t"
PATH="$NEWPATH" "$sh" -c "$which cmd"
# prints the absolute path of ./bin/cmd

popd >/dev/null
rm -rf "$tmpd"
```

</details>

The shadowing use case you discuss in [this comment](https://github.com/harryfei/which-rs/issues/100#issuecomment-2451005509) is interesting, but I think it can be pretty hard to distinguish between an unmounted directory or unreadable file from unrelated directory in `PATH`. e.g., `$(HOME)/.cargo/bin` is going to be early in my `PATH` but I won't have `ls` in there.

That said, I think was still worthwhile to rewrite a simplified version of the `which` crate for `just`, for the following reasons:

- My implementation does _not_ swallow errors related to paths with invalid unicode, consistent with the rest of the builtin functions. It will also complain if `PATH` is not set.
- I noticed that the `which` crate performs [`tilde_expansion`](https://github.com/harryfei/which-rs/blob/1b5f7be8ca46e832d044adc2fec688e1cd0e9e53/src/finder.rs#L265-L283) for paths in `PATH`, which I don't think it should (and besides, it only expands `~` and does not handle `~username`). I'll file a separate issue there about this when I get the chance.
- When given a relative path, or when `PATH` contains a relative path, the `which` crate does not convert those to absolute paths. This is consistent with the behavior of the `which` behavior, but I personally think `just` is more robust resolving relative paths relative to `justfile_directory()` (and complaining if `justfile` doesn't have a parent directory).
- My implementation trades off a dependency on `which` for a dependency on `is_executable`, which is much smaller and was simpler for me to understand (ok I'm not entirely sure about what it's doing for Windows, but that's more so because I just don't know Windows at all).

Something else I wanted to respond to:

> reading an environment variable called `PATHEXT` to control whether or not an extension is added

I'm only just learning about this myself, but `PATHEXT` seems to be a standard thing: https://superuser.com/questions/1027078/what-is-the-default-value-of-the-pathext-environment-variable-for-windows. So we should probably respect that, and that is what `is_executable` does as well.

0xzhzh:
@casey  I finally got around to updating the tests to try out the internal `which` implementation.

I've only run this on my local macOS machine; I'm wondering if you could run those tests in CI to see if my implementation works on other OSes (Windows especially)?

src/function.rs
```diff
@@ -667,6 +669,61 @@ fn uuid(_context: Context) -> FunctionResult {
...
+
+  let cmd = PathBuf::from(s);
+
+  let path_var;
+  let candidates = match cmd.components().count() {
+    0 => Err("empty command string".to_string())?,
+    1 => {
+      // cmd is a regular command
+      path_var = env::var_os("PATH").ok_or("Environment variable `PATH` is not set")?;
+      Either::Left(env::split_paths(&path_var).map(|path| path.join(cmd.clone())))
```
casey: Can we avoid the dependency on `either` by returning a vec?

tests/lib.rs
casey: Ahh, gotcha

casey (changes requested):
Finally took another look at this, sorry for the delay!

I set the tests to run, and it looks like they all pass.

I think delegating to `is_executable` is fine, and that what you wrote about swallowing I/O errors is reasonable, given that it's what the external `which` command does.

Good to know that PATHEXT is a standard windows thing, I wasn't familiar with it at all.

I didn't do a super in-depth review, my only comment is about trying to avoid a dependency on `either`.

We could also consider adding a variant of `which()`, in a follow-up PR (doesn't have to be by you!) which fails if the binary isn't found. I think it's pretty common to need a bunch of commands available, and not want to run if they aren't present.

tests/lib.rs
0xzhzh: I'm happy to rename it from `which_exec` to something, e.g., `which_fn`, if you prefer. I chose `which_exec` to evoke something like "`which_exec`utable is `cmd`"

0xzhzh:
> Finally took another look at this, sorry for the delay!

No worries! Thanks for running the CI tests, I'm glad that everything works on Windows too.

> Good to know that PATHEXT is a standard windows thing, I wasn't familiar with it at all.

In all honesty I'm not familiar with this either... I will ask some of my friends from the Windows universe whether we are handling this correctly.

> We could also consider adding a variant of `which()`, in a follow-up PR (doesn't have to be by you!) which fails if the binary isn't found. I think it's pretty common to need a bunch of commands available, and not want to run if they aren't present.

Yeah, maybe something like `requires()`? I think that's a good idea, but it's also something that could be accomplished via #1059 if that ever gets worked on (though I understand that that's a whole bag of wormholes in and of itself). I'm happy to follow up this PR with one that implements `requires()`.

src/function.rs
```diff
@@ -667,6 +668,61 @@ fn uuid(_context: Context) -> FunctionResult {
...
+      // cmd contains a path separator, treat it as a path
+      vec![cmd]
+    }
+  };
+
+  for mut candidate in candidates {
+    if candidate.is_relative() {
+      // This candidate is a relative path, either because the user invoked `which("./rel/path")`,
+      // or because there was a relative path in `PATH`. Resolve it to an absolute path.
+      let cwd = context
```
casey: I think relative paths should be joined with the working directory, which you can get with:

```rust
context.evaluator.context.working_directory()
```

tests/lib.rs
casey: How about `which_function`?

tests/which_exec.rs
```diff
@@ -0,0 +1,170 @@
...
+#[test]
+fn skips_non_executable_files() {
+  let tmp = temptree! {
+    "hello.exe": "#!/usr/bin/env bash\necho hello\n",
+    "hi": "just some regular file",
+  }
+  .executable("hello.exe");
+
+  Test::new()
+    .justfile(r#"p := which("hi")"#)
```
casey: You can avoid needing `r#"…"#` strings if you use single quotes:

```suggestion
    .justfile("p := which('hi')")
```

tests/which_exec.rs
```diff
@@ -0,0 +1,170 @@
...
+      fs::set_permissions(file, perms).unwrap();
+    }
+
+    self
+  }
+}
+
+#[test]
+fn finds_executable() {
+  let tmp = temptree! {
```
casey: I try to avoid `temptree!` these days. I wrote it, but I think it's overcomplicated.

I try to use `Test::write`, to create individual files in the test tempdir.

Can this be expressed with something like:

```rust
  Test::new()
    .justfile(r#"p := which("hello.exe")"#)
    .write("subdir/hello.exe": "#!/usr/bin/env bash\necho hello\n")
    .make_executable("subdir/hello.exe")
    .env("PATH", "subdir")
    .args(["--evaluate", "p"])
    .stdout(format!("{}", tmp.path().join("hello.exe").display()))
    .run();
```

src/function.rs
```diff
@@ -661,6 +662,61 @@ fn uuid(_context: Context) -> FunctionResult {
   Ok(uuid::Uuid::new_v4().to_string())
 }
 
+fn which(context: Context, s: &str) -> FunctionResult {
```
casey, line 675: Let's handle the empty case up-front:

```rust
if s.is_empty() {
  return Err("empty command".into());
}
```

casey, line 675: What do you think about this version? I found the original logic a little hard to follow:

(Also included some comments which don't need to go into the final PR.)

```rust
fn which(context: Context, command: &str) -> FunctionResult {
  use std::path::Component;

  let command = Path::new(command);

  let relative = match command.components().next() {
    None => return Err("empty command".into()),
    // Any path that starts with `.` or `..` can't be joined to elements of `$PATH` and should be considered on its own. (Is this actually true? What about `C:foo` on windows? Is that a thing?
    Some(Component::CurDir) | Some(Component::ParentDir) => vec![command.into()],
    _ => {
      let paths = env::var_os("PATH").ok_or("`PATH` environment variable not set")?;

      env::split_paths(&paths)
        .map(|path| path.join(command))
        .collect()
    }
  };

  let working_directory = context.evaluator.context.working_directory();

  let absolute = relative
    .into_iter()
    // note that an path.join(absolute_path) winds up being absolute_path
    // lexiclean is hear to remove unnecessary `.` and `..`
    .map(|relative| working_directory.join(relative).lexiclean())
    .collect::<Vec<PathBuf>>();

  for candidate in absolute {
    if is_executable::is_executable(&candidate) {
      return candidate
        .to_str()
        .map(str::to_string)
        .ok_or_else(|| format!("Executable path not unicode: {}", candidate.display()));
    }
  }

  Ok(String::new())
}
```

casey (changes requested):
Nice! Left a bunch of comments, check them out.

src/function.rs
0xzhzh: I personally find my implementation easier to understand, but that might only be because I wrote it.

The main difference between our implementations seems to be how we determine whether to resolve via `PATH`. Mine checks `command.components.count()`: anything with more than one component is considered a relative path and skips `PATH` resolution. Meanwhile, yours checks `command.components.next()`: anything whose first element is `.`, `..`, or `/` is considered a relative path and skips `PATH` resolution.

Crucially, they differ on how they handle something like `which("dir/cmd")`, which has more than one component, but the first component is a `Component::Normal("dir")`. Thus, my implementation would consider it the same as `./dir/cmd` (i.e., resolved relative to CWD), while yours would resolve it using the `PATH` variable.

I wrote a small test, and (at least on macOS with `sh`, `bash`, and `zsh`) it seems like my implementation has the correct behavior in this instance:

```sh
#!/bin/sh

set -e
tmpdir="$(mktemp -d)"
cd "$tmpdir"

mkdir -p path/dir dir

printf "#!%s\necho resolved via PATH" "$SHELL" > path/dir/cmd
chmod +x path/dir/cmd

printf "#!%s\necho resolved via CWD" "$SHELL" > dir/cmd
chmod +x dir/cmd

PATH="$(realpath path)" dir/cmd   # prints 'resolved via CWD'

rm -rf "$tmpdir"
```

The other differences are:
- You use `Path::new(command)` rather than `PathBuf::from(command)`. I've adopted this change since it's more efficient (and also 
- You use `lexiclean()` to remove extraneous `..` or `.` components. However, at least the `which` implementation on my system (macOS) does not do this.
- You handle absolute paths (those beginning with `/`) implicitly, through the behavior of `Path::join()` (TIL about this behavior!). But I personally think it's clearer to handle this case explicitly.
- You make two passes through the candidates, one to prepend the working directory, and one to check whether it is an executable. I didn't do this to avoid making another copy.

0xzhzh: Just checking: do you still want to do this, now that I'm using `Path::new(s)` instead of `PathBuf::from(s)`? I noticed your implementation does not handle `s.is_empty()` early either.

0xzhzh:
@casey I made some changes according to your suggestions, and also added some test cases.

It might be nice to test the behavior of prefixed-paths on Windows but I don't know the platform well enough to say what the expected behavior should be.

src/function.rs
casey: Yah, you're right, in addition to your test, I looked at the [POSIX shell standard](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/V3_chap02.html_ (search for "command search and execution"), and it says that for any path with a, the `PATH` variable isn't consulted.

So I think `command.components.count()` is best. (Which, if we're doing that, it makes sense to use that to check for an empty command, and not `.is_empty()`)

GNU which does actually remove `..`:

```
$ gwhich ../just/bin/forbid
/Users/rodarmor/src/just/bin/forbid
```

So I'm in favor of using `lexiclean`, since it makes the returned paths easier to read, in case they return `..` or `.`.

I personally like using the relative path behavior to do joining, but I think that's kind of a wash, since it's a little weird.

casey (changes requested):
Looks good, see minor comments.

src/function.rs
0xzhzh: > So I think command.components.count() is best. (Which, if we're doing that, it makes sense to use that to check for an empty command, and not .is_empty())

Seems like this particular comment is resolved

0xzhzh: > GNU which does actually remove `..` [...]  So I'm in favor of using lexiclean, since it makes the returned paths easier to read, in case they return `..` or `.`.

I wasn't aware of that, thanks for finding that behavior! But if someone really wants to clean up the path, can't they do something like `clean(which("some/../funny/path"))`? Or even `canonicalize(which("some/../funny/path"))`?

The argument against cleaning the path is that doing so removes information that can no longer be recovered. I'm wary of over-normalization for weird edge cases like the example mentioned in the Rust [Path::components()](https://doc.rust-lang.org/std/path/struct.Path.html#method.components) docs. 

Admittedly, I can't personally foresee any scenario where this relative path information might actually be useful, so I'm strongly against calling `lexiclean()` here. But since the user can opt into normalization with other Just functions like `clean` and `canonicalize`, shouldn't we leave this to the user?

casey: I think it's probably okay, and `..` is likely to be relatively common in things passed to `which()`, since you're likely to want to go back to parent directories from the justfile, and it would be nice to get rid of them so the output reads better.

0xzhzh: Sounds good, I've added `lexiclean()`.

woutervh:
Looking forward to see this feaure released.

tests/test.rs
```diff
@@ -205,6 +205,31 @@ impl Test {
...
+    // Windows uses file extensions to determine whether a file is executable.
+    // Other systems don't care. To keep these tests cross-platform, just make
+    // sure all executables end with ".exe" suffix.
+    assert!(
+      file.extension() == Some("exe".as_ref()),
+      "executable file does not end with .exe: {}",
+      file.display()
+    );
+
+    println!("hi: {}", file.display());
```
casey: I think this is debug output?

casey (changes requested):
Looks good!

One small comment about what looks like left in debug output.

There's actually been a lot of conversation about how to represent `true` and `false` in just, which goes along with some progress about supporting array variables in `just`, in which case `false` would ideally be the empty array, and not the empty string. Conversation is all over the place, but see #2411 and #2458. (And probably more that I'm forgetting.)

If this happened, and we had a stable `which` command, then it would be a backwards incompatible change to change the no-binary-found return value to the empty list.

What do you think about changing this PR to add `require(…)`, which prints out an error message on failure, instead of `which(…)`. Almost exactly the same logic, but since `require(…)` doesn't return anything on failure, there aren't any issues relating to the value of `true` and `false` to think about.

Sorry to keep requesting changes, I really want to get this in, but I'm very paranoid about anything related to backwards compatibility and painting myself into a corner 😅

tests/test.rs
0xzhzh: Oopsies, sorry about that!

0xzhzh:
> What do you think about changing this PR to add require(…), which prints out an error message on failure, instead of which(…). Almost exactly the same logic, but since require(…) doesn't return anything on failure, there aren't any issues relating to the value of true and false to think about.

Sure, happy to add that. Just to make sure I understand: is your thinking here to make `require()` stable, and keep the return value of `which()` unstable until you finalize the design of string arrays as values?

> Sorry to keep requesting changes, I really want to get this in, but I'm very paranoid about anything related to backwards compatibility and painting myself into a corner 😅

No worries, I totally understand!

casey:
> Sure, happy to add that. Just to make sure I understand: is your thinking here to make `require()` stable, and keep the return value of `which()` unstable until you finalize the design of string arrays as values?

I was thinking we wouldn't add `which()`, and would only add `require()`, but adding `which()` as unstable is also an option, it's up to you. If we do, you'll need to add a new `UnstableFeature` enum variant.

laniakea64:
> adding `which()` as unstable is also an option,

Please do, and please make it (initially) return the empty string when program is not found.

A real example from one my actual justfiles showing the sort of construct `which()` would be useful for:
```
fdfind := `which fd 2>/dev/null || which fdfind`
```
could be written in pure `just` as
```
fdfind := which('fd') || require('fdfind')
```

casey:
I'm not at all opposed to adding `which()` as unstable, but we could consider doing it in a follow-up PR, to get this one in and keep it small. Either way is fine with me.

0xzhzh:
> I'm not at all opposed to adding which() as unstable, but we could consider doing it in a follow-up PR, to get this one in and keep it small. Either way is fine with me.

I made `which()` unstable, by adding to `unstable_features` during parsing (similar to what you do for the logical operators). Does that work?

nogweii:
I'm definitely excited to see `require()` added, as that would also resolve [my earlier request](https://github.com/casey/just/issues/2334).

Would it stop execution, or would it only log an error and keep moving on?

casey:
Nice! This looks good, although you'll need to add the `--unstable` argument or `set unstable` for the tests. Can you also add a test that doesn't add `--unstable`, and checks that we get the correct error message?

casey:
Also it looks like formatting failed, that should be fixable by running `cargo fmt`.

Blacksmoke16:
The new `require` function should also get some docs/tests.

Blacksmoke16:
Sorry for the double post, but I'm looking forward to this feature and think I found a bug?

I have a `justfile` like:

```just
KCOV := which('kcov')

@test:
    just _test-{{ if KCOV != '' { 'with-coverage' } else { 'without-coverage' } }}

@_test-with-coverage:
    echo 'with kcov'

@_test-without-coverage:
    echo 'no kcov'
```
And running this results in:
```sh
$ ~/dev/just/target/debug/just --unstable test
error: Call to unknown function `which`
 ——▶ justfile:1:9
  │
1 │ KCOV := which('kcov')
  │         ^^^^^
error: Recipe `test` failed on line 4 with exit code 1
```
However if I update the `test` task to be:
```just
@test:
    echo {{ if KCOV != '' { 'with-coverage' } else { 'without-coverage' } }}
```
I get:
```sh
$ ~/dev/just/target/debug/just --unstable test
with-coverage
```
So seems to be this one specific context that results in the error?

0xzhzh:
@Blacksmoke16 I think you need to invoke the recursive `just` invocation to use the `--unstable` flag. Alternatively, you can set the `JUST_UNSTABLE` environment, which should automatically propagate down (e.g., `JUST_UNSTABLE=1 ~/dev/just/target/debug/just --unstable test`).

0xzhzh:
@casey Sorry for missing those! I was in a rush when I made the `which()` function unstable and forgot to test it...

I just pushed some updates to fix the issues. Let me know if there's anything else to address!

0xzhzh:
@nogweii :
> Would it stop execution, or would it only log an error and keep moving on?

I ran the following test to see:

```console
$ cat justfile
before_cmd := `touch before_cmd.ran`
failure := `touch before_expr.ran` + require("does-not-exist") + `touch after_expr.ran`
after_cmd := `touch after_cmd.ran`

target:
  touch target.ran

$ rm -f *.ran ; ./just ; ls *.ran
error: Call to function `require` failed: could not find required executable: `does-not-exist`
 ——▶ justfile:2:38
  │
2 │ failure := `touch before_expr.ran` + require("does-not-exist") + `touch after_expr.ran`
  │                                      ^^^^^^^
 after_cmd.ran   before_cmd.ran   before_expr.ran
```

It seems like all variable-binding expressions evaluate, regardless of failures, whereas expressions stop evaluating as soon as a subexpression fails. I'm not sure if this execution order is specified somewhere, though I haven't really looked very hard. Perhaps @casey knows?

casey:
@0xzhzh The order of evaluation of assignments in a `justfile` is not linear. `just` actually evaluates assignments in alphabetical order, since it holds them in a sorted b-tree. However, if any assignments depend on other assignments, the order changes. E.g., if `foo` depends on `bar`, `bar` will be evaluated before `foo`.

If an assignment fails, then `just` should stop evaluating assignments. (Try naming them `a` `b` and `c`.)

The order of evaluation of arguments to `+` and `/` wasn't defined, but I just merged #2593, which forces them to evaluate from left to right.

casey (approved):
LGTM! Tweaked the docs a little, and moved them into their own section.

casey:
@0xzhzh Merged, thanks for sticking with this one!

0xzhzh:
@casey Thanks for your help in designing and implementing the feature too, I'm super happy to see this feature in Just now!

casey:
You bet!

Just released [1.39.0](https://github.com/casey/just/releases/tag/1.39.0), which includes the unstable `which` function and the stable `require` function.

PR #2441, opened 2024-10-27
Fix Nix Links

yunusey:
In the documentation, I realized the links were no longer valid.


casey:
Nice, thanks for the PR!

PR #2459, opened 2024-11-10
Update npm and pypi installation commands in README docs

gnpaone:


casey:
Looks good! But why the change from `pip` to `pipx`? Honestly I always forget the difference and I'm not sure why I used `pipx` in the first place.

gnpaone:
It works with pip, pipx, rye, poetry, uv etc. pip is python inbuilt/default package installer, pipx needs to be installed first separately before using pipx cli commands I guess. For beginners to keep it simple I changed it to pip 🙂. pipx is fine too if you want to retain it

casey:
I think I used `pipx` because on MacOS with Homebrew, `pip` doesn't work, and it's a common configuration. I changed it back to `pipx`. Otherwise looks good!

gnpaone:
Okay 🙂

PR #2460, opened 2024-11-10
Suppress recipe not found error code

### LINKED ISSUE #2275: Command line flag to exit without error if recipe not found
In some cases it's helpful to run `just` so that it silently exists with code 0 if requested recipe is not found.  Possible usecase: running same recipe in multiple directories (e.g. with GNU parallel) when in some of them the recipe is not defined, in which case we just want to skip the dir.

It is currently possible to do with NPM scripts, for example `npm run --ws --if-present <script>` will run *script* in every workspace where script is defined.

R3ZV:
Closes #2275. It adds a new verbosity version which is used to convert the error code of an `UnknownRecipe` to 0, due to it making an early return it will also suppress any error messages. This behaviour is the same as the one `npm --if-present` has as it was referenced by the issue author.

I choose to add another verbosity variant due to not wanting to add further fields in `Config`.  I thought it fits the verbosity config since it affects the output of the program.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -110,7 +110,7 @@ jobs:
       shell: bash
 
     - name: Publish Archive
-      uses: softprops/action-gh-release@v2.0.9
+      uses: softprops/action-gh-release@v2.1.0
       if: ${{ startsWith(github.ref, 'refs/tags/') }}
       with:
         draft: false
@@ -120,7 +120,7 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Publish Changelog
-      uses: softprops/action-gh-release@v2.0.9
+      uses: softprops/action-gh-release@v2.1.0
       if: >-
         ${{
           startsWith(github.ref, 'refs/tags/')
@@ -157,7 +157,7 @@ jobs:
         shasum -a 256 * > ../SHA256SUMS
 
     - name: Publish Checksums
-      uses: softprops/action-gh-release@v2.0.9
+      uses: softprops/action-gh-release@v2.1.0
       with:
         draft: false
         files: SHA256SUMS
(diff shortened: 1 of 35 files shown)


laniakea64:
The name `--recipe-exists` is neither descriptive nor intuitive for this functionality.  Could this option be named `--if-present`, same as the `npm` option?

`--if-present` is also more natural considering how this option could mesh with optional imports/modules.

R3ZV:
Fair enough, I will make the according naming changes, are there any changes that should be made, regarding either the implementation or the tests?

casey:
Verbosity levels should only affect what's printed and not what error codes are returned, so I think this should be a flag. I'm not entirely satisfied with `--if-present`, but I could be swayed.

Other ideas:

- `--ignore-missing`
- `--optional`
- `--allow-missing`

`--optional` is nice because it's short, and `--allow-missing` is nice because you can imagine other flags of the form `--allow-X`, to allow different errors.

R3ZV:
I myself am a fan of either `--ignore-missing` or `--allow-missing`, I do have one question, by "this should be a flag" you mean to have another field in `struct Config`, or are there other mechanisms in the codebase? If it is the former perhaps I can add a field `allow_missing` to `Config`.

casey:
Let's go with `--allow-missing`. Yup, it should be flag on `Config`.

R3ZV:
Changed everything to match the feedback, if there is anything else you would like me to change, please let me know!

Perhaps the description could be more specific than `Suppress error code`, maybe something like `Suppresses 'UnknownRecipe' error code`.

casey:
I tweaked this a bit. The main change was to move the check for unknown recipes into `Subcommand::run`, which makes it only apply when running a recipe, and not other subcommands, like `--show` which still return an error if `--allow-missing` is supplied and the recipe is missing. I also extended it to missing submodules, so `just --allow-missing foo::bar` where there is no `foo` module is okay.

PR #2465, opened 2024-11-15
Replace `derivative` with `derive-where`

laniakea64:
Running `cargo audit` on `just` repo gives the following warnings -
```
Crate:     ansi_term
Version:   0.12.1
Warning:   unmaintained
Title:     ansi_term is Unmaintained
Date:      2021-08-18
ID:        RUSTSEC-2021-0139
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0139
Dependency tree:
ansi_term 0.12.1
└── just 1.36.0

Crate:     derivative
Version:   2.2.0
Warning:   unmaintained
Title:     `derivative` is unmaintained; consider using an alternative
Date:      2024-06-26
ID:        RUSTSEC-2024-0388
URL:       https://rustsec.org/advisories/RUSTSEC-2024-0388
Dependency tree:
derivative 2.2.0
└── just 1.36.0

warning: 2 allowed warnings found
```

This PR fixes one of these two warnings by replacing `derivative` with `derive-where`, which is [how `rustc` handled this](https://github.com/rust-lang/rust/pull/127042).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -237,7 +237,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn 2.0.79",
+ "syn",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


casey:
Nice, great PR! If you look at the `Cargo.lock` changes, you can also see that this gets rid of a dependency on syn v1, since `derive_where` uses syn v2. So this should speed up builds a bit too, since syn is a heavy dependency.

casey:
Just measured it and build time went down from ~34.5s to ~32.5 seconds.

Issue #2252, opened 2024-07-14
Module improvement tracking issue


casey:
This issue tracks improvements and missing features of submodules:

- Recipes can depend on recipes in parent or sibling modules
- Variables defined in one module can be referenced from another
- #2442
- #2271
- Allow submodules to load `.env` files
- #2119
- Make all errors, not just compilation errors, refer to module source file path as appropriate
- Make `just --show` work with modules
- #2629

This thread should probably be used for general discussion. If you're interested in one of the above features, mention it in this thread and I'll create a dedicated issue for it.

vasdee:
I am definitely interested in seeing modules fleshed out a bit more. I'm coming from being a fairly heavy make user where our team could organize make 'stubs' and include them on a per-repo basis. 

Something like this was nice and handy in Make-land, and is definitely _nearly_ there with Just

`Makefile`
```
-include .env
-include make.d/*.make
```

`make.d/docker.make`
```
docker/build:
   docker build ...

docker/login:
   docker login ....
```

The recipe naming with the `/` gave us faux-namespacing, which worked fairly well, but is arguably better implemented with Just modules. The one thing that keeps me from using them in this Make style way, is that variables from main/root Justfile don't cascade into the sub-modules 

This has made it difficult to fully embrace modules, so we resort to using imports in Just and doing a more 'restrictive' faux-namespacing in the recipe names. 

I would therefore +1 all the following, as I think they all contribute 

* Recipes defined in one module can be depended on from another
* Variables defined in one module can be referenced from another
* Allow creating inline modules that are declared and defined in a single module
* Add a setting which allows a parent module to propagate settings to submodules
* Consider adding a reverse setting, which allows a child module to inherit settings from its parent

As an aside, I think I'd be perfect happy as well if recipe names could include some special characters that allow the 'faux' namespacing I mention, `/` is ideal ( or even `\`, but even a `.`  would suffice. Then I could just use imports

prasannavl:
Adding this one to the list: https://github.com/casey/just/issues/2271

expelledboy:
We have nested just mods that goes:

```
./Justfile
./iac/mod.just
./iac/docker/mod.just
./iac/docker/stack/mod.just
```

Which allows us to call `just iac docker stack deploy`

However from `./iac/docker/stack/mod.just` I want to call a task defined in `./iac/docker/stack/mod.just` and not have to use the fully qualified `just iac docker stack <local-task>` because this module is used in multiple projects with different file hierarchies.

[...]

PR #2466, opened 2024-11-16
Add source information as path of the information

psibi:
Motivation: To add support for module integration in justl which is an Emacs extension for driving justfiles within Emacs.

This would allow me to easily discover the file locations of different just modules and allow the ability to individually open justl buffers for them within the editor.

More info here:
- https://github.com/casey/just/issues/2252#issuecomment-2474171211
- https://github.com/psibi/justl.el/issues/57


casey:
Seems reasonable to me! Would it be better to have relative paths or absolute paths?

psibi:
> Would it be better to have relative paths or absolute paths?

Yeah, I think absolute paths would be more convenient. And it was trivial to implement it. Let me see how to make the tests pass using that.

psibi:
@casey  This is ready for review now.

src/analyzer.rs
```diff
@@ -182,6 +182,7 @@ impl<'run, 'src> Analyzer<'run, 'src> {
       unstable_features.insert(UnstableFeature::ScriptInterpreterSetting);
     }
 
+    let source = root.canonicalize().unwrap();
```
casey: Canonicalize produces weird paths. On Unix, it resolves symlinks, which is often not what you want, and on Windows it produces paths prefixed with `\\?\` which is not very nice.

Canonicalize is also a system call, which means it can fail with an I/O error, so we can't unwrap here.

If we need to clean the path, we should use `.lexiclean()` instead of canonicalize, which cannot fail and doesn't produce weird paths.

tests/json.rs
```diff
@@ -1,22 +1,37 @@
 use super::*;
 
-fn case(justfile: &str, value: Value) {
+fn case<F: Fn(&Path) -> Value>(justfile: &str, value: F) {
```
casey: The JSON tests are pretty horrible. I changed the `case` function so that you pass in a struct which should serialize to the output of `--dump`.

You can take advantage of this by putting the relative path in the `source` field of the passed in `Module`, and then, in `case`, prepending the path to the temporary directory, which will make them match.

casey (changes requested):
See comments!

src/analyzer.rs
psibi: I have pushed a fix, but the tests in MacOS fails because I think of symbolic link. Do you know how to handle it ?

tests/json.rs
psibi: Yeah, I agree. :-)

src/analyzer.rs
psibi: Okay, I have pushed a fix where I have to use `canonicalize` only for the tests to make it pass. Let me know what you think.

casey (approved):
LGTM. `source` is now absolute, instead of relative, and it's used in some other places which were expecting a relative path. I fixed those up and I think they're okay.

psibi:
Thank you!

PR #2468, opened 2024-11-16
docs: note that set shell is not used for [script]

iloveitaly:
this was surprising functionality for me.

IMHO I think inheriting `shell` if `script-interpreter` is not set might be better


casey:
Agreed it's confusing!

The reason for not using the value of `set shell` is that the argument to `set shell` is a string, whereas the argument to `set script-interpreter` is a path to a script. So `set shell` needs `-c`, and `set script-interpreter` doesn't.

iloveitaly:
Yeah, definitely makes sense once you understand  that `[script]` essentially generates an executable file with the contents of the recipe.

PR #2482, opened 2024-11-26
feat: github groups

GauBen:
Work in progress: bringing https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#grouping-log-lines to just!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,6 +13,14 @@ defaults:
     shell: bash
 
 jobs:
+  run-dummy:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - uses: Swatinem/rust-cache@v2
+    - run: cargo run -- dummy
+
   lint:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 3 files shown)


GauBen:
It's me again @casey, can you please allow workflows to run on this PR?

GauBen:
Thank you! What do you think of this https://github.com/casey/just/actions/runs/12038853590/job/33566940915#step:4:209 ?

casey:
Hmm, interesting. I think that a group per command is probably too much. If a recipe has many small commands, then putting each in a group is probably going to be too noisy.

Can you change it to be per-recipe so we can see what it looks like?

A risk with a feature like this is that we won't be able to do it well automatically, and we should leave it to the `justfile` author, since they are in a better position to know how to group output.

GauBen:
I updated this MR with per-recipe grouping, can you please run the workflow?

casey:
Just re-ran it. The output still isn't great, for example, the nested group isn't within the outer group, and the commands in the outer group after the nested group don't appear in the right group.

Also, recipe names are sometimes not the most explicit, and so aren't great for log messages.

I'm thinking that this is something which is probably best handled by the `justfile` author themselves, and we won't be able to do it very well without understanding how they're using `just`.

GauBen:
Yep, that's not a good fit for a native `just` feature, but glad we tried!

PR #2490, opened 2024-11-30
Add "not matching regex" conditional

### LINKED ISSUE #2416: Conditional operator for NOT match regex?
I have run into a number of cases where a "not matching regex" condition would be useful, i.e. the opposite of `=~`.  Latest example is [here](https://github.com/NoahTheDuke/vim-just/pull/114/files#diff-dccf001585af8f37bb6b80fa80eb166ba6749533115939348bc40d59601e965fR20): without a "not matching regex" conditional operator, this check cannot use `assert()` which was specifically designed for cases like this.

Does `just` have such operator?  If not, could it please be added?

BTW I'm not finding the documentation of the `assert()` function-like expression in the README?

laniakea64:
Closes https://github.com/casey/just/issues/2416


casey:
The `Unspecified` token kind isn't really intended to be surfaced to users, it's for constructing an compile error, which normally requires a token, when there is none.

See the test I added in the last commit, which fails with something like:

```
error: Expected '&&', '!=', '!~', '||', '==', '=~', '+', or '/', but found unspecified
 ——▶ justfile:1:12
  │
1 │ x := if '' !! '' { '' } else { '' }
  │            ^
```

I think the signature of lex_choices should be changed to:

```
  fn lex_choices(
    &mut self,
    first: char,
    choices: &[(char, TokenKind)],
    otherwise: Option<TokenKind>,
  ) -> CompileResult<'src> {
```

And if `otherwise` is `None`, it generates an `UnexpectedCharacter` error.

laniakea64:
Done and fixed the failing test.  Thanks for the hint how to make that error message more sensible :+1:

casey (approved):
Nice, LGTM!

laniakea64:
@casey Actually there is an issue with this approach to the error message:

With the following invalid justfile content:
```
a := if x !
```
If this is saved in a justfile **without trailing newline**, so that the `!` is the very last character in the file, `just` will produce this error -
```
$ just -l
error: Internal error, this may indicate a bug in just: Lexer advanced past end of text
consider filing an issue: https://github.com/casey/just/issues/new
 ——▶ justfile:1:12
  │
1 │ a := if x !
  │            ^
```

But this is not a bug in `just`, advancing one character past the end of text is the expected behavior in this case.

Adding the trailing newline to the justfile results in the expected error message.

An alternative approach I considered when making this PR was adding a `Bang` token, to enable lexing `!=` and `!~` by the exact same approach as the pre-existing lexing of `==` and `=~`, so that the error message would look like that one:
```
$ just -l
error: Expected '&&', '!=', '!~', '||', '==', '=~', '(', '+', or '/', but found '='
 ——▶ justfile:1:11
  │
1 │ a := if x =
  │           ^
```

I didn't go that route because `!` isn't valid syntax anywhere in `just`, and I had doubts about adding a token that is never valid syntax.

Not sure what would be best fix this (and not sure if I'll have time to make another PR myself :frowning_face: )

casey:
Nice catch! This is an easy fix, there's an `UnexpectedEndOfToken` error, for just this eventuality. Fixed in #2520.

laniakea64:
> Nice catch! This is an easy fix, there's an `UnexpectedEndOfToken` error, for just this eventuality. Fixed in #2520.

Much better, thanks!

Seems to be some extra backticks in the error message? -
```
error: Expected character ``=` or `~`` but found end-of-file
 ——▶ justfile:1:12
  │
1 │ a := if x !
  │            ^

```

casey:
Good catch, fixed in #2522!

PR #2507, opened 2024-12-07
Add "file_content(filename)" function

begoon:
This PR adds `file_content()` function, which loads a file content as a string.

In task runners like `just`, one often needs to load from the file system (like "VERSION.txt" etc.).


casey:
Seems reasonable to me! Changed the name to `read_to_string` to be more explicit.

PR #2513, opened 2024-12-10
docs: multiline variable clarification

iloveitaly:
this was slightly confusing to me

### DIFF
--- a/README.md
+++ b/README.md
@@ -1466,7 +1466,8 @@ escapes := "\t\n\r\"\\"
 ```
 
 Indented versions of both single- and double-quoted strings, delimited by
-triple single- or double-quotes, are supported. Indented string lines are
+triple single- or double-quotes, are supported. Recipe variables cannot
+be interpolated in a triple-quoted string. Indented string lines are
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1466,7 +1466,8 @@ escapes := "\t\n\r\"\\"
 ```
 
 Indented versions of both single- and double-quoted strings, delimited by
-triple single- or double-quotes, are supported. Indented string lines are
+triple single- or double-quotes, are supported. Recipe variables cannot
+be interpolated in a triple-quoted string. Indented string lines are
```
laniakea64, line 1470: `{{...}}` is not interpolated in _any_ `just` string (see https://github.com/casey/just/issues/11 ).  This is not specific to triple-quoted strings or "recipe variables", so being specific here is only confusing.

casey:
I think I agree with @laniakea64, interpolation isn't possible in any kind of string, so calling out triple-quoted strings in particular is confusing. I opened #2515 to add a more general note about all strings.

iloveitaly:
Awesome! That definitely adds some clarity.

PR #2514, opened 2024-12-10
Enable fish to complete recipes in modules

senekor:
Hi, thanks for making just. One of the big reasons for me to use it is because of the completions. However, recipes in modules aren't completed at all (in fish), so I basically didn't use that feature.

I ended up writing the completions for that myself. I have them checked into my dotfiles now, so merging this PR isn't urgent for me. I thought I'd share it in case it is deemed useful for others.

It's quite a bit more verbose, but I didn't find a way to make it more concise without also making it more obtuse and hard to understand.

I should also say that this logic parses the output of:
```
just --list \
     --list-heading='' \
     --list-prefix='>' \
     --list-submodules
```
I might have missed some edge cases of that output format if it's related to a feature I personally don't use. For example, I haven't used groups so far, I only noticed my lack of handling `[group_name]` headers when I tried out my completions on the `just` repo itself. Recipe arguments and descriptions should be handled the same way as the existing completions.


casey:
Nice! Can you use `just --summary` instead, which doesn't have any formatting?

senekor:
Wow, that's so much simpler! 👍

This approach does lose hints for the recipe arguments and description, but I don't think it's a big loss.

casey:
Hmm, it's unfortunate to lose recipe arguments and descriptions, but I do think overall it's an improvement to be able to complete recipes in modules.

PR #2528, opened 2024-12-13
Add env_var_with_value() function

begoon:
Explained in https://github.com/casey/just/issues/2529


casey:
Closing, see https://github.com/casey/just/issues/2529#issuecomment-2558299554

PR #2531, opened 2024-12-19
Add experimental [no-shell] attribute

casey:
This PR adds an experimental `[no-shell]` attribute.

When the `[no-shell]` attribute, linewise recipe lines are *not* run with `sh` (or whatever the shell is set to) but instead is run with a very simple command parser and interpreter.

The goal is to provide the ability to run commands with no external dependencies, other than the commands themselves.

This came up in https://github.com/astral-sh/uv/issues/5903 when discussing the possibility of using `just` as a command runner for `uv`. Because `just` relies on a shell, `sh` by default, to run commands, it can't be used by projects which must be cross platform with no other dependencies.

This PR, which isn't really mergable at the moment, only supports word splitting on whitespace and `{{…}}` interpolations.

For this to be minimally useful, I think it also needs to support:

- Single and double quoted strings with the usual semantics
- Environment variable access with `$VAR`
- Selectively joining the result of `{{…}}` interpolations with adjacent text, depending on whether it's whitespace separated or not
- Making unsupported shell syntax an error, instead of passing unwanted arguments (so `echo hello > foo` is an error, instead of calling `echo` with `hello`, `>`, and `foo`)

I've been hoping for a long time that a Rust-only bourne shell would be written, but that doesn't appear to be happening, thus this PR.

I'm not sure if this PR is worth it, to be honest. There are a ton of ways to get a working `sh`, in almost every environment. If `sh` is not available, any other shell or interpreter can be used. So this covers a pretty niche use-case. Also, I'm not sure `just` is a good fit for `uv`, even if this feature was available.

Also, this feature would have to be unstable for a long time, until we were sure that we had added everything that we wanted to the shell language.

### DIFF
--- a/src/attribute.rs
+++ b/src/attribute.rs
@@ -18,6 +18,7 @@ pub(crate) enum Attribute<'src> {
   NoCd,
   NoExitMessage,
   NoQuiet,
+  NoShell,
   Openbsd,
   PositionalArguments,
   Private,
@@ -37,6 +38,7 @@ impl AttributeDiscriminant {
       | Self::NoCd
       | Self::NoExitMessage
       | Self::NoQuiet
+      | Self::NoShell
       | Self::Openbsd
       | Self::PositionalArguments
       | Self::Private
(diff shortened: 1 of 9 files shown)


casey:
Closing, since this is quite stale at this point.

PR #2536, opened 2024-12-20
Show full example of getting XDG user directory

laniakea64:
Per [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/latest/), getting an XDG user directory involves a bit more than reading an environment variable:

> All paths set in these environment variables must be absolute. If an implementation encounters a relative path in any of these variables it should consider the path invalid and ignore it.
> ...
> `$XDG_CONFIG_HOME` defines the base directory relative to which user-specific configuration files should be stored. If `$XDG_CONFIG_HOME` is either not set or empty, a default equal to `$HOME/.config` should be used.

This PR expands the example to fully show how to get an XDG user directory in a justfile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1927,8 +1927,14 @@ details.
 - `home_directory()` - The user's home directory.
 
 If you would like to use XDG base directories on all platforms you can use the
-`env(…)` function with the appropriate environment variable, e.g.,
-`env('XDG_CACHE_HOME')`.
+`env(…)` function with the appropriate environment variable and fallback, e.g.
+```just
+xdg_config_dir := if env('XDG_CONFIG_HOME', '') =~ '^/' {
+  env('XDG_CONFIG_HOME')
+} else {
+  home_directory() / '.config'
+}
+```
 
 ### Constants
 
--- a/README.md
+++ b/README.md
@@ -1927,7 +1927,11 @@ details.
 - `home_directory()` - The user's home directory.
 
 If you would like to use XDG base directories on all platforms you can use the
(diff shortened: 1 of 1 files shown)


casey:
Interesting!

This does at least slightly make the case that getting the XDG dirs correctly is painful, and should be better supported by `just`.

I wonder if this is something that users should actually care about. I.e., if the XDG_CONFIG_HOME isn't absolute, the user should really fix that, and the `justfile` shouldn't try to compensate, even if it is acting according to the spec.

Because if so you could do:

```just
xdg_config_dir := env('XDG_CONFIG_HOME', home_directory() / '.config')
```

Which is much nicer.

laniakea64:
> I wonder if this is something that users should actually care about. I.e., if the XDG_CONFIG_HOME isn't absolute, the user should really fix that, and the `justfile` shouldn't try to compensate, even if it is acting according to the spec.

The justfile is likely seeking to match other XDG-directory-using programs.

Testing a few applications in Xubuntu 22.04, they don't seem to care about relative paths.  However, they do care about the env vars being set to empty values: most apps ignore the empty env var and use the default, a few apps proceed as if there is no user directory available.

So even when a justfile wants to be relaxed about the XDG specification, getting an XDG user directory still involves more than a single `env()` call for the most likely case:
```
xdg_config_dir := if env('XDG_CONFIG_HOME', '') != '' {
  env('XDG_CONFIG_HOME')
} else {
  home_directory() / '.config'
}
```

casey:
Gotcha. And that isn't much shorter than the fully correct version which uses `=~ '^/'`, so we might as well do that.

PR #2547, opened 2024-12-23
Add `?` guard sigil

### LINKED ISSUE #2544: Guard sigil
We have `@` and `-` sigils which control linewise recipe execution. I thought that we could consider having a "guard" sigil which indicates that an error on a particular line should terminate the execution of the current recipe, but not the whole run.

For example, to make a recipe only execute if the environment variable `FOO` is set to `yes`:

```
foo:
  ?[[ $FOO == yes]]
  # the rest of the recipe
```

This is backwards incompatible, so would need a setting to opt-in to the new behavior. (Although it's unlikely to break most `justfile`, `?` at the beginning of the line is unlikely in `sh` and derivatives, and I think most scripting languages that people are likely to use.)

casey:
Closes #2544.

This is new weird syntax, which I am quite fond of, but I know other people may not like semantic line noise as much as I do, so going to let this sit for a few days to see what people think before merging.


liquidaty:
I like it!

torinthiel:
How does that interact with shebang recepies?

casey:
`?` couldn't be used in shebang recipes, since they aren't evaluated line-by-line. However, shebang recipes, since they are normal scripts, would have better facilities for stopping the script without an error if something fails, so they don't really need it.

PR #2549, opened 2024-12-26
Add `quiet` setting in settings table + fix typos

unennhexium:
Hi, Casey. I just finished reading the README.md and along the way I probably discovered a couple of typos and a one missing setting in the table. Please review my changes, maybe some of this is actually correct. Please let me know if I made wrong corrections or just reject the request.


casey:
Nice fixes, thank you!

PR #2556, opened 2024-12-31
refactor: changelog rev

bingryan:
## Description

`just --changelog` should prioritize displaying the latest changelog.

## preview

image


casey:
I'm on the fence about this! On the one hand it's nice, on the other hand, you can always do `just --changelog | less` to be able to browse it easily, and if we reverse the order, `just --changelog | less` is less useful, since it will start at the top with the oldest changes.

Issue #2458, opened 2024-11-08
Make all values lists of strings, instead of strings


casey:
Just values are all strings. This is convenient, because it means that Just is effectively statically typed. It is impossible to provoke a type error.

However, it means that many desirable things are impossible:

- Can't have lists.
- No way to represent a `*arg`. Currently `arg` is a space-separated string, but users may want to do something with each entry in the list, and re-splitting is error-prone.
- Representation of booleans is tricky. Some functions return `true/false`, but these are also valid strings.

We could add a type system, see #528. But another interesting option would be to change our single type from strings to lists of strings. This is done in [rc](https://www.scs.stanford.edu/nyu/04fa/sched/readings/rc.pdf), Plan 9s shell.

This seems crazy! Everything is a list? However, consider this: The shell, and Just, benefit from enormous simplicity because everything is a string. No type errors! No type system! We could abandon this, and add a type system, but another path is to make the one type we do have more powerful.

All current values, instead of being strings, would be lists containing single strings. When used in interpolations, or as arguments to `/` and `+`, an list of a single string would behave in exactly the same way as a string does now, so this would be a backwards compatible change.

This would allow the following nice things:

- The only `false` value would be the empty list, and any non-empty list is true. `['']` is true, so it would be possible to provide the empty string as a value, and still allow a fallback with `value || fallback`.
- `make` has a ton of functions which treat a string as a space-separated list. These are useful, but error prone, so I haven't wanted to add these. We could add useful, non-error-prone versions of these.
- `+` could append an extension to every element of a list: `["hello", "goodbye"] + '.c'` -> `["hello.c", "goodbye".c]`
- Ditto for `/`: `["/", "/usr"] / "bin"` -> `["/bin", "/usr/bin"]`
- `false` could be represented by the empty list, and all other values would be interpreted as true. This `['']` would be true, allowing the user to provide the empty string for a value.
- `for` which evaluates to a list: `for a in var { a + ".c" }`

The main downside, I think, is that it is very unfamiliar and counter-intuitive. Users are used to shell-like languages, including `make` and `just`, where everything is a string. Everything is a list of strings is very weird. `rc` implements and makes a compelling case for lists of lists being the single type, but it's unfamiliar and mostly forgotten.

Open questions:

- Should we call them lists or arrays?
[...]

Issue #1988, opened 2024-04-03
Question: How to forward *args?


gabyx:
When I have 

```justfile

run *args:
   echo "$1, $2,..."

do *args: (run args)
```

How can I forward arguments properly such that $2 is `b` in  `just do a b c`?

casey:
Unfortunately this currently isn't possible. `args` will be passed to `run` as a single space-separated argument.

The best way to support this would by some kind of splat syntax:

```just
run *args:
   echo "$1, $2,..."

do *args: (run *args)
```

casey:
Whoops, didn't mean to close.

drmacdon:
A work-around today might be to call back into self. 
```
set positional-arguments

run *args:
   echo "$1, $2,..."

@do *args:
   {{just_executable()}} -f {{justfile()}} run "$@"
```

latk:
> Unfortunately this currently isn't possible. `args` will be passed to `run` as a single space-separated argument.

This currently makes it impossible to use more complicated arguments to pass shell arguments safely to a dependency (i.e., without extra word-splitting or interpolation).

I understand why arguments must be immediately stringified, because the Just data model consists of just `String`. Changing that would be tedious, but in principle possible. I think the trick to do this in a backwards-compatible way would be to implement something like Perl's "dualvar", where a value is both a string value and possibly another value at the same time.

**I'd be interested in prototyping that approach if there will be interest in merging it.**

Sketch of such a data structure that could replace `String` in bindings:

```rust
struct Value {
  /// String representation of the value.
  stringy: Box<str>,
  /// Optionally, array representation of this value.
  items: Option<Box[Box<str>]>,
}

impl Value {
  pub fn new(s: String) -> Self {
    Self { stringy: s, items: None }
  }
  pub fn from_vec(items: Vec<Box<str>>) -> Self {
    Self { stringy: items.join(" "), items: Some(items.into()) }
  }
}

impl Deref<str> for Value { ... } // convenience compatibility with String
```

Just's values are immutable, so the more compact `Box<str>` can be used instead of `String`, and `Box<[...]>` instead of Vec.

Single-value strings `s` would subsequently be represented with `items: None`, whereas multi-value variadics could retain the correctly split `items`, while still being compatible with string-like use.

Once that is done, a dependency splat syntax `caller *args: (callee *args)` should be implementable. Whereas the current evaluator produces a single value, a list-context evaluator might be necessary. Something like:

```rust
pub(crate) fn evaluate_expression(
[...]

Issue #558, opened 2019-12-07
Dispatching variadic arguments to dependencies with splats


casey:
It should be possible, when a variadic parameter receives multiple arguments, to call a dependency for each of those arguments:

This could look like:

```just
foo *args: (bar *args)
@bar arg:
  echo bar: {{arg}}
```

Which would have the effect of, for each value in `args`, call `bar` with that value:

```
$ just a b c d
bar: a
bar: b
bar: c
bar: d
```

redsun82:
👋 I stumbled on this through https://github.com/casey/just/issues/1988 and https://github.com/casey/just/pull/2567. I just wanted to drop the suggestion that to me `(bar *ARGS)` should rather indicate the forwarding feature described in https://github.com/casey/just/issues/1988 rather than dispatching multiple dependencies as proposed here. But this seems like a valid use case as well: I think an alternative intuitive syntax could be
```just
foo *args: *(bar args)
```

casey:
@redsun82 Good catch, agreed!

casey:
Implemented in #3379 as:

```just
foo *args: *(bar *args)
bar arg:
```

So you must have a star outside the group, as well as on the actual argument you want to map over.

PR #2567, opened 2025-01-11
Implement `Val` data model, a list of strings

latk:
In support of:

* https://github.com/casey/just/issues/2458
* https://github.com/casey/just/issues/1988

A `Val` is a list of strings, but at each point of usage we must decide whether to view it as a singular joined string or as a list of its parts. Previously, Just values were single strings, so in most places we have to invoke `to_joined()` in order to maintain compatibility. In particular, recipes, functions, and operators like `+` or `/` operate solely on strings. This
includes logical operators like `&&`, which continue to be defined on strings. That means, the values `[]` and `['']` are currently completely equivalent.

So far, this is a purely internal change without externally visible effects. Only the `Bindings`/`Scope`/`Evaluator` had API changes.

No new syntax is implemented. However, in expectation of expressions that build lists, a new `evaluate_list_expression() -> Vec<String>` method is introduced that could be used to implement splat or generator expressions. It is already referenced wherever we have lists of arguments, e.g. variadic functions like `shell()` and dependency arguments. But because singular expressions are equivalent to a joined string, this is an unobservable detail for now.

For experimenting with lists of strings, variadic recipe parameters like `*args` now produce a multi-part Val, and default to an empty list (not a list with an empty string). Because all operators use `to_joined()`, this is an unobservable implementation detail. However, if any operator becomes list-aware, then this detail should be reverted, or moved behind an unstable setting.

For better understanding of the current behavior, I added a bunch of tests. These will help detect regressions if functions or operators become list-aware. No existing tests had to be touched.

Next steps: This change is just the foundation for other work, but some ideas are mutually exclusive. Relevant aspects:

* list syntax in https://github.com/casey/just/issues/2458
* list aware operators in https://github.com/casey/just/issues/2458
* lossless forwarding of variadics: https://github.com/casey/just/issues/1988
* invoking dependencies multiple times: https://github.com/casey/just/issues/558

The preparatory work like `evaluate_list_expression()` is biased towards implementing a splat operator that would enable #2458 list syntax and #1988 list forwarding, but doesn't commit us to any particular design yet. I intend to write an RFC-style issue comment that proposes a concrete path forward and discusses the various tradeoffs.

### DIFF
--- a/src/evaluator.rs
+++ b/src/evaluator.rs
@@ -38,7 +38,7 @@ impl<'src, 'run> Evaluator<'src, 'run> {
           file_depth: 0,
           name: assignment.name,
           private: assignment.private,
-          value: value.clone(),
+          value: Val::from_str(value),
         });
       } else {
         unknown_overrides.push(name.clone());
@@ -65,7 +65,7 @@ impl<'src, 'run> Evaluator<'src, 'run> {
     Ok(evaluator.scope)
   }
 
-  fn evaluate_assignment(&mut self, assignment: &Assignment<'src>) -> RunResult<'src, &str> {
+  fn evaluate_assignment(&mut self, assignment: &Assignment<'src>) -> RunResult<'src, &Val> {
     let name = assignment.name.lexeme();
 
     if !self.scope.bound(name) {
@@ -83,40 +83,53 @@ impl<'src, 'run> Evaluator<'src, 'run> {
     Ok(self.scope.value(name).unwrap())
   }
 
+  /// A place for adding list operators in the future.
+  ///
+  /// List expressions return zero or more strings.
+  pub(crate) fn evaluate_list_expression(
+    &mut self,
+    expression: &Expression<'src>,
+  ) -> RunResult<'src, Vec<String>> {
+    // currently, all expression produce a single item
+    Ok(vec![self.evaluate_expression(expression)?.to_joined()])
+  }
+
   pub(crate) fn evaluate_expression(
     &mut self,
     expression: &Expression<'src>,
-  ) -> RunResult<'src, String> {
+  ) -> RunResult<'src, Val> {
     match expression {
       Expression::And { lhs, rhs } => {
         let lhs = self.evaluate_expression(lhs)?;
-        if lhs.is_empty() {
-          return Ok(String::new());
+        if lhs.to_joined().is_empty() {
+          return Ok(Val::new());
         }
         self.evaluate_expression(rhs)
       }
       Expression::Assert { condition, error } => {
         if self.evaluate_condition(condition)? {
-          Ok(String::new())
+          Ok(Val::new())
         } else {
           Err(Error::Assert {
-            message: self.evaluate_expression(error)?,
+            message: self.evaluate_expression(error)?.to_joined(),
           })
         }
       }
       Expression::Backtick { contents, token } => {
         if self.context.config.dry_run {
-          Ok(format!("`{contents}`"))
+          Ok(format!("`{contents}`").into())
         } else {
-          Ok(self.run_backtick(contents, token)?)
+          Ok(self.run_backtick(contents, token)?.into())
         }
       }
       Expression::Call { thunk } => {
         use Thunk::*;
-        let result = match thunk {
+        // All functions are currently of type (...String) -> Result<String>.
+        // They do not take or return a `Val`.
+        let result: Result<String, String> = match thunk {
           Nullary { function, .. } => function(function::Context::new(self, thunk.name())),
           Unary { function, arg, .. } => {
-            let arg = self.evaluate_expression(arg)?;
+            let arg = self.evaluate_expression(arg)?.to_joined();
             function(function::Context::new(self, thunk.name()), &arg)
           }
           UnaryOpt {
@@ -124,9 +137,9 @@ impl<'src, 'run> Evaluator<'src, 'run> {
             args: (a, b),
             ..
           } => {
-            let a = self.evaluate_expression(a)?;
+            let a = self.evaluate_expression(a)?.to_joined();
             let b = match b.as_ref() {
-              Some(b) => Some(self.evaluate_expression(b)?),
+              Some(b) => Some(self.evaluate_expression(b)?.to_joined()),
               None => None,
             };
             function(function::Context::new(self, thunk.name()), &a, b.as_deref())
@@ -136,10 +149,10 @@ impl<'src, 'run> Evaluator<'src, 'run> {
             args: (a, rest),
             ..
           } => {
-            let a = self.evaluate_expression(a)?;
+            let a = self.evaluate_expression(a)?.to_joined();
             let mut rest_evaluated = Vec::new();
             for arg in rest {
-              rest_evaluated.push(self.evaluate_expression(arg)?);
+              rest_evaluated.extend(self.evaluate_list_expression(arg)?);
             }
             function(
               function::Context::new(self, thunk.name()),
@@ -152,8 +165,8 @@ impl<'src, 'run> Evaluator<'src, 'run> {
             args: [a, b],
             ..
           } => {
-            let a = self.evaluate_expression(a)?;
-            let b = self.evaluate_expression(b)?;
+            let a = self.evaluate_expression(a)?.to_joined();
+            let b = self.evaluate_expression(b)?.to_joined();
             function(function::Context::new(self, thunk.name()), &a, &b)
           }
           BinaryPlus {
@@ -161,11 +174,11 @@ impl<'src, 'run> Evaluator<'src, 'run> {
             args: ([a, b], rest),
             ..
           } => {
-            let a = self.evaluate_expression(a)?;
-            let b = self.evaluate_expression(b)?;
+            let a = self.evaluate_expression(a)?.to_joined();
+            let b = self.evaluate_expression(b)?.to_joined();
             let mut rest_evaluated = Vec::new();
             for arg in rest {
-              rest_evaluated.push(self.evaluate_expression(arg)?);
+              rest_evaluated.extend(self.evaluate_list_expression(arg)?);
             }
             function(
               function::Context::new(self, thunk.name()),
@@ -179,19 +192,23 @@ impl<'src, 'run> Evaluator<'src, 'run> {
             args: [a, b, c],
             ..
           } => {
-            let a = self.evaluate_expression(a)?;
-            let b = self.evaluate_expression(b)?;
-            let c = self.evaluate_expression(c)?;
+            let a = self.evaluate_expression(a)?.to_joined();
+            let b = self.evaluate_expression(b)?.to_joined();
+            let c = self.evaluate_expression(c)?.to_joined();
             function(function::Context::new(self, thunk.name()), &a, &b, &c)
           }
         };
-        result.map_err(|message| Error::FunctionCall {
-          function: thunk.name(),
-          message,
-        })
+        result
+          .map(Val::from_str)
+          .map_err(|message| Error::FunctionCall {
+            function: thunk.name(),
+            message,
+          })
       }
       Expression::Concatenation { lhs, rhs } => {
-        Ok(self.evaluate_expression(lhs)? + &self.evaluate_expression(rhs)?)
+        let a = self.evaluate_expression(lhs)?.to_joined();
+        let b = self.evaluate_expression(rhs)?.to_joined();
+        Ok(Val::from_str(a + &b))
       }
       Expression::Conditional {
         condition,
@@ -205,19 +222,26 @@ impl<'src, 'run> Evaluator<'src, 'run> {
         }
       }
       Expression::Group { contents } => self.evaluate_expression(contents),
-      Expression::Join { lhs: None, rhs } => Ok("/".to_string() + &self.evaluate_expression(rhs)?),
+      Expression::Join { lhs: None, rhs } => {
+        let rhs = self.evaluate_expression(rhs)?.to_joined();
+        Ok(Val::from_str("/".to_string() + &rhs))
+      }
       Expression::Join {
         lhs: Some(lhs),
         rhs,
-      } => Ok(self.evaluate_expression(lhs)? + "/" + &self.evaluate_expression(rhs)?),
+      } => {
+        let lhs = self.evaluate_expression(lhs)?.to_joined();
+        let rhs = self.evaluate_expression(rhs)?.to_joined();
+        Ok(Val::from_str(lhs + "/" + &rhs))
+      }
       Expression::Or { lhs, rhs } => {
         let lhs = self.evaluate_expression(lhs)?;
-        if !lhs.is_empty() {
+        if !lhs.to_joined().is_empty() {
           return Ok(lhs);
         }
         self.evaluate_expression(rhs)
       }
-      Expression::StringLiteral { string_literal } => Ok(string_literal.cooked.clone()),
+      Expression::StringLiteral { string_literal } => Ok(Val::from_str(&string_literal.cooked)),
       Expression::Variable { name, .. } => {
         let variable = name.lexeme();
         if let Some(value) = self.scope.value(variable) {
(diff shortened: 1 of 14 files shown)


src/evaluator.rs
```diff
@@ -83,50 +83,63 @@ impl<'src, 'run> Evaluator<'src, 'run> {
     Ok(self.scope.value(name).unwrap())
   }
 
+  /// A place for adding list operators in the future.
```
casey, line 86: This is a little confusing to me, since it adds some indirection to follow to figure out what's going on. I think we should probably remove it, and just use `evaluate_expression`, and we can add it back later when we need it.

src/val.rs
```diff
@@ -0,0 +1,128 @@
...
+// The current implementation focuses on simplicity,
+// but tries to preserve flexibility for future optimizations
+// by mostly only dealing with owned data at the API boundaries (no borrowed views).
+
+/// The internal data type of Just, an immutable list of strings.
+///
+/// It can be viewed as a "joined string", or as the individual "parts".
+#[derive(Debug, Clone, Default)]
+pub(crate) struct Val {
+  parts: Rc<[Box<str>]>,
```
casey, line 29: `Rc<[Box<str>]>` is a little exotic. Is the advantage that extra capacity is freed when compared to an `Rc<Vec<String>>`?

I'd be in favor of using `Rc<Vec<String>>`, just because it's easier to read, and switching to `Rc<[Box<str>]` if we really need to.

src/val.rs
```diff
@@ -0,0 +1,128 @@
...
+
+  /// Construct a `Val` consisting of a single string.
+  pub fn from_str<S: AsRef<str>>(s: S) -> Self {
+    Self {
+      parts: Rc::new([s.as_ref().into()]),
+    }
+  }
+
+  /// Construct a `Val` consisting of multiple parts.
+  pub fn from_parts<I, S>(parts: I) -> Self
```
casey, line 46: I think `element` is the familiar term for things in array or linked list, so maybe this should be `from_elements`? and `val.parts` should be `val.elements`?

src/val.rs
```diff
@@ -0,0 +1,128 @@
...
+    I: IntoIterator<Item = S>,
+    S: AsRef<str>,
+  {
+    Self {
+      parts: parts.into_iter().map(|s| s.as_ref().into()).collect(),
+    }
+  }
+
+  /// Convert to a single joined string.
+  pub fn to_joined(&self) -> String {
```
casey, line 57: I think this can just be `join()`.

src/val.rs
```diff
@@ -0,0 +1,128 @@
...
+  pub fn to_parts(&self) -> Vec<Box<str>> {
+    self.parts.to_vec()
+  }
+}
+
+// Also provides a `to_string()` method
+impl std::fmt::Display for Val {
+  /// When formatted as a string, a `Val` is joined by whitespace.
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    f.write_str(&self.to_joined())
```
casey, line 71: This should do:

```
for (i, element) in &self.elements {
  if i > 0 {
    write!(f, " ")?; // or a f.write_char if that exists
  }
  f.write_str(element)?;
}

Ok(())
```

And then `join` should call `self.to_string()`. Doing it this way means that you don't need to allocate if you don't have to, i.e., you're printing to stdout or using it in a format string.

src/val.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+impl Val {
+  /// Create an empty `Val`
+  pub fn new() -> Self {
+    Self::default()
+  }
+
+  /// Construct a `Val` consisting of a single string.
+  pub fn from_str<S: AsRef<str>>(s: S) -> Self {
```
casey, line 39: I think this can be removed in favor of `val.into()` along with the `From` implementations.

src/val.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+
+  /// Convert to a single joined string.
+  pub fn to_joined(&self) -> String {
+    self.parts.join(" ")
+  }
+
+  /// Convert to individual parts.
+  pub fn to_parts(&self) -> Vec<Box<str>> {
```
casey, line 62: Can this return just `&[String]`, and rely on the caller to convert it to a vec? Avoids allocation if the caller doesn't need to.

src/val.rs
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+
+impl<A: AsRef<str>> FromIterator<A> for Val {
+  fn from_iter<T: IntoIterator<Item = A>>(iter: T) -> Self {
+    Self::from_parts(iter)
+  }
+}
+
+#[cfg(test)]
```
casey, line 95: This this and the tests should be inside a `#[cfg(test)] mod tests { … }` block.

tests/logical_operators.rs
```diff
@@ -81,3 +81,25 @@ fn nesting() {
   evaluate("'' || '' || '' || '' || 'foo'", "foo");
   evaluate("'foo' && 'foo' && 'foo' && 'foo' && 'bar'", "bar");
 }
+
+#[test]
+fn empty_variadics_are_falsey() {
```
casey, line 86: We should also add test cases for comparing empty variadics with the empty string with`==`, `!=`, `=~`, and `!~`, since those are cases where I can imagine accidentally introducing a breaking change.

tests/logical_operators.rs
```diff
@@ -81,3 +81,25 @@ fn nesting() {
...
+  #[track_caller]
+  fn eval_variadics<'a>(args: impl AsRef<[&'a str]>, expected: &'a str) {
+    Test::new()
+      .justfile("@foo *args:\n echo {{ args && 'true' || 'false' }}")
```
casey, line 90: Maybe we should have separate tests for `&&` and `||`. Operator precedence is always confusing, so separating them might help comprehension.

casey (changes requested):
Nice work!

This is bikeshedding, but what do you think calling it `List` instead of `Val`?

I originally suggested `Val`, but I'm thinking that the "everything is a list of strings" data model is somewhat exotic, so calling it a `List` explicitly in the codebase might be helpful for comprehension.

On the other hand, we could also leave it a `Val` for now, and consider this rename once we support user-facing changes which actually use it as a list.

There's aready a `List` type, but it's a minor type used to help formatting, so renaming the existing `List` to something else so we can use `List` for what is now called `Val` might be worth it.

Don't worry about making this change now, we can either make it before the PR goes in, or in a follow-up.

latk:
Thank you for the feedback, I'll get around to applying the review comments in the course of this week.

ofek:
@latk Hello! Are you still working on this?

latk:
Hi @ofek. No, sorry, I haven't found the “tuits” to finish this.

As this PR demonstrates, it's relatively easy to change the Just data model – but not at all easy to figure out appropriate semantics. It is not practically possible to implement multi-item lists without deciding how those lists should behave in various contexts, in particular deciding when a list should be joined into a single item.

Casey likes to point to the *rc shell* as prior art, but I think Just is more like a template language than like a shell. Interactions with boolean operators are tricky (e.g. the values `[]` vs `[""]` vs `""`, all of which might express an empty string). While the immediate goal of my change was to facilitate recipe argument forwarding `foo *args: (bar *args)`, once list-style values exist it's also necessary to figure out how they should interact with functions – if concatenation like `{{ files + ".c" }}` might distribute over the list, should functions given a list argument like `{{ quote(files) }}` also auto-distribute? Or join the list into a single space-separated string before invoking the function? But if functions can auto-distribute, what about variadic functions like `{{ join(";", files) }}` – should the list be splatted into multiple arguments, or should it be joined into a single string before invoking a function? A potential solution would be to distinguish between scalar and list contexts as in Perl, but if we can learn anything from Perl, then that we definitely shouldn't do this.

I was unable to come up with a cohesive design that I could propose, so I felt mentally blocked. Merging this PR (even once all review comments had been fixed) would have implemented semantics (always join lists) that are likely inappropriate in the vast majority of cases, and probably shouldn't ever be shipped in a release. It didn't feel right to work towards a temporarily-broken state without also knowing how to continue from there.

Some pointers for the next person who takes a jab at this:

* Just may be too complicated to be unityped. Neither single `String` nor multi-valued `List` seem to be a very good choice. It may be more helpful to use an `enum Value {...}` as in many other interpreters, potentially even a Json-ish data model.
* A large chunk of the complexity comes from the idea that operators might distribute/broadcast over lists, and that lists may sometimes be coerced into single strings. I believe this is a dead end, because this leads to edge cases that are impossible to resolve elegantly. Treating lists as a separate category of values, and requiring explicit conversion to/from lists, allows for safer exploration. For example, `{{ files + ".c" }}` might be a hard error, unless the user explicitly states an intention like `{{ join(" ", files) + ".c" }}` or `{{ [for f in files: f + ".c"] }}` (compare [Terraform/HCL for-expressions](https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md#for-expressions)).
* This PR attempted to maintain backwards compatibility by converting variadic recipe arguments into lists, and then joining lists into single strings wherever that had been expected by the current Just semantics. This was wrong, and makes it really difficult to maintain compatibility. This PR should only have introduced lists in the internal data model, but without any syntax to create them. Instead, a separate feature toggle should have changed how variadic recipe arguments are evaluated. This would have made it possible to safely explore the impact of lists, without conflicting with Just's backwards compatibility guarantees.

casey:
Very understandable, and thank you for the comment, lots of good stuff in there!

> but if we can learn anything from Perl, then that we definitely shouldn't do this.

😂

> Just may be too complicated to be unityped. Neither single String nor multi-valued List seem to be a very good choice. It may be more helpful to use an enum Value {...} as in many other interpreters, potentially even a Json-ish data model.

Can you expand on this a little more? I'm sort of inclined to agree, in the sense that I think a single type has downsides, many of which you mention. However, I'm not sure that on net, the alternatives are better. Being dynamically typed means run-time type errors and difficulty debugging, and being statically typed means making just much more complicated, not just adding a type system, but also lots of supporting syntax and operations.

On thing that could work is introducing lists, everything is a list and everything just gets joined with spaces in any context where it's used, but adding `.` methods which expose list functionality. Like `foo.join(bar)` if you want to join the list `foo` with a custom separator `bar`. Initially, all `.` methods are unstable, so they can be iterated on, and even possibly removed if the experiment doesn't pan out, and there is no change in behavior.

latk:
Thank you, I'll try to illustrate some of my thoughts on that matter.

## Different types might not be strictly necessary.

> > It may be more helpful to use an `enum Value {...}` as in many other interpreters, potentially even a Json-ish data model.
> 
> Can you expand on this a little more? […]
>
> On thing that could work is introducing lists, everything is a list and everything just gets joined with spaces in any context where it's used

That “everything is a list but gets joined where necessary” was attempted in this PR, and there are 35 instances of `.to_joined()` in this PR. This is definitely technically possible, it just might not be desirable.

* Users might be surprised about when lists are auto-joined into a single item.
* The empty string can be expressed via `[]` (empty list) or `['']`[^1] (empty list containing a single empty item). This might be difficult to teach, especially if these two values look the same in some cases, but have different semantics (e.g. different interactions with boolean operators). It might be preferable to make this list-nature more visible to users, rather than having a value that phases between strings and lists depending on context.
* Function calls are the big problem. Some function arguments must be a single string, others could be a list, for others users might want to splat/spread a list into variadic arguments. This PR didn't attempt to change the function interface (operating on plain strings), so all arguments had to be joined before. That was inflexible. It's probably more appropriate to allow functions to operate on lists.

For none of this, a distinction between single strings and lists is strictly necessary. However, using an enum might enable more clarity when designing behaviors. In particular, conversions from a plain string to a single-item list might become more visible.

[^1]: I'm well aware that Just has no expression-level list syntax, but to illustrate the runtime values I'll use the same `["a", "b", "c"]` syntax that's used in Just settings and in many other languages like JSON, JavaScript, Python.

## Different types may enable useful function overloading.

I'm looking at the documentation for Just functions to see whether they would be sensitive to different str vs list representations. Given a list, functions generally have the choice of joining that list into a single string, or applying the function's behavior to each element (as in array programming languages like R or Python's Numpy). For example, `env(["A", "B"])`[^1] could be defined to evaluate as `env("A B")` or `[env("A"), env("B")]` and either design is reasonable/consistent. But there are a couple of more tricky cases.

* `shell(["some", "command"])` – joining into `shell("some command")` might be insecure, and invoking each list item as a separate command will rarely be intuitive. However, if the `shell(command)` function can distinguish between a single string vs a list, then it could offer different semantics – like skipping the shell and invoking the command directly.
* Functions `append()`/`prepend()` currently operate on whitespace-separated strings. Auto-joining lists would keep the same behavior, e.g. `append(".c", ["foo", "bar"])` could output `["foo.c", "bar.c"]` or `"foo.c bar.c"`, with no meaningful difference between the two. However, being able to distinguish between strings and lists would make it possible to safely use this function on strings that contain spaces. For example, `append(".c", ["foo", "bar baz"])` would be able to return `["foo.c", "bar baz.c"]`, rather than `"foo.c bar.c baz.c"`.
* `trim([" a ", " b "])` should likely produce different results from `trim(" a   b ")` with respect to the internal space. However, this could also be satisfied with an explicit looping construct, e.g. `[for s in [" a ", " b "]: trim(s)]`.
* `choose("1", ["ab", "cd"])` should likely produce different results from `choose("1", "ab cd")`
* I am unsure about the implications for `replace()` and `replace_regex()` functions.

So, there might be an argument that for some functions, being able to distinguish between single strings and lists has value, which would require an `enum Value { ... }`. And if Just already makes that change, it could be worth considering supporting an entire JSON-like data model, especially since boolean-like values are currently inconsistent (some operations use empty strings for falsey values, others return the string `"false"`).

## Different types can be user-friendly, and Just is well-suited for static type checks.

> Being dynamically typed means run-time type errors and difficulty debugging […] being statically typed means making just much more complicated

I don't think that's the case.

There is a ton of prior art on dynamically typed languages. Useful implicit conversions can be designed wherever appropriate, e.g. wrapping a string in a single-element list, or joining a list into a space-separated string. I think, implicit conversions have a bad reputation largely because JavaScript's operators have some surprising asymmetries, especially when it comes to concepts of equality between arbitrary objects. But it is not necessary to repeat the same mistakes. In particular, Just has the advantage that values are immutable, whereas some JS problems stem from having to account for object identity.

Runtime errors are rare in Just, but do already exist. E.g. note `choose("a", "b")`, which is essentially a runtime type error (`a` is not an integer). And of course, the `error()` function.

Static checks or syntactic features can guide users away from things that don't work, but aren't always tractable. For example, if Just were to get support for arbitrary nested data structures, static checks would be difficult or useless. However, if there are only types “string” and “list of strings” (without nesting), static checks might very well be easy. Some languages use sigils like `@foo` to statically mark variables as list-only (compare also the current `*args` syntax in Just recipes).

Fortunately, Just variables are assign-once and immutable, and there are no functions where the return type would be dynamic, so complete type inference would be unusually easy. (Exception: expressions like `if cond { "string" } else { ["list"] }` would have a statically unknowable type as it depends on the runtime value of `cond`, but this can be solved via implicit conversions, or by turning that into a static error, or by modelling a `string | list` union type).

Adding static type checks would make Just a lot more complicated because the type checker would duplicate a lot of the evaluation code. But also – not as much as it might seem! This is basically a much condensed version of the evaluator's `match` statement, plus the `Function`/`Thunk` concept would have to be extended to allow the return type to be computed from the argument types. The `Function::argc()` concept is already rudimentary type information. Types could be resolved together with name resolution, since type information only has to flow “forwards”.

Personally, I suspect that the same language features of Just that make static type checking relatively easy, also mean that static checks are relatively useless. Variables are relatively static, and there are few control flow constructs, and Just is mostly pure (aside from shell/backticks). Testing an expression by evaluating it will mostly detect the same errors as any type check (with some caveats due to laziness, and expressions within recipes).

OpenTofu/Terraform/HCL is relevant prior art (see [docs on expressions](https://opentofu.org/docs/language/expressions/)). It is a mostly-pure language operating on a JSON data model. It is effectively dynamically typed, with `any` types being common in practice, but static type constraints can be declared at module boundaries – useful for establishing explicit stable interfaces. This is not so pressing for Just, because Just doesn't have user-defined functions, but it shows that languages in this space can strike a successful balance. 

## Making backwards-compatible changes.

> On thing that could work is introducing lists, everything is a list and everything just gets joined with spaces in any context where it's used, but adding . methods which expose list functionality. […] Initially, all `.` methods are unstable, so they can be iterated on, and even possibly removed if the experiment doesn't pan out, and there is no change in behavior.

Deciding that lists are joined by default is a one-way-door decision. This PR attempted that, and it would have committed Just to potentially-undesirable semantics, especially with regard to function calls. That was a mistake.

Experimenting with unstable `list.methods()` only allows exploring such methods, but not any other interactions between lists and other Just features.

The minimal user-visible feature set to safely explore list values would be:

* a way to create lists (e.g. variadic recipe arguments, or `[...]` syntax) – these mechanisms should be guarded behind a feature toggle
* an *explicit* way to convert lists to strings – implicit conversions can be explored later
* any other usage of lists should initially result in an error

Then, uses that would initially be an error can be investigated one after another and get more useful semantics, without breaking previously-working code.

Issue #2566, opened 2025-01-08
Setting to apply `[no-exit-message]` to all recipes


xavdid:
Most of our justfile usage so far has been to wrap other tools, so we've been making heavy use of the `[no-exit-message]` attribute. That got me thinking: it would be nice to be able to apply that to everything and reduce some of the cruft. 

We'd also probably want a matching `[exit-message]` annotation to invert the setting as needed (similar to `set quiet` and `[no-quiet]`).

casey:
That sounds reasonable to me.

xavdid:
This was added as `no-exit-message` in 1.39.0. Closing!

PR #2568, opened 2025-01-12
Adds `no-exit-message` Setting and `[exit-message]` attribute

ArchieAtkinson:
This adds both the `no-exit-message` setting and the `[exit-message]` attribute. Current behaviour is `[exit-message]` overrides both `set no-exit-message` and `[no-exit-message]`.

Idea from #2566. 

I would like feedback on whether I need to add anything to the documents and whether any test cases are missing.


src/attribute.rs
```diff
@@ -17,6 +17,7 @@ pub(crate) enum Attribute<'src> {
   Macos,
   NoCd,
   NoExitMessage,
+  ExitMessage,
```
casey: These should be sorted alphabetically.

src/attribute.rs
```diff
@@ -36,6 +37,7 @@ impl AttributeDiscriminant {
       | Self::Macos
       | Self::NoCd
       | Self::NoExitMessage
+      | Self::ExitMessage
```
casey: These should be sorted alphabetically.

src/attribute.rs
```diff
@@ -84,6 +86,7 @@ impl<'src> Attribute<'src> {
       AttributeDiscriminant::Macos => Self::Macos,
       AttributeDiscriminant::NoCd => Self::NoCd,
       AttributeDiscriminant::NoExitMessage => Self::NoExitMessage,
+      AttributeDiscriminant::ExitMessage => Self::ExitMessage,
```
casey: Sort

src/attribute.rs
```diff
@@ -133,6 +136,7 @@ impl Display for Attribute<'_> {
       | Self::Macos
       | Self::NoCd
       | Self::NoExitMessage
+      | Self::ExitMessage
```
casey: Sort

src/justfile.rs
```diff
@@ -724,6 +724,54 @@ mod tests {
     }
   }
 
+  run_error! {
```
casey: I'm trying to move away from the test generation macros. New tests should be integration tests, and use `Test::new()`, see examples in the `tests` directory.

src/justfile.rs
```diff
@@ -724,6 +724,54 @@ mod tests {
...
+    },
+    check: {
+      assert_eq!(recipe, "fail");
+      assert_eq!(code, 100);
+      assert_eq!(line_number, Some(5));
+      assert!(print_message);
+    }
+  }
+
+  run_error! {
```
casey: Convert to integration test with `Test::new()`

src/keyword.rs
```diff
@@ -21,6 +21,7 @@ pub(crate) enum Keyword {
   Mod,
   PositionalArguments,
   Quiet,
+  NoExitMessage,
```
casey: Sort

src/node.rs
```diff
@@ -293,6 +293,7 @@ impl<'src> Node<'src> for Set<'src> {
       | Setting::Fallback(value)
       | Setting::PositionalArguments(value)
       | Setting::Quiet(value)
+      | Setting::NoExitMessage(value)
```
casey: Sort

src/parser.rs
```diff
@@ -1095,6 +1095,7 @@ impl<'run, 'src> Parser<'run, 'src> {
       Keyword::IgnoreComments => Some(Setting::IgnoreComments(self.parse_set_bool()?)),
       Keyword::PositionalArguments => Some(Setting::PositionalArguments(self.parse_set_bool()?)),
       Keyword::Quiet => Some(Setting::Quiet(self.parse_set_bool()?)),
+      Keyword::NoExitMessage => Some(Setting::NoExitMessage(self.parse_set_bool()?)),
```
casey: Sort

src/recipe.rs
```diff
@@ -131,10 +131,10 @@ impl<'src, D> Recipe<'src, D> {
       || (cfg!(windows) && windows)
   }
 
-  fn print_exit_message(&self) -> bool {
-    !self
-      .attributes
-      .contains(AttributeDiscriminant::NoExitMessage)
+  fn print_exit_message(&self, settings: &Settings) -> bool {
```
casey, line 134: How about this? It reads a little clearer to me:

```rust
fn print_exit_message(&self, settings: &Settings) -> bool {
  if settings.no_exit_message {
    self.attributes.contains(AttributeDiscriminant::ExitMessage)
  } else {
    !self.attributes.contains(AttributeDiscriminant::NoExitMessage)
  }
}
```

src/setting.rs
```diff
@@ -13,6 +13,7 @@ pub(crate) enum Setting<'src> {
   IgnoreComments(bool),
   PositionalArguments(bool),
   Quiet(bool),
+  NoExitMessage(bool),
```
casey: Sort

src/setting.rs
```diff
@@ -34,6 +35,7 @@ impl Display for Setting<'_> {
       | Self::IgnoreComments(value)
       | Self::PositionalArguments(value)
       | Self::Quiet(value)
+      | Self::NoExitMessage(value)
```
casey: Sort

src/settings.rs
```diff
@@ -18,6 +18,7 @@ pub(crate) struct Settings<'src> {
   pub(crate) ignore_comments: bool,
   pub(crate) positional_arguments: bool,
   pub(crate) quiet: bool,
+  pub(crate) no_exit_message: bool,
```
casey: Sort

src/settings.rs
```diff
@@ -67,6 +68,9 @@ impl<'src> Settings<'src> {
         Setting::Quiet(quiet) => {
           settings.quiet = quiet;
         }
+        Setting::NoExitMessage(no_exit_message) => {
```
casey: Sort

tests/json.rs
```diff
@@ -86,6 +86,7 @@ struct Settings<'a> {
   ignore_comments: bool,
   positional_arguments: bool,
   quiet: bool,
+  no_exit_message: bool,
```
casey: Sort

tests/no_exit_message.rs
```diff
@@ -129,3 +129,67 @@ hello:
   stderr: "",
   status: 100,
 }
+
+test! {
```
casey: These should use the `Test::new()` builder.

tests/no_exit_message.rs
```diff
@@ -129,3 +129,67 @@ hello:
...
+  stdout: "Hello, World!\n",
+  stderr: r#"
+    echo 'Hello, World!'
+    exit 100
+  "#,
+  status: 100,
+}
+
+
+test! {
```
casey: I think this test can be removed, since whether or not a recipe is silent doesn't interact with exit messages.

tests/no_exit_message.rs
```diff
@@ -129,3 +129,67 @@ hello:
...
+  #!/usr/bin/env bash
+  echo 'Hello, World!'
+  exit 100
+",
+  stdout: "Hello, World!\n",
+  stderr: "",
+  status: 100,
+}
+
+test! {
```
casey: I think this test can be removed. It would be a quite weird interaction in this code if this failed.

casey (changes requested):
Looks good!

- I think there is an additional test case, just a plain recipe with the `[no-exit-message]` attribute with no setting.
- - `[exit-message]` and `[no-exit-message]` on the same recipe should produce an error, and this should also have a test

tests/no_exit_message.rs
ArchieAtkinson: Done

tests/json.rs
ArchieAtkinson: done

src/attribute.rs
ArchieAtkinson: Done

src/attribute.rs
ArchieAtkinson: Done

src/attribute.rs
ArchieAtkinson: Done

src/attribute.rs
ArchieAtkinson: Done

src/keyword.rs
ArchieAtkinson: Done

src/node.rs
ArchieAtkinson: Done

src/parser.rs
ArchieAtkinson: Done

src/justfile.rs
ArchieAtkinson: Done

src/justfile.rs
ArchieAtkinson: Done

src/recipe.rs
ArchieAtkinson: This isn't logically equivalent, the test case you suggested would fail. 

Updated to the below which I think is easier to read:
```rust
    if self.attributes.contains(AttributeDiscriminant::ExitMessage) {
      return true;
    }
    
    if settings.no_exit_message {
        return false;
    }
  
    !self.attributes.contains(AttributeDiscriminant::NoExitMessage)
```

src/setting.rs
ArchieAtkinson: Done

src/setting.rs
ArchieAtkinson: Done

src/settings.rs
ArchieAtkinson: Done

src/settings.rs
ArchieAtkinson: Done

tests/no_exit_message.rs
ArchieAtkinson: Done

tests/no_exit_message.rs
ArchieAtkinson: Done

ArchieAtkinson (commented):
All comments done.

Added additional suggested tests and moved all tests to `no_exit_messsage.rs`.

Also added a commit to refactor the other tests in `no_exit_messsage.rs` to `Test::new()` if that is suitable for this PR. 

Let me know if there is anything else that needs changing.

src/recipe.rs
casey: Nice, yeah that's much better.

casey (changes requested):
Looks good! I merged latest master in, and changed the order of tests so they're easier to review.

Only one thing left, having `[exit-message]` and `[no-exit-message]` on the same recipe should be a compiler error.

You can see an example with the `working-directory` attribute and `no-cd` attribute in parser.rs:

```rust
    let working_directory = attributes.contains(AttributeDiscriminant::WorkingDirectory);

    if working_directory && attributes.contains(AttributeDiscriminant::NoCd) {
      return Err(
        name.error(CompileErrorKind::NoCdAndWorkingDirectoryAttribute {
          recipe: name.lexeme(),
        }),
      );
    }
```

You'll have to create a new `CompileErrorKind` variant.

ArchieAtkinson:
@casey added the requested changes, thank you for reviewing and for the feedback!

casey (approved):
LGTM!

Issue #1658, opened 2023-08-10
`trim_end_match` doesn't work as expected


avi-cenna:
I'm not sure if I'm misunderstanding the documentation, but I would expect the following lines to print the same result.

```bash
test:
    echo {{ replace_regex('ping-pong', '-.*$', '') }}
    echo {{ trim_end_match('ping-pong', '-.*$') }}
```

In fact, the first line prints `ping` as expected, but the second line prints `ping-pong`. Isn't `trim_end_match` supposed to take a regular expression?

casey:
Actually this is as-intended. Only `_regex` takes a regular expression, the other `_match` functions take a string literal which is matched verbatim.

PR #2574, opened 2025-01-14
Clarify behavior of `trim_end_match` in `README`

### LINKED ISSUE #1983: clarify the `matching pat` in replacement documentation
[String manipulation](https://just.systems/man/en/chapter_31.html#string-manipulation) contains this section which is not very clear. Does `pat` refer to an exact string match, e.g. removing `trim_end_match("foobar", "bar")` would result in `"foo"`, or does it support some pattern matching (as the variable name would imply)? I tried it with regex, but it seems like it is not working.

```
trim_end_match(s, pat) - Remove suffix of s matching pat.
trim_end_matches(s, pat) - Repeatedly remove suffixes of s matching pat.
trim_start_match(s, pat) - Remove prefix of s matching pat.
trim_start_matches(s, pat) - Repeatedly remove prefixes of s matching pat.
```

xavdid:
I was super confused about why my regex wasn't working with `trim_end_match` and it wasn't until I found https://github.com/casey/just/issues/1658 that I figured out why.

I think calling the second parameter `pat`tern implies that it's a regex pattern when it's not. I changed it to `substr` to imply it's just a regular string (it is), we could also add additional docs to that effect.

fixes https://github.com/casey/just/issues/1983

### DIFF
--- a/README.md
+++ b/README.md
@@ -1791,13 +1791,13 @@ The process ID is: 420
   [Replacement string syntax](https://docs.rs/regex/latest/regex/struct.Regex.html#replacement-string-syntax).
 - `trim(s)` - Remove leading and trailing whitespace from `s`.
 - `trim_end(s)` - Remove trailing whitespace from `s`.
-- `trim_end_match(s, substr)` - Remove suffix of `s` matching `substr`.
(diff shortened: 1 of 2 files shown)


casey:
Good idea, that's much better. I changed `substr` to `substring`.

PR #2596, opened 2025-01-22
Fix README.md copy paste error

CramBL:
Seems like it was copy pasted but forgot to replace `require` with `which`.

EDIT: Also did a pass with [typos](https://github.com/crate-ci/typos) and found a single typo.


casey:
Good catch, thank you!

PR #2610, opened 2025-01-29
Prefer pwsh.exe over powershell.exe in documentation

### LINKED ISSUE #2604: Update powershell example
*Issue*

Windows now mostly uses Powershell V2, not `powershell.exe`

*Ask*

Consider updating  https://github.com/casey/just/blob/master/examples/powershell.just

casey:
Fixes #2604.

@john-cd Since which version of windows is `pwsh.exe` installed on by default? I want to make sure it's widely available before changing the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ Like PowerShell:
 
 ```just
 # use PowerShell instead of sh:
-set shell := ["powershell.exe", "-c"]
+set shell := ["pwsh.exe", "-c"]
 
 hello:
   Write-Host "Hello, world!"
@@ -120,7 +120,7 @@ list:
 ```
 
 You can also set the shell using command-line arguments. For example, to use
-PowerShell, launch `just` with `--shell powershell.exe --shell-arg -c`.
+PowerShell, launch `just` with `--shell pwsh.exe --shell-arg -c`.
 
 (PowerShell is installed by default on Windows 7 SP1 and Windows Server 2008 R2
 S1 and later, and `cmd.exe` is quite fiddly, so PowerShell is recommended for
@@ -1211,7 +1211,7 @@ an additional flag, often `-c`, to make them evaluate the first argument.
 use `windows-shell`:
 
 ```just
-set windows-shell := ["powershell.exe", "-NoLogo", "-Command"]
+set windows-shell := ["pwsh.exe", "-NoLogo", "-Command"]
 
 hello:
   Write-Host "Hello, world!"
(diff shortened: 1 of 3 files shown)


john-cd:
Not the expert, but the transition from "Windows PowerShell" to cross-platform "PowerShell" started with PowerShell Core 6, GA on January 10, 2018. We are now at "Powershell 7.5": https://en.wikipedia.org/wiki/PowerShell

Now, MS designed old Windows PowerShell and new Powershell to run side-by-side. 

"Windows PowerShell is the version of PowerShell that ships in Windows. This version of PowerShell uses the full .NET Framework, which only runs on Windows. The latest version is Windows PowerShell 5.1. Microsoft is no longer updating Windows PowerShell with new features."  https://learn.microsoft.com/en-us/powershell/scripting/what-is-windows-powershell 

So MS still ships Windows PowerShell, but for a couple of years there has been a nag message that prompts you to install cross-platform Powershell aka pwsh.exe. 

So, developers could use both but most likely have upgraded to pwsh

casey:
Hmm, interesting. @runeimp do you happen to know? I'd like to change all the docs to use `pwsh.exe` instead of `powershell.exe`, but want to make sure that `pwsh.exe` is available by default on enough recent versions of windows.

runeimp:
As I understand it (guesswork based on Microsoft information which is generally deplorable) `pwsh.exe` (PowerShell 6+) ships with .NET Core 2.0+; or later, simply _.NET_. Windows 11 ships with .NET Core (simply called .NET eventually) at some point. I understand Windows 11 22H2 (released September 2022) was the first version with the .NET Framework 4.8.1. This _MIGHT_ be the first version of Windows 11 that includes `pwsh.exe` instead of `powershell.exe`.

doterik:
I recently installed Windows 11 Pro 23H2 and then upgraded to 24H2 to use for various tests. On this machine, surprisingly, I didn't get `pwsh` installed by default (only powershell), so it might be on 0% of installations, I'm not sure...

On the other hand, for those who install and use `just`, I would guess that 99% of them have `pwsh` on their machines.

If/when you make changes, also consider:
"--completions ... [possible values: bash, elvish, fish, nushell, powershell, zsh]"

cspotcode:
AFAIK `pwsh` is never installed by default.  At least, I haven't heard anything to the contrary.

At the very least, installing the latest `pwsh` is as easy as installing `just` via winget:

```shell
winget install Casey.Just
winget install Microsoft.PowerShell
```

Maybe that's good enough?  I forget which versions of Windows include `winget`.

casey:
@cspotcode I think you're right. I'm not sure where I got the notion that `pwsh.exe` was ever installed by default. I think I just assumed that since it was new and shiny, it would eventually replace `powershell.exe` as the default.

casey:
Closing, since I think we have to go with the installed default powershell.

Issue #2349, opened 2024-09-06
[alias] attribute


casey:
PR #2344 added a new attribute, `[alias]` providing an alternative way for declaring recipe aliases. I personally lean towards not adding a new attribute, but I don't feel super strongly, so if it's something that people wanted, I wouldn't mind adding it.

I think ideally, would take a string literal: `[alias: "foo"]`, which would initially be constrained to be a valid identifier. This could be relaxed in the future to allow non-identifier aliases.

See the original PR for some of the discussion.

runeimp:
I like the idea of keeping all parts of a recipe in one place. Especially when considering modules, etc.

marcaddeo:
@casey I'm not sure if you had a chance to fully review the original PR, but my [first commit is my first attempt in which I tried to use the `[alias("string")]` syntax](https://github.com/casey/just/pull/2344/commits/d3aedbfa30b30aa1e0fccf927139a893cc546193). 

I wasn't liking how the code turned out, and felt like I was fighting Just to implement it. That's why I moved to using Identifiers/Names/Tokens (not sure which is the most technically correct qualifier here 😄), and because aliases are already processed that way.

I've made a couple more attempts at implementing it with strings, but haven't been successful.

Would love to make another attempt and get this feature eventually merged, but if you could possibly point me in the right direction for implementing this with strings (if you've got any ideas for it) that would be super helpful.

casey:
I'm still honestly not totally sold on this feature, since it feels like it just duplicates the existing way of creating aliases. `just` is already pretty bloated 😅 Although in some ways that's what makes it a nice tool. So I don't want to send you off trying to get strings to work if it won't ultimately get merged. I think I'm inclined to let this issue sit for a bit before acting on it, to get some more feedback.

xavdid:
I would also like this! Having everything in one block reads nicer and makes it easier to do natural grouping of the definition:

```just
thing:
  echo whatever

alias extra := thing
```

vs

```just
[alias("extra")]
thing:
  echo whatever
```

---

_edit_: This is coming soon, apparently!

Similarly, it would be nice if aliases weren't listed on their own lines for summary. Today:

```
    thing              # some cool script
    double_extra       # alias for `thing`
    extra              # alias for `thing`
```

Ideally:

```
    thing              # some cool script (aliases: extra, double_extra)
```

or similar.

marcaddeo:
> Similarly, it would be nice if aliases weren't listed on their own lines for summary. Today:
> 
[...]

PR #2615, opened 2025-02-01
Add alias attribute

marcaddeo:
Implements #2349

### DIFF
--- a/src/alias.rs
+++ b/src/alias.rs
@@ -1,7 +1,7 @@
 use super::*;
 
 /// An alias, e.g. `name := target`
-#[derive(Debug, PartialEq, Clone, Serialize)]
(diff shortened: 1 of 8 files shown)


marcaddeo:
@casey how do you feel about this implementation?

casey (changes requested):
Just took a first look at this!

I think it's probably not a good idea to try to construct a `Name` token, when the source was in fact a string literal, and that `Alias` should be refactored to make the `name` field a `String`. If we need a token so we can report error locations, we shoiuld use a `Token`, and the underlying `TokenKind` may be a `Identifier` or a `StringLiteral`, depending on the source.

You can use `Lexer::is_identifier` to check if `StringLiteral::cooked` is a valid identifier. `StringLiteral::cooked` should be use, since it may contain escape characters.

casey:
Closing, since this seems to have stalled out.

Issue #2629, opened 2025-02-11
Aliases to recipes in submodules


john-cd:
To "alias" a recipe in a module, I believe one has to call `just` again:  

```
# Format all bin and lib files
f:
  @just code fmt

# Format all packages and their local path-based dependencies
fa:
  @just code fmtall
```

Could we have the following? 

```
alias f  := code::fmt 
```

casey:
Good suggestion! I added it to #2252.

I think this should be pretty easy to add, so added the good first issue label in case anyone wants to take a crack at it.

corvusrabus:
Can I grab this one?

casey:
Sure, go for it! Rough sketch of what I think will be needed:

1. Add a `ColonColon` token kind, which will be used for path separators
2. Add a `Parser::parse_path` function, which will parse a colon-separated path of names
3. Change `Parser::parse_alias` to parse a path
4. During analysis, verify that the path points to a valid recipe, and resolve the alias to that recipe.

4 is probably the trickiest, since this is the first time we've had to deal with inter-module dependencies.

casey:
Implemented in #2632! It's been a while since the last release, so I'm releasing 1.40.0 with it.

PR #2632, opened 2025-02-17
feat: module path aliases

corvusrabus:
Implements #2629 .

The usage of `NamePath` feels clunky. Maybe I could make it so that it's guaranteed that `NamePath` is not empty and that all tokens sit after each other separated by ColonColon. Then, we could generate a `&str` and avoid the usage of `String` in `UnknownAliasTarget`


src/compile_error_kind.rs
```diff
@@ -138,7 +138,7 @@ pub(crate) enum CompileErrorKind<'src> {
   UnicodeEscapeUnterminated,
   UnknownAliasTarget {
     alias: &'src str,
-    target: &'src str,
+    target: String,
```
casey: Can this be a namepath?

```suggestion
    target: Namepath<'src>,
```

src/analyzer.rs
```diff
@@ -289,19 +289,61 @@ impl<'run, 'src> Analyzer<'run, 'src> {
...
       Some(target) => Ok(alias.resolve(Rc::clone(target))),
       None => Err(alias.name.token.error(UnknownAliasTarget {
         alias: alias.name.lexeme(),
-        target: alias.target.lexeme(),
+        target: target_str,
       })),
     }
   }
+
+  fn traverse_nonempty_module_name_path<'modules>(
```
casey: I think this can be made simpler if it's recursive. Maybe with this signature:

```
fn foo(
  path: &[Name],
  modules: &Table<Justfile>,
  recipes: &Table<Rc<Recipe>>,
) -> Result<Option<Rc<Recipe>>, ()>;
```

If the path is empty, then resolution has failed, and you can return `Err(())`.

If path is length one, then look in `recipes` for the target recipe.

Otherwise, try to find the next module to look in using the first item in `path`. If not found, then return an error. If found, then make the recursive call with `path[1..]`, and the submodules recipes and modules.

src/analyzer.rs
```diff
@@ -289,19 +289,61 @@ impl<'run, 'src> Analyzer<'run, 'src> {
...
+      name,
+      target,
+    } = alias;
+    let target_str = format!("{target}");
+    let mut path = target.into_inner();
+    let target = path
+      .pop()
+      .expect("Internal error: Alias target path can't be empty");
+
+    let alias = Alias {
```
casey: I think that `resolve` can be changed so that it's implemented for the correct type, so we don't need to take apart and put the alias back together again.

src/parser.rs
```diff
@@ -863,6 +863,19 @@ impl<'run, 'src> Parser<'run, 'src> {
     self.expect(Identifier).map(Name::from_identifier)
   }
 
+  /// Parse a path of the forms `a` or `a::b` or `a::b::c` etc.
+  fn parse_double_colon_path(&mut self) -> CompileResult<'src, Namepath<'src>> {
```
```suggestion
  fn parse_namepath(&mut self) -> CompileResult<'src, Namepath<'src>> {
```

tests/alias.rs
```diff
@@ -0,0 +1,45 @@
+use super::*;
+#[test]
+fn alias_nested_module() {
```
casey, line 4:
```suggestion

fn alias_nested_module() {
```

casey (changes requested):
See comments! I think that this can be made simpler by making the analysis function recursive. Also, I landed another PR which guarantees that a namepath is non-empty. Although the addition of `new` in this PR breaks that. Although I think the actual answer is using a recursive function and erroring when the path is empty. See comments.

corvusrabus:
Sorry I've been a bit busy. Will follow up on the weekend

src/analyzer.rs
corvusrabus: Hmm I didn't see added value in having an `Err(())` and a `Ok(None)` so I made it just an Option with `None` indicating that the recipe was not found

corvusrabus:
@casey I 
1. removed the `NamePath::new` so that len >=1 is again guaranteed
2. Needed to add new `PartialEq` for `NamePath` because namepaths should be considered equal if they lex to the same path

Once we merge this I'd be happy to implement dependencies on recipes in other modules, which I think is 60% done with this MR already (that's the feature that I came here for in the first place :) )

src/analyzer.rs
corvusrabus: Done

src/analyzer.rs
```diff
@@ -318,14 +336,16 @@ mod tests {
     kind: Redefinition { first_type: "alias", second_type: "alias", name: "foo", first: 0 },
   }
 
+  const UNKNOWN_ALIAS_TARGET: &str = "alias foo := bar\n";
```
casey: I don't like these kinds of constants which do a very limited amount of deduplication. I'd just inline this in both places.

src/analyzer.rs
```diff
@@ -319,14 +337,16 @@ mod tests {
...
   analysis_error! {
     name: unknown_alias_target,
-    input: "alias foo := bar\n",
+    input: UNKNOWN_ALIAS_TARGET,
     offset: 6,
     line: 0,
     column: 6,
     width: 3,
-    kind: UnknownAliasTarget {alias: "foo", target: "bar"},
+    kind: UnknownAliasTarget {alias: "foo", target: Namepath::from(Name::from_identifier(Token{
```
casey: This should be formatted to be on multiple lines:

```suggestion
    kind: UnknownAliasTarget {
      alias: "foo",
      target: Namepath::from(Name::from_identifier(Token {
        column: 13,
        …
```

src/namepath.rs
```diff
@@ -42,3 +63,11 @@ impl Serialize for Namepath<'_> {
     serializer.serialize_str(&format!("{self}"))
   }
 }
+
+impl<'src> Deref for Namepath<'src> {
```
casey: We shouldn't add `deref` to `Namepath`. It means that any function on `Vec` (I actually changed it to slice before I decided we shouldn't provide it at all.) is valid for namepath, which I think is not ideal. See comment about `last` later on.

src/analyzer.rs
```diff
@@ -290,19 +290,37 @@ impl<'run, 'src> Analyzer<'run, 'src> {
...
       })),
     }
   }
+
+  fn recipe_from_path<'a>(
+    path: &[Name],
+    mut modules: &'a Table<'src, Justfile<'src>>,
+    mut recipes: &'a Table<'src, Rc<Recipe<'src>>>,
+  ) -> Option<Rc<Recipe<'src>>> {
+    let name = path.last()?;
```
casey: What happens if `.last()` return `None`? This will be treated as an unknown alias target, but is in fact a bug, since namepath should be non-empty, and thus if `.last()` returns `None` it should be an internal error. This is why we shouldn't iplement `Deref` for `Namepath`, because the function signature `fn last(&self) -> Option<Name>` doesn't make sense. It should be `fn last(&self) -> Name`.

src/namepath.rs
```diff
@@ -1,8 +1,21 @@
 use super::*;
 
-#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[derive(Clone, Debug, Eq, Ord, PartialOrd)]
 pub(crate) struct Namepath<'src>(Vec<Name<'src>>);
 
+impl PartialEq for Namepath<'_> {
```
casey: I don't think we should implement a `PartialEq` implementation which ignores everything but the lexeme.

I think we should do `fn is_same_path(&self, other: &Self) -> bool` which compares the lexemes.

src/namepath.rs
```diff
@@ -14,6 +27,14 @@ impl<'src> Namepath<'src> {
       spaced: true,
     }
   }
+
+  pub fn push(&mut self, name: Name<'src>) {
+    self.0.push(name);
+  }
+
+  fn lexeme_iter(&self) -> impl ExactSizeIterator<Item = &str> {
```
casey: I don't think a helper function which is only used in one place adds much. `.all(|(a, b)| a.lexeme() == b.lexeme())` is fine.

casey (changes requested):
Nice, much cleaner! See comments.

src/analyzer.rs
corvusrabus: Will `expect()` panic now saying that `Namepath` can't be empty

src/namepath.rs
```diff
@@ -14,6 +27,37 @@ impl<'src> Namepath<'src> {
...
+      .0
+      .last()
+      .expect("Internal error: Namepath can not be empty")
+  }
+
+  /// Splits a module path into the last name and its parent
+  /// a::b::c -> (a::b, c)
+  /// a       -> (empty, a)
+  pub fn split_last(&self) -> (&[Name<'src>], &Name<'src>) {
+    let name = self.last();
```
corvusrabus: Assigning this variable first instead of making this a one liner because `last()` will give a more meaningful panic message than `&self.0[..self.0.len() - 1]` panicking due to wrong indexing if `len()==0`

casey (approved):
LGTM! I made some tweaks and fixed a small bug: The assert could actual fire. If you defined a recipe in a submodule, the path you referred to it with, `bar::baz`, might not be its root path, `foo::bar::baz`.

PR #2637, opened 2025-02-23
Improve unknown start of token error message for invisible characters

### LINKED ISSUE #1016: Improve unknown start of token error message for invisible characters
This came up because someone had a BOM in their justfile, which just was choking on.

The error message looked like:

```
error: Unknown start of token:
  |
1 | ﻿test:
  | ^
```

We should print the unknown start of token message such that invisible characters show up, for example by surrounding the offending character in quotes. We could also think about displaying non-printing characters as unicode escape sequences. Perhaps there's a function in the standard library that helps with that.

CramBL:
Resolves https://github.com/casey/just/issues/1016

The approach I implemented is displaying the unknown token in single quotes, but if it's not an ASCII character then it is also escaped and displayed in parenthesis.

### Before

```
error: Unknown start of token:
  |
1 | ﻿test:
  | ^
```

### After

```
error: Unknown start of token '' (\u{200b}):
  |
1 | ﻿test:
  | ^
```

### Before

```
error: Unknown start of token:
  |
1 | %﻿test:
  | ^
```

### After

```
error: Unknown start of token '%':
  |
1 | %﻿test:
  | ^
```

### DIFF
--- a/src/compile_error.rs
+++ b/src/compile_error.rs
@@ -291,7 +291,15 @@ impl Display for CompileError<'_> {
       }
       UnknownFunction { function } => write!(f, "Call to unknown function `{function}`"),
       UnknownSetting { setting } => write!(f, "Unknown setting `{setting}`"),
-      UnknownStartOfToken => write!(f, "Unknown start of token:"),
+      UnknownStartOfToken { token } => write!(
+        f,
+        "Unknown start of token {}:",
+        if token.is_ascii() {
+          format!("'{token}'")
+        } else {
+          format!("'{token}' ({})", token.escape_unicode())
+        }
+      ),
       UnpairedCarriageReturn => write!(f, "Unpaired carriage return"),
       UnterminatedBacktick => write!(f, "Unterminated backtick"),
       UnterminatedInterpolation => write!(f, "Unterminated interpolation"),
--- a/src/compile_error.rs
+++ b/src/compile_error.rs
@@ -291,15 +291,13 @@ impl Display for CompileError<'_> {
       }
       UnknownFunction { function } => write!(f, "Call to unknown function `{function}`"),
       UnknownSetting { setting } => write!(f, "Unknown setting `{setting}`"),
-      UnknownStartOfToken { token } => write!(
-        f,
-        "Unknown start of token {}:",
-        if token.is_ascii() {
-          format!("'{token}'")
-        } else {
-          format!("'{token}' ({})", token.escape_unicode())
+      UnknownStartOfToken { start } => {
+        write!(f, "Unknown start of token '{start}'")?;
+        if !start.is_ascii_graphic() {
+          write!(f, " (U+{:04X})", *start as u32)?;
         }
-      ),
+        write!(f, ":")
+      }
       UnpairedCarriageReturn => write!(f, "Unpaired carriage return"),
       UnterminatedBacktick => write!(f, "Unterminated backtick"),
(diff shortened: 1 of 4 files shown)


src/compile_error_kind.rs
```diff
@@ -153,7 +153,9 @@ pub(crate) enum CompileErrorKind<'src> {
   UnknownSetting {
     setting: &'src str,
   },
-  UnknownStartOfToken,
+  UnknownStartOfToken {
+    token: char,
```
casey: Perhaps call this `start`? Since it's not really a whole token. Also, `start` is the name of the variable in `lexer.rs`, so you can use shorthand initializer syntax there.

src/compile_error.rs
```diff
@@ -291,7 +291,15 @@ impl Display for CompileError<'_> {
...
       UnknownFunction { function } => write!(f, "Call to unknown function `{function}`"),
       UnknownSetting { setting } => write!(f, "Unknown setting `{setting}`"),
-      UnknownStartOfToken => write!(f, "Unknown start of token:"),
+      UnknownStartOfToken { token } => write!(
+        f,
+        "Unknown start of token {}:",
+        if token.is_ascii() {
+          format!("'{token}'")
+        } else {
+          format!("'{token}' ({})", token.escape_unicode())
```
casey: I think the `escape_unicode` format isn't super useful. Let's use the `U+NNNN` representation, which is commonly used to represent unicode codepoints. `NNNN` should be four hex digits. Leading zeros should not be elided.

casey: Also, let's use `char::is_ascii_graphic` as the test for whether to print the unicode codepoint. This will cause it to be printed for ASCII control characters, as well as ascii whitespace, which I think is probably good.

casey: Also, I think this should just be done with two calls to `write`:

```
write!(f, "Unknown start of token '{start}')?;
if !start.is_ascii_graphic() {
  // print (U+NNNN)
}
Ok(())
```

casey (changes requested):
Nice, see comments!

src/compile_error.rs
CramBL: Implemented. It needs 3 `write!` however, since there's a colon at the end

tests/misc.rs
```diff
@@ -1679,6 +1679,36 @@ assembly_source_files = %(wildcard src/arch/$(arch)/*.s)
    status:   EXIT_FAILURE,
 }
 
+test! {
```
casey: Sorry I didn't catch this before: The `test!` macro is kind of deprecated. I don't think it's very clear, and it's often inflexible, so anything custom often requires workarounds, so I try to use the `Test` struct and constructor for new tests (which the `test!` macro uses under the hood). So these should be converted to use the `Test` struct.

casey (changes requested):
Whoops, missed something. New tests should use the `Test` struct directly. Also, you can remove the `:` at the end of the error message. I'm not sure why it's there, and other messages don't have it.

tests/misc.rs
CramBL: I know you said for **new tests** but I took the liberty to convert the other tests that this PR touches as well.

casey: Nice, that's great. Maybe one day I'll get around to actually converting all the tests 😅 It strikes me that AI might actually be really good at this kind of tedious but complex refactoring.

casey:
Excellent, thank you!

PR #2645, opened 2025-02-27
Add pre-commit hook

sherbang:
Can we add a pre-commit hook for the new format parameter?

I original wrote this as:
```yaml
- id: just-format
  name: just-format
  description: "Runs just --unstable --fmt on changed files"
  entry: just --unstable --fmt -f 
  files: '\.?[jJ][uU][sS][tT][fF][iI][lL][eE]'
  language: rust
```
This has the benefit that it would auto-install just, and doesn't require it to be on the local system, but it will only work when there's one justfile in the repo since the just command won't allow multiple filenames to be passed in.

So instead, this PR uses a script runner to process each matching file individually, but depends on just to be already installed in the local environment.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+- id: just-format
+  name: just-format
+  description: "Runs just --unstable --fmt on changed files"
+  entry: just --unstable --fmt -f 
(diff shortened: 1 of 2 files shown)


casey:
I think that `just --fmt` is just not yet ready for primetime. It's unstable because a bunch of shortcomings, like removing certain kinds of comments. So I don't think I would recommend using it in a pre-commit hook.

PR #2659, opened 2025-03-07
In README.md adds package for adJ/OpenBSD

vtamara:
The package `just` is already part of adJ.

It was proposed to OpenBSD but it still has not been included: https://marc.info/?l=openbsd-ports&m=173410756516801&w=2


casey:
Thanks for the PR! I think this is just a little bit too niche to include, I hope you understand.

PR #2672, opened 2025-03-20
feat: module recipe dependency support

corvusrabus:
Support for recipe dependencies in a submodule

```justfile
mod foo

baz: foo::bar
```

Contributes to #2252

Still needs a bit of work.


corvusrabus:
Sorry. Took me a while to come back to this. I think it's ready for review now @casey

corvusrabus:
Hey @casey. Any chance you have time to have a look :) ?

casey:
Nice work! I refactored `resolve_recipe` to delegate to another function `resolve_dependency` which makes it a little cleaner and gets rid of some nesting, but otherwise looks great!

PR #2677, opened 2025-03-23
fix: add cygpath handling to support Windows paths in Cygwin/MSYS2 environments

constantin-curicheru:
- Dynamically locate and check for cygpath.exe
- Convert Windows-style paths to Unix-style paths when running in Cygwin or MSYS2
- Ensure compatibility by preserving original make/gmake logic
- Improve cross-platform usability when traversing directories and handling recipe arguments

### DIFF
--- a/contrib/just.sh
+++ b/contrib/just.sh
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+# locate cygpath if available
+CYGPATH=$(command -v cygpath 2>/dev/null)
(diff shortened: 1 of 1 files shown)


casey:
`just.sh` is more of a historical artifact than a useful program. It's `just` before it was rewritten in Rust. It doesn't have any tests, and I'm not sure many people use it, so I'm not planning on adding new features to it.

Issue #2314, opened 2024-08-21
Improvements to cross-platform path environment variable manipulation


W1M0R:
I find it helpful to run just recipes within a specific PATH environment. Here is an example:

```justfile
path_sep := if os_family() == 'windows' {
  ';'
} else {
  ':'
}

aqua_bin := join(env("AQUA_ROOT_DIR"), "bin")
proto_shims := join(env("PROTO_HOME"), "shims")
proto_bin := join(env("PROTO_HOME"), "bin")
local_gobin := clean(join(source_directory(), ".go/bin"))
tms_bin := clean(join(source_directory(), "tms/build/bin"))
tms_node_bin := clean(join(source_directory(), "tms/node_modules/.bin"))

export PATH := if os_family() == 'windows' {
  aqua_bin + path_sep + proto_shims + path_sep + proto_bin + path_sep + local_gobin+ path_sep + tms_bin + path_sep + tms_node_bin + path_sep + env("PATH")
} else {
  # on Linux we don't use aqua and proto
  local_gobin+ path_sep + tms_bin + path_sep + tms_node_bin + path_sep + env("PATH")
}

the-job:
  echo "executing in the just path environment"
```

This approach is quite convenient as it allows me to set the PATH on windows and linux without shelling out to powershell in windows and bash/sh on linux. This also ensures that all my recipes have access to the tools that I want them to have access to.

One side-effect of this approach that is not ideal, but also not problematic for my use case, is that calling one recipe from another recipe leads to duplication in the path environment.

It would be great if the syntax around this pattern could be improved.

Some might suggest using direnv, but direnv doesn't work on windows. And `just` is already doing so well in this front, that needing to introduce another tool for this task, seems unnecessary.

Inspired by direnv, one approach might be to have a built-in function (`env_add("PATH", "bla")` or `path_env_add("bla")`) that knows how to work with the PATH environment variable (this hides away operating system details like the path separator and whether we need to work with `PATH` or `Path` and avoiding duplicate entries, etc):

```justfile
aqua_bin := join(env("AQUA_ROOT_DIR"), "bin")
proto_shims := join(env("PROTO_HOME"), "shims")
proto_bin := join(env("PROTO_HOME"), "bin")
local_gobin := clean(join(source_directory(), ".go/bin"))
tms_bin := clean(join(source_directory(), "tms/build/bin"))
tms_node_bin := clean(join(source_directory(), "tms/node_modules/.bin"))

path_env_add(aqua_bin)
[...]

PR #2679, opened 2025-03-24
Add `COMPONENT_SEPARATOR` and `PATH_SEPARATOR` constants

casey:
See https://github.com/casey/just/issues/2314#issuecomment-2740579171.

@cspotcode, what do you think of the naming here? `/` is the COMPONENT_SEPARATOR, because it separates components, `;` is the PATH_SEPARATOR, because it separates paths.

I want to keep the names short. We could also go with `COMPONENT_SEP` and `PATH_SEP`, although that is pretty incomprehensible.


cspotcode:
Yeah, any of those options work for me: `COMPONENT_SEPARATOR`, `COMPONENT_SEP`, `DIR_SEP`.

casey:
I changed the name to `PATH_SEP` and `PATH_VAR_SEP`. Kinda wanna let this sit for a second in case anyone has any feedback.

PR #2680, opened 2025-03-25
Start noodling on cheat sheet

casey:
People like cheat sheets! It would be nice to have an official one. This adds `just --sheet` which prints out a self-contained HTML cheat sheet. We can also add it to the website at `just.systems/cheat-sheet/`.


mikeyhew:
I think this is a great idea but would suggest the more descriptive option name `--cheat-sheet`

casey:
@mikeyhew I think that's right. Just renamed it to --cheet-sheet.

casey:
Closing my stale PRs.

PR #2685, opened 2025-03-28
fix doc regarding doc keyword

tcrasset:
I think this is what is meant, given the example, and given testing I did locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1316,12 +1316,12 @@ Available recipes:
 The `[doc]` attribute can be used to set or suppress a recipe's doc comment:
 
 ```just
-# This comment won't appear
+# This comment will appear
 [doc('Build stuff')]
 build:
   ./bin/build
 
-# This one won't either
+# This one won't
 [doc]
 test:
   ./bin/test


casey:
I think these changes are incorrect. The comments refer to whether or not the comments (not the `[doc]` attribute) will appear in the output of `--list`.

README.md
```diff
@@ -1316,12 +1316,12 @@ Available recipes:
 The `[doc]` attribute can be used to set or suppress a recipe's doc comment:
 
 ```just
-# This comment won't appear
+# This comment will appear
 [doc('Build stuff')]
```
laniakea64, line 1320: The point of the original wording
```
# This comment won't appear
```
is to illustrate that the text of that comment ("This comment won't appear") is **not** the documentation comment for the recipe, since the doc comment is specified via the `[doc]` attribute.

If this wasn't clear enough from the original wording, could you instead suggest wording that more clearly conveys the original intent?

tcrasset:
@laniakea64 Oh I get it now. Indeed the wording threw me off. Though I don't have a particularly better way to explain it, now that I understand how it works, the doc is clear.

It may be the fact that the empty `doc` attribute being used to suppress the documentation is different from how many other tools do a "negation" e.g. `--no-coverage`  vs `--coverage`.
This codebase already has `no-cd`, maybe using `no-doc` might better convey the intent to way to suppress the documentation.

PR #2689, opened 2025-03-31
Use raw_arg on Windows to fix handling of spaces in commands

### LINKED ISSUE #1639: Quoting problems on Windows
I have the following script (reduced for brevity):

```just
set windows-shell := ["cmd.exe", "/c"]

clang := if `where clang 2>NUL || exit 0` == "" {
    error("cannot find clang compiler")
} else {
    trim(replace(`where clang`, "\\", "\\\\"))
}

test:
    {{clang}} --version
```

Running `just test` fails with the error message: `'C:\\Program' is not recognized as an internal or external command, operable program or batch file.`. This is expected as that is how cmd normally operates.

Changing `{{clang}}` to `{{quote(clang)}}` fails with the error message: `The filename, directory name, or volume label syntax is incorrect.`. This is because `quote()` uses single quotes which has a specific use case on Windows, rather than double quotes which works universally.

Changing `{{clang}}` to `"{{clang}}"` fails with the error message: `'\"C:\\Program Files\\LLVM\\bin\\clang.exe\"' is not recognized as an internal or external command, operable program or batch file.`. I'm not entirely sure what's going on here, but the most I could conclude is that the quotations from the command are being escaped for some reason.

With these issues in mind, what is the right way to go about running executables with spaces in the file path?

jhasse:
On Windows Rust escapes arguments for security reasons, but since we are passing an argument to a shell (i.e. can execute arbitrary code anyway), this should be disabled in order to correctly support quoting and spaces in commands.

https://doc.rust-lang.org/std/process/index.html#windows-argument-splitting

Fixes #1639.

### DIFF
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -282,7 +282,10 @@ impl<'src, D> Recipe<'src, D> {
       }
 
       #[cfg(target_os = "windows")]
-      std::os::windows::process::CommandExt::raw_arg(&mut cmd, command);
+      {
+        use std::os::windows::process::CommandExt;
+        cmd.raw_arg(command);
+      }
       #[cfg(not(target_os = "windows"))]
       cmd.arg(command);
 
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -281,6 +281,9 @@ impl<'src, D> Recipe<'src, D> {
         cmd.current_dir(working_directory);
       }
 
+      #[cfg(target_os = "windows")]
+      std::os::windows::process::CommandExt::raw_arg(&mut cmd, command);
+      #[cfg(not(target_os = "windows"))]
       cmd.arg(command);
 
       if self.takes_positional_arguments(&context.module.settings) {
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -284,7 +284,7 @@ impl<'src, D> Recipe<'src, D> {
       #[cfg(target_os = "windows")]
       {
         use std::os::windows::process::CommandExt;
-        cmd.raw_arg(command);
+        cmd.raw_arg(format!("\"{}\"", command.replace("\"", "\\\"")));
       }
       #[cfg(not(target_os = "windows"))]
       cmd.arg(command);
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -107,6 +107,8 @@ jobs:
 
     - name: Test
       run: cargo test --all
+      env:
+        RUST_BACKTRACE: '1'
 
     - name: Test install.sh
       run: |
--- a/tests/windows.rs
+++ b/tests/windows.rs
@@ -13,3 +13,20 @@ fn bare_bash_in_shebang() {
     .stdout("FOO\n")
     .run();
 }
+
+#[test]
+fn spaces_in_windows_shell_arg() {
+  Test::new()
+    .justfile(
+      "
+        set windows-shell := [\"cmd.exe\", \"/c\"]
+        default:
+            echo FOO && \"echo FOO\"
+      ",
+    )
+    .shell(false)
+    .stdout("FOO \n")
+    .stderr("echo FOO && \"echo FOO\"\n'\"echo FOO\"' is not recognized as an internal or external command,\noperable program or batch file.\nerror: Recipe `default` failed on line 3 with exit code 1\n")
+    .status(1)
+    .run();
+}


src/recipe.rs
```diff
@@ -281,6 +281,9 @@ impl<'src, D> Recipe<'src, D> {
         cmd.current_dir(working_directory);
       }
 
+      #[cfg(target_os = "windows")]
+      std::os::windows::process::CommandExt::raw_arg(&mut cmd, command);
+      #[cfg(not(target_os = "windows"))]
```
```suggestion
      #[cfg(target_os = "windows")]
      {
        use std::os::windows::process::CommandExt;
        cmd.raw_arg(command);
      }
      #[cfg(not(target_os = "windows"))]
```

casey (changes requested):
Nice, thank you for the PR!

1. Can you add a test for this? I want to make sure I understand the issue being fixed, and since it's subtle, make sure that there isn't a regression in the future.
2. Should we use this for all arguments? If not, why not?
3. See comments.

jhasse:
1. I'm not good at Rust, but I can try :) In which file would such a test go and is there a similar test I can look at?
2. Hm ... yes I think so, although I can't think of a scenario where it would matter. Maybe when someone has a shell with a parameter that needs spaces. I guess 99% of the time the shell parameter is `/c` or something similar though.

casey:
@jhasse The test can go in `tests/windows.rs`. I don't understand what the issue is myself, since windows argument handling seems extremely arcane to me, so I can't give much guidance about what an existing test to copy is. Check out any test that uses the `Test::new()` constructor to create a new test. If 99% of the time the shell parameter doesn't contain a space, that means 1% of the time it will break, which is too often ;) So we should definitely do whatever is the right thing.

jhasse:
I've tried to run the test in windows.rs but it failed:

```
› cargo test bare_bash_in_shebang
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.14s
     Running unittests src\lib.rs (target\debug\deps\just-601a7f99ed7e8a65.exe)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 499 filtered out; finished in 0.00s

     Running tests\lib.rs (target\debug\deps\integration-a255eaaa12765568.exe)

running 1 test
test windows::bare_bash_in_shebang ... FAILED

failures:

---- windows::bare_bash_in_shebang stdout ----
Bad status: Diff < left / right > :
 Some(
<    127,
>    0,
 )

Bad stdout: Diff < left / right > :
>FOO
>

Bad stderr: Diff < left / right > :
</bin/bash: C:UsersjhasseAppDataLocalTempjust-cUgCqDdefault: No such file or directory
<error: Recipe `default` failed with exit code 127
<


thread 'windows::bare_bash_in_shebang' panicked at tests\windows.rs:14:6:
Output mismatch.


failures:
    windows::bare_bash_in_shebang

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 885 filtered out; finished in 0.12s

error: test failed, to rerun pass `--test integration`
```

Is there anything I need to set up on Windows to run the tests? Do I need to install Cygwin?

casey:
Can you commit and push your failing test to the branch, so I can see it?

jhasse:
`bare_bash_in_shebang` was an existing test which I didn't touch, so there seems to be something fundamentally wrong with my setup.

I've pushed my attempt for a test for this issue.

casey:
Nice, thank you! I think the issue is that `&&` isn't valid syntax for `cmd.exe`, that's a bash-ism.

jhasse:
No, `&&` works the same way as in bash: https://superuser.com/a/1695562/114883

casey:
@jhasse TIL! Check out the test logs, I'm seeing some failures in the unit tests which I don't understand.

jhasse:
Hm ... me neither :(

I've tried setting RUST_BACKTRACE=1 to get a better understanding. No idea if that is the right way forward.

casey:
I think that in order to fix this, you'll need to be able to run the tests locally, otherwise the debug loop will be really slow if you have to run the tests on CI. Are you set up to do Windows Rust development? I have a Windows laptop I use for stuff like this, but I'm pretty busy so I don't know when I'll be able to get it out to take a look.

jhasse:
I have a Windows setup, but the test fail, even on the master branch (see https://github.com/casey/just/pull/2689#issuecomment-2780701876).

jhasse:
Okay after a bit of research and testing: It's basically `cmd.exe` that is the issue. It parses the command line flags differently. Rust assumes the standard C library parsing, which will break when using raw_arg. But cmd.exe would require it.

I guess the only real fix is to have a config option, so that raw_arg is enabled for cmd only.

I will use Git Bash instead ...

btw: Would be great if just would look at the default install path of Git bash on Windows (`C:/Program Files/Git/bin/bash.exe`).

PR #2692, opened 2025-04-02
Add support for XDG_CONFIG_HOME in justfile search paths

### LINKED ISSUE #2687: Error: Global justfile not found
`just` cannot find my global justfile. I believe this behavior is specific to macOS, as the same exact dotfile configuration works just fine in WSL on Windows, as well as in Ubuntu docker containers. My belief, based on [this comment](https://github.com/casey/just/issues/2434#issuecomment-2431713908) is that the dirs library just uses to find XDG directories explicitly ignores user specified environment variables like XDG_CONFIG_HOME.

### Steps to reproduce
1. Be in a shell on macOS
2. Set a custom directory in the environment variable called XDG_CONFIG_HOME, like `/Users/malikoth/.nonstandard_config_dir`
3. Place a justfile in a standard location relative to that non-standard XDG directory, such as `/Users/malikoth/.nonstandard_config_dir/just/justfile`
4. Run `just --global-justfile`

### Expected behavior
Due to the path matching the first example given in [the docs](https://just.systems/man/en/global-and-user-justfiles.html#global-justfile), `$XDG_CONFIG_HOME/just/justfile`, the `_default` recipe in the global justfile should be executed.

### Actual Behavior
Just emits the error message `error: Global justfile not found` and fails with a non-zero error code.

### Demonstrative samples from my command line
```
❯ uname -v
Darwin Kernel Version 21.6.0: Mon Jun 24 00:56:10 PDT 2024; root:xnu-8020.240.18.709.2~1/RELEASE_X86_64

❯ echo $XDG_CONFIG_HOME
/Users/malikoth/.local/etc

❯ la ~/.local/etc/just
total 4.0K
-rw-r--r-- 1 malikoth staff 2.3K Mar 28 21:30 justfile

❯ just -g
error: Global justfile not found
```

### Side Note (Just me complaining, lol)
(Why is .local used for most home directory stuff, except for config files? Why stuff those into ~/.config? That's never made sense to me! .local to me represents the user-scoped equivalent of all the directories in a Unix filesystem root, including etc, so I move it all into .local/etc. I have a shell init script explicitly setting all XDG_* env vars, and another script dedicated exclusively to setting env vars to "help" the apps that need a little more assistance dealing with that fact, but it's not too bad.)

pojknamn:
- Added support for checking `$XDG_CONFIG_HOME/just/justfile` path when looking for config files with global option
- add a few tests

Fixes #2687 

feedback welcome, this is my first time contributing rust

### DIFF
--- a/src/search.rs
+++ b/src/search.rs
@@ -395,4 +395,34 @@ mod tests {
       assert_eq!(have, Path::new(want));
     }
   }
+  #[test]
+  fn global_justfile_path_from_xdg_env(){
+    let old_value = std::env::var_os("XDG_CONFIG_HOME");
+
+    std::env::set_var("XDG_CONFIG_HOME", "/test/config");
+
+    let paths = Search::global_justfile_paths();
+    assert!(paths.contains(&PathBuf::from("/test/config/just/justfile")));
+
+    if let Some(old) = old_value {
+      std::env::set_var("XDG_CONFIG_HOME", old);
+    } else {
+      std::env::remove_var("XDG_CONFIG_HOME");
+    }
+  }
+  #[test]
+  fn global_justfile_path_when_xdg_env_not_set(){
+    let old_value = std::env::var_os("XDG_CONFIG_HOME");
+
+    std::env::set_var("XDG_CONFIG_HOME", "");
+
+    let paths = Search::global_justfile_paths();
+    assert!(!paths.contains(&PathBuf::from("/test/config/just/justfile")));
+
+    if let Some(old) = old_value {
+      std::env::set_var("XDG_CONFIG_HOME", old);
+    } else {
+      std::env::remove_var("XDG_CONFIG_HOME");
+    }
+  }
 }
--- a/src/search.rs
+++ b/src/search.rs
@@ -26,6 +26,14 @@ impl Search {
           .join(DEFAULT_JUSTFILE_NAME),
       );
 
+      if let Some(xdg_config_home) = std::env::var_os("XDG_CONFIG_HOME") {
+        paths.push(
+          PathBuf::from(xdg_config_home)
+              .join("just")
+              .join(DEFAULT_JUSTFILE_NAME)
+        );
+      };
+
       for justfile_name in JUSTFILE_NAMES {
         paths.push(home_dir.join(justfile_name));
       }


src/search.rs
```diff
@@ -26,6 +26,14 @@ impl Search {
           .join(DEFAULT_JUSTFILE_NAME),
       );
 
+      if let Some(xdg_config_home) = std::env::var_os("XDG_CONFIG_HOME") {
```
laniakea64, line 24: Is the goal to add the option of [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/latest/)-compliant path for global justfile, or only to use `XDG_CONFIG_HOME` literally as per documentation?

If the former, checking whether `XDG_CONFIG_HOME` is set is not the only condition to consider - see related discussion in https://github.com/casey/just/pull/2536

malikoth: So if I understand the conversation from the linked PR, it doesn't seem like getting full XDG-compliant paths is that hard. The only two concrete examples I could find were relative paths which are not allowed, and empty paths (the environment variable is set, but to an empty value). So you just take the contents of the environment variable, filter out any non-absolute or empty paths, and then default to ~/.config if there's nothing left. Did I get that correct?

Just my $0.02 as the issue submitter, I don't really care either way about this. I love rigorous standards, and it seems good to follow the standard, but if we're not too strict about it, then I can still make sure that the environment variable is set correctly on my end, and everything still works just fine for me. As long as there's something I can fix with my environment, then I can deal with whatever. I just need the documented example to work on all platforms. :)

And thanks for @laniakea64 for doing this, I was going to see about patching this, but I have never done anything in rust, and I hadn't gotten around to it yet. ❤️

laniakea64: > So if I understand the conversation from the linked PR, it doesn't seem like getting full XDG-compliant paths is that hard. ... So you just take the contents of the environment variable, filter out any non-absolute or empty paths, and then default to ~/.config if there's nothing left. Did I get that correct?

Yes, you are correct about full XDG spec compliance.

> And thanks for @laniakea64 for doing this, I was going to see about patching this, but I have never done anything in rust, and I hadn't gotten around to it yet. ❤️

:confused: I wasn't involved in authoring this PR

malikoth: My bad, lol... Thanks to @pojknamn for writing this!!! ❤️ You're a hero!

malikoth:
Obviously this needs to rebased onto latest code, but other than that, is this something you'd merge, @casey? I haven't actually needed this for a while as I've primarily been using Windows for a bit, but now I'm back on a mac, and would dearly appreciate having this fix merged.

casey:

Issue #2583, opened 2025-01-17
Catch Functionality


ArchieAtkinson:
There is no current way in standard Just to run additional commands if the execution of a recipe fails to allow clean-up. This is available in other languages with keywords like `catch`. This type of feature is useful when you have a multistage process that produces an output like building and running a container or server. 

The current best method is to use the `trap` shell command which forces you to make your recipe a shell script. 
```
test-in-container:
    #!/usr/bin/env sh
    trap 'docker rm -f test-container' EXIT
    docker run -d --name test-container test-image
    docker exec test-container run-tests
```

Another current solution is using an `||`:
```
test-in-container:
    docker run -d --name test-container test-image
    docker exec test-container run-tests || (docker rm -f test-container && exit 1)
    docker rm -f test-container
```

However, this only works on a single line and gets repetitive for longer recipes.

Below I've thought of some possible ways this could be implemented in Just while maintaining backwards compatibility: 

Using an attribute:
```
[catch: docker rm -f test-container]
test-in-container:
    docker run -d --name test-container test-image
    docker exec test-container run-tests
```
Where the command after `catch:` would run if the recipe errored.

Using a keyword:
```
test-in-container:
    docker run -d --name test-container test-image
    docker exec test-container run-tests
    ::catch::
   docker rm -f test-container
```
Where all the commands after `::catch::` would run if the recipe errored.

Extend the dependency system:
```
test-in-container-cleanup:
    docker rm -f test-container

test-in-container: !& test-in-container-cleanup
    docker run -d --name test-container test-image
    docker exec test-container run-tests
```

Where recipes after the`!&` would run if the recipe errored.

The attribute method would likely be the most straightforward to implement, while extending the dependency system with `!&` does seem more elegant. However, I can think of some potential complications already, which might be tricky to solve. 

This could also be extended with a `finally` type feature that would run at the end of a recipe error or not. 

While I acknowledge this functionality is already available through other methods, and adding it might introduce unnecessary complexity, I believe it would be a useful addition that could help keep more recipes within the standard Just framework.

I'm curious about everyone's opinions on this, and if there's interest, I'd be willing to attempt to create an implementation.

casey:
I've had a `||` syntax in  mind for dependencies that should run if a recipe fails:

```
foo: || bar
```

`bar` would only run if `foo` terminates unsuccessfully.

But it sounds like this is for dependencies that run regardless of whether a recipe succeeds or fails?

[...]

PR #2709, opened 2025-04-14
Recovery Dependencies - Execute Receipe(s) on Recipe Failure 

ArchieAtkinson:
This PR adds the `||` syntax to the dependency system, allowing recipes to run on failure. 

This is based on discussion in https://github.com/casey/just/issues/2583

Example:
```
# justfile
foo: || bar
  echo foo
   exit 1

bar:
  echo bar

# stdout
foo
bar
```

Please let me know if any additional work needs to be done or changes to the current behaviour.

One potential improvement that could be considered is passing the status code from the failing recipe into the recovery recipes

### DIFF
--- a/src/justfile.rs
+++ b/src/justfile.rs
@@ -219,9 +219,7 @@ impl<'src> Justfile<'src> {
 
       let mut evaluator = Evaluator::new(&context, true, &scope);
       if let Err(err) = result {
-        if invocation.recipe.recoveries().peekable().peek().is_none()
-          || context.config.no_dependencies
-        {
+        if invocation.recipe.recoveries().next().is_none() || context.config.no_dependencies {
           return Err(err);
         }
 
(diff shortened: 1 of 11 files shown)


casey:
Thank you for the PR! Just a warning, I'm super backlogged, so I'm not sure when I'll be able to review this. Just letting you know so that you don't interpret no response as the PR not being welcome or interesting!

casey (changes requested):
Thanks for your patience!

Please ignore the copilot review. I was curious what it would say.

Correct me if I'm wrong, but I believe there is a problem here: `||` recipes will only run if the recipe that failed was giving on the command line, but not if the recipe was a dependency of another recipe.

So if we have:

```just
a: || recovery
  exit 1

b: || recovery
  exit 1

c: b

recovery:
```

Then `just a` will run `recovery`, but `just c` will not run `recovery`.

I think the fix is to put the recovery code into `Justfile::run_recipe`, where it will execute for all recipes, not just top-level command-line invocations.

casey:

PR #2711, opened 2025-04-19
zinit dcs

iloveitaly:
link to example zinit installation

### DIFF
--- a/README.md
+++ b/README.md
@@ -4119,6 +4119,9 @@ fpath=($HOMEBREW_PREFIX/share/zsh/site-functions $fpath)
 # compinit
 ```
 
+*zinit Note:* If you are using zinit, [here's example](https://github.com/iloveitaly/dotfiles/blob/a18fd4b0745877cfa03de2736caa39af49525afa/.zsh_plugins#L51) of how to install `just`
+shell completions easily.
+
 ### Man Page
 
 `just` can print its own man page with `just --man`. Man pages are written in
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4520,7 +4523,7 @@ dependency, or running some kind of codegen step.
 `just`, on the other hand, is for all the other miscellaneous commands you
 might run as part of development. Things like running tests in different
 configurations, linting your code, pushing build artifacts to a server,
-removing temporary files, and the like.
+removing temporary files, and the like.pppppp
```
vaibhavdaren, line 4526: I think this line was accidentally added to the commit.

vaibhavdaren (changes requested):
Remove accidental changes.

README.md
```diff
@@ -4119,6 +4119,9 @@ fpath=($HOMEBREW_PREFIX/share/zsh/site-functions $fpath)
 # compinit
 ```
 
+*zinit Note:* If you are using zinit, [here's example](https://github.com/iloveitaly/dotfiles/blob/a18fd4b0745877cfa03de2736caa39af49525afa/.zsh_plugins#L51) of how to install `just`
+shell completions easily.
+
```
laniakea64, line 4124: That link seems like a lot of indirection for [this code](https://gist.github.com/iloveitaly/9103604f442e0474dd18b4acd596739a/), which seems simple enough to copy-paste directly in the README?  I don't use `zsh` myself, although I do have a very similar snippet in my `bashrc`, am I missing some context to understand a purpose of the layers of indirection in `zsh` config?

iloveitaly: not sure i follow. There is a single zinit plugin here:

https://gist.githubusercontent.com/iloveitaly/9103604f442e0474dd18b4acd596739a/raw/just.plugin.zsh

casey:
The plugin is very short, so I think there isn't a strong advantage towards using it as a plugin, as opposed to just putting it into your zsh init file directly.

PR #2714, opened 2025-04-20
Add uv instructions to README & fix issue with pipx instructions.

jamesturk:
Hi,

Given the increasing usage of uv over pipx/conda, I wanted to provide alternate instructions for uv users. In doing so I also changed the "Package Manager" section for the existing pipx entry to avoid confusing since both use PyPI as a directory.

Glad to make any desired changes & thank you for just and for providing a rust-just package on PyPI!

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ most Windows users.)
       <td><code>npm install -g rust-just</code></td>
     </tr>
     <tr>
-      <td><a href=https://pypi.org/>PyPI</a></td>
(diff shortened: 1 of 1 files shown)


casey:
Sorry for letting this set! I merged the link to `pipx` in #2799. I'm not sure about adding the instructions for `uv`, since for CLI tools, I'm not sure if `uv` has strong advantages over `pipx`.

Issue #2442, opened 2024-10-29
Inline modules


iloveitaly:
From [this discord conversation.](https://discord.com/channels/695580069837406228/695580069837406231/1300919273270673518)

I have a bunch of related tasks:

```
    db-generate-migration
    db-migrate
    db-nuke
    db-reset
    db-seed
    db-up
    db-destroy
```

What I'd like to do is have a more clear separator (like rake):

```
    db:generate-migration
    db:migrate
```

This is possible with modules, but I cannot create multiple modules in a single Justfile (creating separate Justfiles would be messy in my case).

Two ideas I can think of to solve this problem:

1. Would it be possible to support `:`, `.` in the name grammar?
2. Could we define multiple modules within a single Justfile?

casey:
Thanks for opening this issue!

I would like to eventually support inline modules, which would allow submodules to be defined in a single justfile.

We could allow `::` in recipe names, but modules do more than just provide namespaces for recipes, since they can have their own settings, variables, and submodules, so I'm more keen to just do full blown inline modules, which would support all of that.

The syntax I've had in mind is:

```
bar::
  foo:
    echo 'recipe foo inside module bar'
```

But it could also be:

```
mod bar::
  foo:
    echo 'recipe foo inside module bar'
```

`just` uses indentation to delimit recipe bodies, since recipe bodies are arbitrary text, and `just` is descended from `make`, which does the same. So it would be natural to also use indentation to delimit modules bodies, as in the examples above.

However, I do personally find indentation to be annoying and fiddly, so we could consider using `{…}` for modules, which would allow for indenting the contents:

```
bar {
  foo:
    echo 'recipe foo inside module bar'
}
```

Or not, depending on preference:

```
bar {
foo:
  echo 'recipe foo inside module bar'
}
```

iloveitaly:
I _really_ like the `::` indentation structure:

```
bar::
  foo:
    echo 'recipe foo inside module bar'
```

Feels really natural and easy to read.

Nested indentation isn't awesome, but `{` syntax wouldn't seem to align with the rest of the Justfile grammar IMHO. However, this is a departure from Makefile syntax in a bigger way so maybe it makes sense to do something very different?

Either way sounds great to me! I love this idea.

casey:
[...]

PR #2715, opened 2025-04-20
feat: inline modules

corvusrabus:
Draft for non-nested inline modules according to the syntax mainly agreed upon in #2442.

Syntax:

```justfile
# foo is an inline module!
foo::
  hello:
    @echo "Hello from inside the inline foo module"
```

Contributes to #2252.

I think there's still a couple of open questions that need to be clarified before something like this PR can be merged
- Should the entire module code depending on filepaths be refactored to accommodate for inline modules that don't directly correspond to a file.
- what should `module_directory`  and `module_directory` do when used inside an inline module? Should probably just fail?
- Probably more

### DIFF
--- a/README.md
+++ b/README.md
@@ -3733,6 +3733,15 @@ Available recipes:
     foo ... # foo is a great module!
 ```
 
+Modules can be defined inline. Currently inline modules can not contain further modules or inline modules.
+
+```justfile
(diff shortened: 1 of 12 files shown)


musjj:
Hey, are you still working on this?

Issue #857, opened 2021-06-06
Enum variables


remimimimimi:
Currently `just` have only environment variables, required and optional recipe arguments.

I suggest add enum-like arguments for recipes for those tasks which depends only on certain number of variants for parameter.
This feature highly required for convinient build systems in operating system development and other tasks. 

We need at least two functions and special syntax:
 * `one_of_or_default(enum, default)` - will use default argument if there's no input.
 * `one_of(enum)` - will panic and give possible variants if there's no input.
 
```make
set runners := [
    # input = output
   "qemu" = "qemu",
   "bochs" = "bochs"
]
set qemu_arch_postfix := [
    "x86_64" = "-system-x86_64",
    "riscv64" = "-system-riscv64",
]
run run_cmd=one_of(runners) qemu_postfix=one_of_or_default(qemu_arch_postfix, ""):
    {{ if run_cmd == "qemu" { "qemu" + qemu_postfix } else { bochs } }} /path/to/kernel
```
Also every enum may be callable, so we can extract values by variant.
Don't think that this is particularly necessary, but who knows ¯\_(ツ)_/¯
```make
set answers := [
    "not_equal" = "xyz",
    "equal" = "abc"
]
foo := if "hello" != "goodbye" { answers("not_equal") } else { answers("equal") }

bar:
  @echo {{foo}}
```

casey:
I definitely see the need for this, and like the idea. Supporting this well would require adding a real type system, see #528.

double-daruma:
Personally I think this (and #528 ) could really be the killer feature of `just`. I've written so many bash scripts that are 5 lines of actual logic and the 70 other lines are argument validation and error messages. If one can simply define which arguments are valid for a `just` recipe with enums, simple types and maybe regexes and number ranges then `just` would provide incredible value for writing project scripts beyond just being a convince. 

Now that I think about it, what if writing a `just` target would conceptually be a lot like writing a cli with `clap`-derives? I don’t mean syntactically the same but the nice thing about writing a cli in Rust with clap is, that you can just declare what types you want your arguments to be and some constraints that need to hold, and then you can focus write your logic.

casey:
In #3000 I added the ability to constrain argument values with regular expression patterns:

```just
[arg('n', pattern='\d+')]
double n:
  echo $(({{n}} * 2))
```

This can be used to create enum-like alternatives using `|`:

```just
[arg('arg', pattern='bar|baz')]
foo arg:
```

[...]

PR #2721, opened 2025-04-25
Support for enums in parameters

zbrox:
Hey, 

This is preliminary work to add support for enums. The syntax is similar to what is suggested in #857 with the difference that instead of `set` it uses a new keyword `enum`.

I wasn't sure of the direction this should take, so any comments are very much welcome. I just took a quick stab at it, since I also constantly need enums in my justfiles :)

Currently what is missing is tests and the `Node` impl (currently a `todo!()` since I wasn't sure what it should be exactly).


casey:
Could you add a test or example which demonstrates the syntax? I'd like to see how it looks.

woile:
This looks really awesome! Really looking forward to it. This would definitely help creating super robust `justfiles`!

Might be worth evaluating if the naming is proper. Is this actually an `enum` or more like a `struct`, `set`, `hashmap` or `dict`? As the name might create certain expectations, and if it's implemented as in the example of #857 , it's not an `enum` per se, because it has key and value. I don't have any preference, I think if users expectation don't align with the implementation, the maintainers may see a raise in issues.

Some other names could be "choices" or "options".

zbrox:
> Could you add a test or example which demonstrates the syntax? I'd like to see how it looks.

Sorry, I have missed that. It is a bit overwhelming around me the last couple of weeks but I will do this first thing when I have some free time.

zbrox:
Sorry for the huge delay!

I have added a couple of tests and finished the current implementations because there were some issues. 

Here's a sample syntax on how to declare an enum:

```justfile
enum foo := [ 
  "key1" = "value1",
  "key2" = "value2"
]
```

And then to use it here are two sample recipes used in the tests:

```justfile
test-enum p1=one-of(foo):
        @echo {{ p1 }}

test-enum-default p1=one-of-or-default(foo, "c"):
        @echo {{ p1 }}
```

1. If calling `just test-enum key1` you would see `value1`. 
2. In case you try to pass a key not in the enum, you would get an error that the variant is invalid and a list of valid ones.
3. If you call `just test-enum-default` you would see `c`

The syntax with the dashes (instead of underscores) was chosen because it was easier to implement parsing for (multiword keywords are kebab case). But otherwise I don't have a solid opinion about it, eventually I kinda got used to it as well since it is different from functions.

Feedback will be greatly appreciated!

zbrox:
@casey did you manage to take a look at this? any feedback?

casey:
Hey, sorry for taking so long to look at this. I spent a few months not really working on `just` at all.

I decided to go another direction in providing this feature, implemented in #3000, which adds the ability to constrain argument values with regular expression patterns:

```just
[arg('n', pattern='\d+')]
double n:
  echo $(({{n}} * 2))
```

This can be used to create enum-like alternatives using `|`:

```just
[arg('arg', pattern='bar|baz')]
foo arg:
```

This is more flexible than enums, since you can both match a alternatives, like `foo|bar`, or general patterns, like numbers, IP addresses, URLs, etc.

It does not provide the ability to map enum names to enum values, but I think that's best tackled in an alternative PR, and I'm not sure the proposed enum syntax is actually the right way to go for that.

I think the other issue with this approach, is that it adds more complexity to the recipe argument syntax, which is already pretty crazy. There's already exported vs unexported, with and without a default, and singular vs both flavors of variadic. Extending that with the special purpose `one-of` and `one-of-or-default` syntax seemed like it would make something that was already complex more complex.

In addition, adding the ability to pass arguments as flags `just recipe --arg` is another coveted feature, and I've been thinking about how to provide that too. I've always thought that it would use a syntax like:

```just
recipe --arg:
  echo {{arg}}
```

But that also suffers from the problem of making the recipe parameter syntax more complicated.

I came up with the idea of using `arg` attribute with keyword arguments to provide a regex pattern, and I thought that it would also be useful to add a bunch of other features, like flags, in a relatively non-crazy way:

```just
# `just foo --arg value`
[arg('arg', long='arg')]
foo arg:

# `just foo -a value`
[arg('arg', short='a')]
foo arg:

# if `just foo --arg` `arg` is `yes`, if `just foo` `arg` `no`
[arg('arg', long='arg', value='yes')]
foo arg='no':

# provide a help string
[arg('arg', help='this is a great argument')]
foo arg:
```

So that seemed like the way to go, since it allowed for adding some other features in the future. `just` honestly already has a pretty cursed grammar, so it's extra important not to let it get more cursed.

So sorry to close this. I should really do a better job of indicating if I think something is ready to be implemented in issues. A lot of time I say something like "I like this idea", or something like that, but when I sit down to think about it further, or try to implement it myself, it turns out to have complications that I didn't think of, or better alternatives.

zbrox:
Not a problem, it totally makes sense. I understand that you wanna keep the grammar from mutating more

PR #2734, opened 2025-05-14
fix formatting of attributes on modules

tjemke:
First off, love the tool and that it comes with a formatter built in :)

However, I found that while setting attributes on `mod` statements works perfectly, the formatter apparently is not aware of this feature. 
I just added a fairly naive implementation adding the two currently allowed attributes to the module `impl Display`. Also comes with a simple integration test.

I know that the formatter is an unstable feature and not the highest priority, but I personally use it and would love to see this fixed. Let me know if any of this needs fixing up, I'll happily do it :)

### DIFF
--- a/src/item.rs
+++ b/src/item.rs
@@ -45,11 +45,21 @@ impl Display for Item<'_> {
         write!(f, " {relative}")
       }
       Self::Module {
+        doc,
+        groups,
         name,
         relative,
         optional,
         ..
       } => {
+        if let Some(docstr) = doc {
+          writeln!(f, "# {docstr}")?;
+        }
+
+        for group in groups.iter() {
+          writeln!(f, "[group('{}')]", group)?;
+        }
+
         write!(f, "mod")?;
 
         if *optional {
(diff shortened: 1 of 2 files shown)


src/item.rs
```diff
@@ -45,11 +45,21 @@ impl Display for Item<'_> {
...
         optional,
         ..
       } => {
+        if let Some(docstr) = doc {
+          writeln!(f, "# {docstr}")?;
+        }
+
+        if !groups.is_empty() {
+          writeln!(f, "[group('{}')]", groups.first().unwrap())?;
+        }
```
casey: Since a module can belong to multiple groups, this should print out all the groups, not just the first.

casey (changes requested):
Modules may have multiple groups, so we should print them all, not just the first. See comment!

src/item.rs
tjemke: Don't really know how I overlooked that part, thank you. Should be fixed now and I also updated the test accordingly.

casey:
I see this PR has a couple of commits which were added after my review. Is it ready for re-review? I ignore notifications of new commits, since they're noisy and may not indicate that a branch is ready for review, so please comment or hit the request review button.

tjemke:
This was an attempt to fix the same bug that you then later on fixed with #2930, so it's now obsolete. Thank you for your work on this :)

PR #2758, opened 2025-05-30
Add mise tasks as alternative

koppor:
I think [mise tasks](https://mise.jdx.dev/tasks/) are worth mentioning as alternative: `mise` also manages environements and could fit more in some contexts.


casey:
Thank you!

Issue #2763, opened 2025-06-04
Recursive just breaks with remote justfile


chicks-net:
# Setup

I was running `just -d . -f ~/Documents/git/template-repo/justfile merge` with
[remote justfile](https://github.com/fini-net/template-repo/blob/main/justfile).  I had run `just -d . -f ~/Documents/git/template-repo/justfile pr` without any errors, but the `merge` had an error message:

```
just sync # mostly redundant, but just in case
error: Justfile does not contain recipe `sync`
error: Recipe `merge` failed on line 63 with exit code 1
```

So the `just sync` that `just merge` tried to run could not be found.  This makes sense, but it is too bad that the `just sync` couldn't know about the options that were passed to the `just merge`.

# Questions

1. Is there a variable that could be passed to inner calls of `just` that would pass along the options from the original call to `just`?
2. Could/does `just` handle this cleanly when not in a shebang script context?

philipmgrant:
Not a direct answer to your question, but in the specific case you have here, could you replace the recursive `just sync` call with a [post-dependency](https://just.systems/man/en/dependencies.html#running-recipes-at-the-end-of-a-recipe)?

So something like:

```
merge: _on_a_branch && sync
   # do stuff
```

---

More generally (i.e. including situations where recursive `just` isn't the last line of the recipe), can you use the `JUST_JUSTFILE` env var instead of the `-f` command line switch to set the justfile path? That should propagate to child processes and ensure that recursive `just` invocations use that file.

In fact, most command line options can be replaced with environment variables: see `man just` or `just --help` to see which env vars to use.

chicks-net:
> Not a direct answer to your question, but in the specific case you have here, could you replace the recursive `just sync` call with a [post-dependency](https://just.systems/man/en/dependencies.html#running-recipes-at-the-end-of-a-recipe)?
> 
> So something like:
> 
> ```
> merge: _on_a_branch && sync
>    # do stuff
> ```

Yes, yes.  I tested this in https://github.com/fini-net/template-repo/pull/10 and it totally works for me.

This is a feature of `just` that I had not played with at all, so thank yo for pointing it out.  😁 

> More generally (i.e. including situations where recursive `just` isn't the last line of the recipe), can you use the `JUST_JUSTFILE` env var instead of the `-f` command line switch to set the justfile path? That should propagate to child processes and ensure that recursive `just` invocations use that file.
> 
[...]

PR #2783, opened 2025-06-19
Update README.md (mention that environment variables are inherited by recursive invocations of just, unlike command-line options)

philipmgrant:
Points out the difference between command-line options and the corresponding environment variables, as regards the behaviour of recursive just calls.

On one level, this is just "what environment variables do". But #2763 illustrates that it's useful to mention it explicitly.


casey:
Good addition! I did think the example was a little long, and not necessarily clearer than the explanation, so I removed it.

PR #2793, opened 2025-06-25
Update which dev-dependency from 7.0.0 to 8.0.0

musicinmybrain:
https://github.com/harryfei/which-rs/releases/tag/8.0.0

I confirmed that `cargo test` still passes.


casey:
I always bump dependencies before releasing a new version, so unless there's something wrong with a dependency, there's no need to update them between releases.

Thank you for the PR though!

PR #2794, opened 2025-06-26
Add `[metadata]` attribute for adding custom metadata to recipes

### LINKED ISSUE #2703: Custom attributes / metadata attributes
Hi,

I'm using the fantastic dump command (`just --dump --dump-format json`) to build a CI pipeline dynamically, based on what's in the `justfile` and it works great but for some tasks I need more static metadata (e.g. recipe may need a different CI image etc).

I've (ab)used `[group]` attribute with a custom format of the group name, that is then parsed by the pipeline generator but it would be good to have either a way to use custom attributes or at least a `[meta('x', 'y')]` attribute that I can parse with my tooling.

In case it helps the justfile is at: https://gitlab.archlinux.org/archlinux/wkd-exporter/-/blob/main/.justfile and the example recipe looks like this:

```just
# Build docs
[group('ci')]
[group('pacman:rust')]
doc:
    RUSTDOCFLAGS='-D warnings' cargo doc --no-deps
```

I've considered abusing recipe doc comment but that'd also be a workaround. If there are any better ideas I'm all ears.

Thanks for your time!

wiktor-k:
This PR adds `metadata` attribute, which can appear multiple times on a recipe. The attribute takes a list of strings and its value can be inspected via JSON dump of the `justfile`.

The primary motivation is custom tooling built around `just` that can be used to, for example, automate creation of CI pipelines and other pre-processing that is outside of `just`'s scope.

Fixes: https://github.com/casey/just/issues/2703

### DIFF
--- a/README.md
+++ b/README.md
@@ -2093,12 +2093,13 @@ change their behavior.
 | Name | Type | Description |
 |------|------|-------------|
 | `[confirm]`<sup>1.17.0</sup> | recipe | Require confirmation prior to executing recipe. |
-| `[confirm('PROMPT')]`<sup>1.23.0</sup> | recipe | Require confirmation prior to executing recipe with a custom prompt. |
-| `[doc('DOC')]`<sup>1.27.0</sup> | module, recipe | Set recipe or module's [documentation comment](#documentation-comments) to `DOC`. |
-| `[extension('EXT')]`<sup>1.32.0</sup> | recipe | Set shebang recipe script's file extension to `EXT`. `EXT` should include a period if one is desired. |
-| `[group('NAME')]`<sup>1.27.0</sup> | module, recipe | Put recipe or module in in [group](#groups) `NAME`. |
+| `[confirm(PROMPT)]`<sup>1.23.0</sup> | recipe | Require confirmation prior to executing recipe with a custom prompt. |
+| `[doc(DOC)]`<sup>1.27.0</sup> | module, recipe | Set recipe or module's [documentation comment](#documentation-comments) to `DOC`. |
+| `[extension(EXT)]`<sup>1.32.0</sup> | recipe | Set shebang recipe script's file extension to `EXT`. `EXT` should include a period if one is desired. |
+| `[group(NAME)]`<sup>1.27.0</sup> | module, recipe | Put recipe or module in in [group](#groups) `NAME`. |
 | `[linux]`<sup>1.8.0</sup> | recipe | Enable recipe on Linux. |
 | `[macos]`<sup>1.8.0</sup> | recipe | Enable recipe on MacOS. |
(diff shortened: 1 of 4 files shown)


casey:
This mostly looks fine to me! However, I have some questions about the design:

- What's the use-case for more than one metadata attribute?
- Should we allow the metadata attribute to take a list of strings, instead of a single string? I think the main reason to do this is that it's more compact than multiple metadata attributes, since it's one line instead of multiple lines. We should think about this now, since it's a breaking change for JSON deserializers if metadata changes from a string to a list of strings.

wiktor-k:
> This mostly looks fine to me! However, I have some questions about the design:
> What's the use-case for more than one metadata attribute?

I wonder if you're asking about metadata being repeatable or a more generic question about attaching multiple metadata items per recipe. If it's the latter I envisioned multiple tools, each with their own metadata, which can at the same time use `just` so there needs to be some way to specify multiple items. The other approach would be to have the `metadata` attribute take multiple strings and expose that via JSON as a `metadata` array.

(As a practical example, I'm currently parsing [pragma-like comments from recipes](https://gitlab.archlinux.org/archlinux/wkd-exporter/-/blob/main/.justfile?ref_type=heads#L42) using [JavaScript](https://gitlab.archlinux.org/archlinux/wkd-exporter/-/blob/main/scripts/ci/create-gitlab-checks.js?ref_type=heads#L15) and there are multiple different ones in one recipe).

I initially chose "repeatable" as I had key-value structure for metadata in mind (where each external tool would use their different key and they could coexist) but stopped half-way as I was wondering if this is a bit too complex.

> Should we allow the metadata attribute to take a list of strings, instead of a single string? I think the main reason to do this is that it's more compact than multiple metadata attributes, since it's one line instead of multiple lines.

I think it's a reasonable approach and I'm happy to adjust the code if we settle on this.

> We should think about this now, since it's a breaking change for JSON deserializers if metadata changes from a string to a list of strings.

Totally agreed. The PR didn't take too much time and I think having something concrete is a great conversation starter.

Thanks for your time! :wave:

casey:
Let's allow a list of strings, as well as allowing the metadata attribute to appear multiple times. I don't see any issue with it, and it allows metadata authors flexibility in how they'd like to structure their metadata.

wiktor-k:
> I don't see any issue with it, and it allows metadata authors flexibility in how they'd like to structure their metadata.

Good idea.

I've left the constraint to require at least one argument in the `[metadata]` attribute as it seemed weird to have an argumentless node like that.

The code could be restructured a bit because structure-wise what `[metadata]` takes looks identical to the `Interpreter` struct (one required "command" and a set of arguments). I've kept the code simple for now, unless you see value in adjustments.

casey:
LGTM!

PR #2798, opened 2025-06-29
Add `--tempdir` command-line option

casey:


W1M0R:
Thanks @casey, this looks great.

qnerden-ifx:
I'm getting a weird behavior when using `~`
```bash
just --tempdir=~/tmp foo
error: Recipe `foo` could not be run because of an IO error while trying to create a temporary directory or write a file to that directory: No such file or directory (os error 2) at path "/home/myredacteduser/dev/~/tmp/just-3FlApn"
```

But this works:
```
$ mkdir tmp
$ just --tempdir=tmp foo
```

kenden:
@casey  I'm getting a weird behavior when using `~`
```bash
just --tempdir=~/tmp foo
error: Recipe `foo` could not be run because of an IO error while trying to create a temporary directory or write a file to that directory: No such file or directory (os error 2) at path "/home/myredacteduser/dev/~/tmp/just-3FlApn"
```

But this works:
```
$ mkdir tmp
$ just --tempdir=tmp foo
```

Maybe `--tempdir` should take an absolute path.

casey:
@kenden `~` is expanded by the shell, and is only expanded if it is the first character of a whitespace-delimited word. Try this:

```
just --tempdir ~/tmp foo
```

kenden:
@casey thanks, I didn't know `~` / tilde is only expanded if it is the first character of a whitespace-delimited word.
It works 👍

PR #2804, opened 2025-07-05
Resolve cygpath from env `JUST_CYGPATH`

### LINKED ISSUE #2641: Search cygpath.exe from Git's install path
I was trying the [Shebang Recipes](https://just.systems/man/en/shebang-recipes.html) function and got this error:

```cmd
C:\Users\hyrious\Desktop\Sakuna>just ruby
error: Could not find `cygpath` executable to translate recipe `ruby` shebang interpreter path:
program not found
```

This was because I do not have `cygpath` in my PATH. I do have cygpath.exe in the installation directory of Git and MSYS2, but they are not added to PATH for not polluting my normal shell. For example, PowerShell would prefer `echo.exe` (from bash) over the `echo` command, they behave differently obviously.

While the most complete solution may be quite complex regarding to #1549 and #2599. I have a more trivial feature request that, `just` can search for `cygpath.exe` from common paths like a relative path to Git:

```cmd
C:\Users\hyrious\Desktop\Sakuna>C:\progra~1\Git\usr\bin\cygpath --version
cygpath (cygwin) 3.4.10
Path Conversion Utility
Copyright (C) 1998 - 2024 Cygwin Authors
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

C:\Users\hyrious\Desktop\Sakuna>C:\progra~1\Git\usr\bin\cygpath ruby
ruby
```

The basic logic is as follows:

1. See if `C:\progra~1\Git\usr\bin\cygpath.exe` exists. This is the default location that Git for Windows will be installed at. It is highly possible that there's a `cygpath.exe`.
2. If not, search for `git.exe` by `where git` or simply iterating over the PATH environment.
3. If `git` exists, resolve the `cygpath` by `join(git_path, '../../usr/bin/cygpath.exe')`.

   ```js
   > fs.existsSync(path.join(String.raw`C:\Progra~1\Git\cmd\git.exe`, '../../usr/bin/cygpath.exe'))
   true
   ```

You're free to choose not to implement that.

hyrious:
I couldn't pass all the tests from my Windows setup because it seems to rely on a real `sh` program.

I did verify the env works:

```
> cargo b
> target\debug\just.exe --shell "C:\Program Files\Git\usr\bin\sh.exe" _ruby
Hello from ruby!
```

...where otherwise it prints:

```
error: Could not find `cygpath` executable to translate recipe `_ruby` shebang interpreter path:
program not found
```

Fixes #2641

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -209,7 +209,7 @@ impl Config {
           .action(ArgAction::Set)
           .value_parser(value_parser!(PathBuf))
           .default_value("cygpath")
-          .help("Use binary at <CYGPATH> convert between unix and Windows paths."),
+          .help("Use binary at <CYGPATH> to convert between unix and Windows paths."),
       )
       .arg(
         Arg::new(arg::DOTENV_FILENAME)
--- a/src/config.rs
+++ b/src/config.rs
@@ -17,7 +17,7 @@ pub(crate) struct Config {
   pub(crate) check: bool,
   pub(crate) color: Color,
   pub(crate) command_color: Option<ansi_term::Color>,
-  pub(crate) cygpath: Option<PathBuf>,
+  pub(crate) cygpath: PathBuf,
   pub(crate) dotenv_filename: Option<String>,
   pub(crate) dotenv_path: Option<PathBuf>,
   pub(crate) dry_run: bool,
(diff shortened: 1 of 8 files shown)


casey (changes requested):
I think this should be done as a command-line option, `--cygpath`, which also can be set using an environment variable. See `config.rs`. Command line options show up in `--help`, so they're a little more discoverable.

hyrious:
Hi I just added the `--cygpath` command-line option. I'm using `Option<OsString>` as the config value, but there's no other config using the same type. It seems that `clap` does accept this type and my local testing also passes. I'm not sure if it is the prefered way to do this.

casey:
Nice. I think it should be a `PathBuf`, since it could either be just the filename, or an absolute path to the binary.

hyrious:
Done. I have also tested it locally to verify it works.

casey:
Nice, thank you!

Issue #476, opened 2019-09-24
Add flags and options


casey:
Recipes should be able to define options:

```
foo --bar BAZ:
  echo {{BAZ}}
```

Maybe flags too:

```
foo --BAZ:
  echo {{BAZ}} # is it "true" or "false"? Or "" or BAZ?
```

Large changes to the parsing of recipe parameters like this should be considered holistically, along with other possible changes, like `$` to support exporting parameters as environment variables, or changing all just variables to being environment variables.

casey:
It occurs to me that we could use clap to construct recipe argument parsers.

hustcer:
@casey Will flags be supported ? It's a quite common use case , Thanks !

hustcer:
flag shorthand can also be supported:
`foo --bar(-b):`

sascha-wolf:
Just my two cents: I would already be super happy if I could explicitly refer to the argument's name:

```Justfile
foo bar=something:
	echo {{bar}}
```

```
$ just foo bar=stuff
stuff
```

Of course fully featured flags and options would be lovely but for my personal use-case (and I imagine other's as well) this would already be very helpful, because then I can do things like this:

```Justfile
watch folders='lib,test' cmd='just test':
  find {{ '{' }}{{folders}}{{ '}' }} -name '*.ex*' | entr -dc {{cmd}}
```

```
$ just watch cmd='<some other command>'
```

colemannerd:
I've just started using Just and my org loves it! 

One thing that would improve our usage is opts with enums. We use just to coordinate all of our environment scripts and so we have done things like standardize accounts or environments to a certain list.  Right now, they're just strings that we then call a pre-recipe on to check that they are one of a list.  It would be awesome if we had this functionality AND incorporaitng an enum for one of the opts.  I know https://github.com/casey/just/issues/1990 is also similar.

chosroes:
Any hope for this feature?

casey:
I'm not currently working on it. It's pretty complex, but it is a good feature.

alshdavid:
Would definitely love this capability. Right now you can use env vars as named parameters
```
env profile=release os=linux arch=amd64 just build
```
But it would be great if the arguments could be first class citizens of just (with default values if omitted) because they are discoverable with `just --list` and can be added to bash/zsh/fish/etc autocomplete.

For example:
```
just build --profile release --os linux --arch amd64

```
[...]

PR #2806, opened 2025-07-07
Add support for flags and options (issue #476)

chmp:
Implementation for #476. See below for the design decisions I took in the implementation. 

# Example

```just
# excerpt from `examples/kitchen-sink.just`

flags --flag --option="foo" positional:
  @echo "flag:      " {{ flag }}
  @echo "option:    " {{ option }}
  @echo "positional:" {{ positional}}
```

Invocations

```bash
$ cargo run -q -- --justfile ./examples/kitchen-sink.just flags bar
flag:
option:     foo
positional: bar

$ cargo run -q -- --justfile ./examples/kitchen-sink.just flags --option 42 bar
flag:
option:     42
positional: bar

cargo run -q -- --justfile ./examples/kitchen-sink.just flags --flag bar
flag:       flag
option:     foo
positional: bar

$ cargo run -q -- --justfile ./examples/kitchen-sink.just flags --option 42 --flag bar
flag:       flag
option:     42
positional: bar

# out of order
$ cargo run -q -- --justfile ./examples/kitchen-sink.just flags bar --option test --flag
flag:       flag
option:     test
positional: bar

# special cases
$ cargo run -q -- --justfile ./examples/kitchen-sink.just flags -- --flag
flag:
option:     foo
positional: --flag

$ cargo run -q -- --justfile ./examples/kitchen-sink.just flags --option --flag  bar
flag:
option:     --flag
positional: bar
```

## Design

- Flags are defined by `---parameter` without default value
- Options are defined by `--parameter=default` 
- Flags and options must be defined before positional parameters. Positional arguments are handled as before (including `*` and `+` modifiers)
- Flags evaluate to their name when passed. When not passed, they evaluate to the empty string . This way `flag != ""` checks for the flag
- Options consume the next parameter and evaluate to it when passed. When not passed, they evaluate to their default value
- Options and flags can be passed in any order and mixed with positional arguments
- `--` forces all following arguments to interpreted as positional for the given recipe. When specifying multiple recipes, the parser resets for each new recipe
- Recipes without flag parameters accept flags or options as positional arguments. This behavior ensure backwards compatibility (see for example the `trailing_flags` test)
- `--$flag` or `--$option=default` defining environment variables

## Open questions:

- Support `--options-with-hyphens`? At the moment underscores must be used as in `--options_with_underscores`
- How to update the Chinese readme?


tests/misc.rs
```diff
@@ -2065,6 +2065,8 @@ a:
 fn trailing_flags() {
   Test::new()
     .arg("echo")
+    // TODO(chmp): remove when unknown flags should just be grouped
```
chmp: Decide whether to keep this breaking change or adopt another strategy

src/argument_parser.rs
```diff
@@ -75,25 +76,80 @@ impl<'src: 'run, 'run> ArgumentParser<'src, 'run> {
...
+
+      if verbatim {
+        positional_arguments.push(current);
+        consumed += 1;
+      } else if current == "--" {
+        verbatim = true;
+        consumed += 1;
+      } else if let Some(flag) = current.strip_prefix("--") {
+        let Some(parameter) = recipe.flags.get(flag) else {
+          return Err(Error::internal("TODO: fix error message, unknown flag"));
```
chmp: Fix error message

src/argument_parser.rs
```diff
@@ -75,25 +76,80 @@ impl<'src: 'run, 'run> ArgumentParser<'src, 'run> {
...
+        };
+        if parameter.default.is_some() {
+          if consumed + 1 >= args.len() {
+            return Err(Error::internal(
+              "TODO: fix error message, missing option argument",
```
chmp: Fix error message

src/argument_parser.rs
```diff
@@ -75,25 +76,80 @@ impl<'src: 'run, 'run> ArgumentParser<'src, 'run> {
...
-    let argument_count = cmp::min(rest.len(), recipe.max_arguments());
-    if !argument_range.range_contains(&argument_count) {
+    Ok(ArgumentGroup {
+      positional_arguments,
+      flag_arguments,
+      path,
+    })
+  }
+
+  fn parse_recipe_arguments(
```
chmp, line 90: The parser definitely requires more tests

chmp (commented):
Highlight todo items

tests/misc.rs
chmp: Removed the breaking change and document the design.

chmp:
Merge `casey:master` to keep PR up to date

chosroes:
Need this a lot

chmp:
Small observation from my own use of this feature: the definition of flags plays quite nicely with the unstable `&&` operator. 

For example, this justfile

```just
set unstable 

test --full:
    cargo test {{ full && "--all-targets --all-features" }}
```

allows to use 

- `just test` to run `cargo test`, and 
- `just test --full` to run `cargo test "--all-targets --all-features"`

Related example:

```just
set unstable 

run --port="" path="example":
    RUST_LOG="debug" cargo run -p server -- {{ port && ("--port " + port) }} {{ path }}
```

PaulRBerg:
hey @casey I just tipped you $20 via GitHub sponsors. Could you please take a look quick at this PR?

I suggest other folks in this thread to tip, too.

Flags are an essential features for CLIs but they are not currently available in Just :(

chmp:
Merged master to keep the PR up to date

casey:
So sorry to close this, but I wound up implementing this myself in #3026.

I was kind of burned out on working on `just` for a while, which is why I never gave this PR the attention it deserved.

I wound up doing a few things differently, which you might find interesting:

I added an `[arg("arg", pattern="REGEX")]` attribute for enforcing that an argument value must match a regular expression. Since new `key=value` keys can be added, I thought that this would be an ideal way to add flags and options. using `recipe --flag:` syntax is cute, but adds to the complexity of the syntax of parameters, which is already complicated, and is hard to extend in the future. What if you want to add additional long aliases? Or short flags? Using `[arg(…)]` means that these are just new keys, which makes extensions straightforward.

So, the way to specify that an argument should be passed as a long option is:

```just
[arg("bar", long="bar")]
foo bar:
```

This can be invoked as:

```console
$ just foo --bar value
```

I didn't add short options `-x`, or flags, i.e. options which do not take a value, initially for simplicity. For those I'm thinking:

```just
# short option
[arg("bar", short="b")]
foo bar:

# flag
[arg("bar", long="bar", value="value")]
foo bar="default":
```

The somewhat weird `value=` means that it doesn't take a value, and when passed, it gets value `"value"`, and when it doesn't, it gets `"default"`. Kind of as a cop out, I didn't want to mandate what value a flag would take when it's passed. This is because there's a longstanding tension between what values are truthy and falsey, which are different in different places (in some places it's `true` and `false`, in some places the empty string is false and everything else is true, and I just forcing the user to specify values in both cases is an easy punt.

The implementation that I eventually landed benefited from a refactor I did in #3025, which changes the way that arguments are passed around from `Vec<String>`, to `Vec<Vec<String>>`, i.e, arguments are now pre-grouped into which parameter they're for. This allows avoiding passing around a separate `flag_arguments` field. The `InvocationParser` (new name for the `ArgumentParser`) groups arguments, whether positional or options, by parameter, so other parts of the program don't need to know about options.

Anyways, sorry again for closing this, but I was slacking off pretty hard for a while, and eventually came up with a different path to implementation.

chmp:
@casey Thanks a lot for the explanation. And no worries, I fully understood that such a PR without prior discussion has a good chance of being not merged. It was still fun to implement :) 

Looking forward to play around with your design - and to port my local Justfiles ;).

casey:
You bet, and thanks for understanding.

Issue #2824, opened 2025-07-16
Bug: Module dependencies in Recipes don't change cwd


corvusrabus:
### Problem

Example:
Assume a folder structure of
├── justfile
└── a
    └── justfile

with `a/justfile` containing

```justfile
hello:
  @pwd 
```
and `a/justfile` containing

```justfile
mod a

hello: a::hello
```

Running just hello should print `pwd/a` but prints `pwd`

### Solution
I can look at this in 2-3 weeks if no one else picks it up before.

### References
#2672 #2252

eisbaw:
Hi @corvusrabus I believe I fixed this in https://github.com/casey/just/pull/2829 . That PR also has a doc fix.

casey:
Fixed in #2829.

PR #2829, opened 2025-07-18
Submodules: Fix readme and issue 2824

eisbaw:
See commit messages for details.

This fixes issue https://github.com/casey/just/issues/2824

### DIFF
--- a/README.md
+++ b/README.md
@@ -3797,10 +3797,9 @@ Available recipes:
     foo ... # foo is a great module!
 ```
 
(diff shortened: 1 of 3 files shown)


casey:
Thanks for this PR!

I'm not sure about the readme changes, so I reverted them.

I also didn't like the approach of passing in an execution context at the top level. Instead, I re-used the scope dictionary to also store a reference to the justfile, and retrieve it when we retrieve the scope, and build the correct execution context at that point.

PR #2832, opened 2025-07-22
Add --list-all command that comprehensively lists everything

pranjalv123:
We migrated a Makefile to a Justfile with a number of submodules, and one thing that's missing is the ability to easily grep the makefile for the command you're looking for. 

Adding the `--list-all` command allows the user to search for what they're looking for in a nested Just configuration; it also allows you to pass the output to an LLM that can answer questions about available tools.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -28,6 +28,7 @@ pub(crate) struct Config {
   pub(crate) list_heading: String,
   pub(crate) list_prefix: String,
   pub(crate) list_submodules: bool,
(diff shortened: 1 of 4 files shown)


laniakea64:
Isn't this same as `just --alias-style=separate --list-submodules --list`?

pranjalv123:
yup, didn't realize that `--list-submodules` could be invoked without arguments if you also pass `--list`!

Issue #2828, opened 2025-07-17
`--shell` argument not used for evaluating shebangs on Windows


vsalvino:
I have a clean Windows install. Git bash is installed, however it is NOT on the PATH. This is critical, because putting `bash.exe` on the path will conflict with other `bash.exe` on the PATH, notably the one shipped by Windows in System32 which is used for WSL. Overriding this will cause WSL to break, which is NOT a solution. In general, adding too many tool directories to the path inevitably leads to breakage at some point, as many tools ship slightly different versions of exe's and dll's.

Justfile:
```justfile
# Works within Git Bash.
# DOES NOT work when specifying `--shell C:\Program Files\Git\bin\bash.exe`
bash:
  #!/bin/bash
  echo "Hello world"
```

Works within Git Bash
```shell
# Drop into git bash
PS> & 'C:\Program Files\Git\bin\bash.exe'
$ just bash
Hello world
```

Does not work when specifying git bash as the shell:
```shell
PS> just --shell 'C:\Program Files\Git\bin\bash.exe' bash
error: Could not find `cygpath` executable to translate recipe `bash` shebang interpreter path:
program not found
```

It seems to me that the PATH *within* the `--shell` should be used to invoke shebangs. It looks like it is currently looking on the PATH outside of the shell. This is important because the specified `--shell` sets the path to include cygwin/mingw binaries.

This would be a huge win for Windows users, as we could set the `windows-shell` from within the Justfile to have it "just work" for people who naively run it on Windows from CMD or PowerShell.

hrumhurum:
A slightly different approach, but maybe you can use something like this:

```just
set windows-shell := ["gnu-tk", "-i", "-c"]

run:
    echo Just hello
    cp --help
```

GNU-TK is a mutli-platform command-line utility that provides seamless access to GNU toolkits on non-GNU operating systems: https://github.com/gapotchenko/gnu-tk

vsalvino:
The problem is not the shell (Git Bash works perfectly fine). The problem is that `just` does not seem to correctly use the shell that is specified when passed in via the `--shell` argument.

casey:
Am I understanding correctly that the suggestion is that just execute the argument to `--shell`, extract the value of the `PATH` variable from within that shell, and then use that instead of the existing `PATH` variable?

[...]

PR #2840, opened 2025-07-25
Add Windows-specific setup instructions to README

vsalvino:
Adding these instructions based on my own difficulties getting `just` to work on Windows (See #2828)

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,13 +91,18 @@ Installation
 `just` should run on any system with a reasonable `sh`, including Linux, MacOS,
 and the BSDs.
 
+### Notes for Windows
 On Windows, `just` works with the `sh` provided by
 [Git for Windows](https://git-scm.com),
 [GitHub Desktop](https://desktop.github.com), or
 [Cygwin](http://www.cygwin.com).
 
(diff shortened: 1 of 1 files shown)


casey:
I'm wondering if this adds much.

When you install Git for Windows, GitHub Deskop, or Cygwin, do none of them install `sh` in `PATH`? I.e., after installing them you must manually add it to `PATH` yourself if you want to use `just` within PowerShell or `cmd.exe`?

vsalvino:
Git for Windows only sets the specific `git` programs on the PATH, not any of the utilities. My (admittedly very dated) past experience with Cygwin is that it is similar - and encourages you to launch a cygwin shell manually which contains all of the unix utilities on the PATH within that shell only. Can't speak for Github desktop.

Meddling with the PATH in Windows can lead to some weird breakage. For example, Windows now ships `bash.exe` in System32 on the path, which starts a WSL virtual machine - it is not actually a real bash shell.

casey:
Okay, gotcha, thanks for the explanation. I opted for a smaller addition in #2842.

PR #2841, opened 2025-07-26
Fix exported parameters with defaults consuming variadic arguments

### LINKED ISSUE #2825: Recipe-level environment variables
Right now, if I want to set an environment variable just for a single recipe, but for all commands executed on that recipe, this is the only way I'm aware of to do this:

```
test $THE_ENV='hi':
  echo $THE_EN
```

This works great unless you need to accept additional parameters. Then it breaks.

Here's an example:

```
test $THE_ENV='hi' *args:
  echo $THE_ENV
  echo {{args}}
```

And some output:

```
❯ just test this arg gets assigned
echo $THE_ENV
this
echo arg gets assigned
arg gets assigned
```

It would be great if it was possible to assign a static recipe-level environment variable.

chojs23:
## Problem

When using Parameters prefixed with a $ will be exported as environment variables alongside variadic parameters, the exported parameter would incorrectly consume command line arguments instead of using its value.

  **Before:**
  ```just
  test $THE_ENV='env value' *args:
    echo $THE_ENV
    echo {{args}}

  $ just test this arg gets assigned
  # Output:
  this              # THE_ENV got "this" instead of "env value"
  arg gets assigned  # args got remaining arguments
```

  After:
```
  $ just test this arg gets assigned
  # Output:
  env value                       # THE_ENV uses default value "env value"
  this arg gets assigned  # args gets all command line arguments
```
  Root Cause

  The parameter evaluation logic in evaluate_parameters() processed parameters sequentially without considering that exported parameters with defaults should prefer their default values when variadic parameters are present.
   This caused exported parameters to steal arguments intended for variadic parameters.

Fixes #2825

### DIFF
--- a/src/evaluator.rs
+++ b/src/evaluator.rs
@@ -330,6 +330,9 @@ impl<'src, 'run> Evaluator<'src, 'run> {
 
     let mut positional = Vec::new();
 
+    // Check if there's a variadic parameter to determine argument consumption behavior
+    let has_variadic = parameters.iter().any(|p| p.kind.is_variadic());
+
     let mut rest = arguments;
     for parameter in parameters {
       let value = if rest.is_empty() {
@@ -350,6 +353,11 @@ impl<'src, 'run> Evaluator<'src, 'run> {
         }
         let value = rest.to_vec().join(" ");
         rest = &[];
+        value
+      } else if parameter.export && parameter.default.is_some() && has_variadic {
+        // Exported parameters with defaults should use their default value when variadic parameters are present
+        let value = evaluator.evaluate_expression(parameter.default.as_ref().unwrap())?;
+        positional.push(value.clone());
         value
       } else {
         let value = rest[0].clone();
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by laniakea64** [comment]
This would break `just`'s [backwards-compatibility guarantee](https://just.systems/man/en/backwards-compatibility.html).

The goal you're seeking (statically-defined per-recipe/recipe-specific environment variables) sounds useful, but it would need to be implemented with a new [recipe attribute](https://just.systems/man/en/attributes.html) or some other new syntax.

**Comment by casey** [comment]
This change is a feature, not a bug. Just's parameters are all positional, including exported parameters with default values. #2825 requires a design for per-recipe exports, separate from parameters.

**Comment by chojs23** [comment]
Ah, what about adding new receipe attribute? Does this looks fine?

Like

```
[env("VAR1=foo")]
recipeA *args:
  echo $VAR1

[env("VAR1=bar")]
recipe2 *args:
  echo $VAR1

```

PR #2844, opened 2025-07-27
Add env attribute for recipe-level environment variables

chojs23:
This PR adds `env` recipe attribute that can specify static recipe level environment variable.
Add new `[env(VAR=VALUE, ...)]` attribute to set environment variables at the recipe level.

Related #2825 #2841 

I would lie feedback on whether I need to add anything to the documents and whether any tests.


casey:
Thanks for the PR!

Syntax for setting recipe-level environmental variables needs to be in the form of expressions which can use values of parameters. Also, parsing as `=`-separated strings is very inflexible.

In any case, I would wait until there's a good design before opening a PR.

PR #2845, opened 2025-07-27
docs: suggest more sophisticated default recipe

salim-b:
makes the default command
- hidden in `just --list`
- not echo its cmdline
- work when `just` is not on `PATH`
- work with non-standard justfile name / when invoked from another working directory

### DIFF
--- a/README.md
+++ b/README.md
@@ -752,7 +752,7 @@ beginning of your `justfile` that lists the available recipes:
 
 ```just
 _default:
-  @{{just_executable()}} --list --justfile {{justfile()}}
+  @{{quote(just_executable())}} --list --justfile={{quote(justfile())}}
 ```
 
 ### Listing Available Recipes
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -751,8 +751,8 @@ If no recipe makes sense as the default recipe, you can add a recipe to the
 beginning of your `justfile` that lists the available recipes:
 
 ```just
-default:
-  just --list
+_default:
+  @{{just_executable()}} --list --justfile {{justfile()}}
```
laniakea64: Since the paths returned by `just_executable()` and `justfile()` might contain whitespace, should these interpolations use [`quote()` function](https://just.systems/man/en/functions.html#string-manipulation), or would that not work on Windows?

salim-b: I've added the `quote()`s. Still dunno about Windows, though.

casey:
Thanks for the PR!

I actually think the correct approach here is to add a setting which makes `--list` the default action, i.e.:

```just
set default := list
```

This would avoid needing to recursively invoke the `just` binary or forward arguments.

casey:
Opened #3191 to track this.

PR #2851, opened 2025-08-04
Add container image support with multi-arch builds

### LINKED ISSUE #1497: Container image
A multi-stage build could be used to install this, no?

```
FROM just as just

FROM python

COPY --from=just /bin/just /bin/just

# Copy everything in the current directory - use .containerignore / .dockerignore
COPY . .

just
```

~~Has the added benefit of being able to try out just.~~

~~It probably shouldn't be recommended as the base, but perhaps having some common distros like alpine, debian, and ubuntu would be nice for those trying it out or not wanting to pull down layers.~~

~~A con is the overhead of having other stuff in the image but it seems like it would be better than downloading it in a podman/docker command.~~ Or maybe some are distributing container images that can only be used for copying like above.

**My current suggestion is [just to have a bare distroless container](https://github.com/casey/just/issues/1497#issuecomment-1382310193) containing `just` so you can use the above commands to add the `just` static binary to any container.**

elasticdotventures:
## Summary
Adds Docker container support that automatically builds and publishes multi-platform container images to GitHub Container Registry (GHCR) for each release.

## Changes
- **Dockerfile**: Multi-stage build downloading appropriate `just` binary from GitHub releases
- **GitHub Action**: Automatically triggers on releases to build and publish container images  
- **README**: Added container usage documentation with multistage build example

## Key Features
- Multi-architecture support: `linux/amd64`, `linux/arm64`, `linux/arm/v7`
- Minimal ~10MB image using `scratch` base
- Semantic versioning: `1.42.4`, `1.42`, `1`, `latest` tags
- Automatic release integration
- Architecture-aware binary selection

## Test Plan
- [ ] Test manual workflow dispatch
- [ ] Verify multi-arch builds work correctly
- [ ] Confirm container runs and executes `just --help`
- [ ] Test multistage build copying

Fixes #1497


.github/workflows/container.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - name: Extract metadata
+      id: meta
+      uses: docker/metadata-action@v5
+      with:
+        images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+        tags: |
+          type=semver,pattern={{version}},value=${{ steps.version.outputs.version }}
+          type=semver,pattern={{major}}.{{minor}},value=${{ steps.version.outputs.version }}
+          type=semver,pattern={{major}},value=${{ steps.version.outputs.version }}
+          type=raw,value=latest
```
polarathene: Potential caveats that projects should be aware of with this:
- The triggered release event / manual dispatch will always update the `latest` tag, even if unnecessary / inaccurate (_backport fix_).
- Likewise updating an earlier minor version with a patch release will update the major tag even if this was for an earlier minor series.

It's probably a non-issue in this project, but I have seen it happen on larger projects with more complex release management updating tags incorrectly.

.github/workflows/container.yaml
```diff
@@ -0,0 +1,71 @@
...
+      uses: actions/checkout@v4
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Log in to Container Registry
+      uses: docker/login-action@v3
+      with:
+        registry: ${{ env.REGISTRY }}
+        username: ${{ github.actor }}
```
polarathene: Repo owner is a more accurate and consistent value:

```suggestion
        username: ${{ github.repository_owner }}
```

.github/workflows/container.yaml
```diff
@@ -0,0 +1,71 @@
...
+          type=semver,pattern={{version}},value=${{ steps.version.outputs.version }}
+          type=semver,pattern={{major}}.{{minor}},value=${{ steps.version.outputs.version }}
+          type=semver,pattern={{major}},value=${{ steps.version.outputs.version }}
+          type=raw,value=latest
+
+    - name: Build and push Docker image
+      uses: docker/build-push-action@v6
+      with:
+        context: .
+        platforms: linux/amd64,linux/arm64,linux/arm/v7
```
polarathene: There should not be a need to support ARMv7, [those devices are approx over a decade old](https://github.com/docker-mailserver/docker-mailserver/issues/2642) and often cheap to replace with newer aarch64 products. In recent years there has been a push by distros to drop support (_IIRC one was citing EOL support around 2026_).

While it's easy to add platforms to build when supported by the toolchain / environment, it is rather wasteful unless there's actual user demand to warrant the resources used.

```suggestion
        platforms: linux/amd64,linux/arm64
```

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN set -eux; \
+    case "$TARGETARCH" in \
+        "amd64") ARCH="x86_64-unknown-linux-musl" ;; \
+        "arm64") ARCH="aarch64-unknown-linux-musl" ;; \
+        "arm") \
+            # Handle ARM variants - v7 maps to armv7, others to generic arm
+            case "$TARGETVARIANT" in \
+                "v7") ARCH="armv7-unknown-linux-musleabihf" ;; \
+                *) ARCH="arm-unknown-linux-musleabihf" ;; \
+            esac ;; \
```
polarathene: As mentioned please don't burden maintainers with legacy platforms for a minority of users that could either upgrade for a small price or build `just` themselves.

Here you seem to be building with the platform ARMv7, but building for other platforms that aren't v7?

From personal investigation, that is [problematic depending on base image choice](https://github.com/testssl/testssl.sh/issues/2422#issuecomment-2841813704) (_which isn't uncommon to be changed at a later date_), especially if a project brings in deps requiring sys crates.

```suggestion
```

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+        *) echo "Unsupported architecture: $TARGETARCH" && exit 1 ;; \
+    esac; \
+    wget -O just.tar.gz "https://github.com/casey/just/releases/download/${JUST_VERSION}/just-${JUST_VERSION}-${ARCH}.tar.gz"; \
+    tar -xzf just.tar.gz; \
+    chmod +x just
```
polarathene: As explained in [my comment](https://github.com/casey/just/issues/1497#issuecomment-1442903322) of the related issue, you can just pipe the output through into `tar` and extract the contents, it is already executable (_archive preserved it from build output_):

```suggestion
    wget -O - https://github.com/casey/just/releases/download/${JUST_VERSION}/just-${JUST_VERSION}-${ARCH}.tar.gz \
        | tar --directory /usr/local/bin --extract --gzip --file - --no-same-owner just
```

Dockerfile
polarathene: **For reference:** You could also avoid this whole script to acquire `ARCH` with (_granted the below approach isn't as nice, but it is a handy workaround for more complex `Dockerfile` builds when you need to persist `ARG` across instructions_):

```Dockerfile
# Docker automatically provides this for multi-platform builds
ARG TARGETARCH

FROM alpine AS base
SHELL ["/bin/ash", "-eux", "-o", "pipefail", "-c"]

# Map Docker arch convention to Rust musl target
FROM base AS downloader-arm64
ARG ARCH=aarch64-unknown-linux-musl
FROM base AS downloader-amd64
ARG ARCH=x86_64-unknown-linux-musl

FROM downloader-${TARGETARCH} AS downloader
# Version will be passed from GitHub Actions during release builds
ARG JUST_VERSION
ARG RELEASE_URL=https://github.com/casey/just/releases/download/${JUST_VERSION}/just-${JUST_VERSION}-${ARCH}.tar.gz
RUN test -n "${JUST_VERSION}" && \
    wget -O - "${RELEASE_URL}" | tar -xz -f - -C /usr/local/bin just
```

polarathene: This type of logic though is quite a common task, that for projects like `just` you are better off getting an image published of a tool like [`eget`](https://github.com/zyedidia/eget) instead?:

```Dockerfile
# eget bin is about 6.5MB

FROM alpine AS bin-eget
ARG TARGETARCH
ARG RELEASE_VERSION=1.3.4
ARG RELEASE_URL=https://github.com/zyedidia/eget/releases/download/v${RELEASE_VERSION}/eget-${RELEASE_VERSION}-linux_${TARGETARCH}.tar.gz
RUN <<HEREDOC
  apk --no-cache add curl

  curl -fsSL "${RELEASE_URL}" | tar \
    --extract --gzip \
    --directory /usr/local/bin \
    --strip-components=1 \
    --no-same-owner \
    eget-${RELEASE_VERSION}-linux_${TARGETARCH}/eget
HEREDOC
```

Once you have an image published like that (_there isn't an official one for some reason_ 🤷‍♂️), then for projects like `just` you don't need a `Dockerfile` like this PR proposes.

You'd instead have this:

```Dockerfile
FROM scratch
RUN --mount=type=bind,from=eget:latest,source=/usr/local/bin/eget,target=/usr/local/bin/eget \
  eget casey/just --tag 1.42.4 --to /usr/local/bin
```

or

```Dockerfile
FROM eget as downloader
RUN eget casey/just --tag 1.42.4 --to /just

FROM scratch
COPY --from=downloader /just /usr/local/bin/just
```

The `--tag` and `--to` options are completely optional. By default they'll select the latest release (_not always ideal, similar to the caveats I mentioned regarding how your PR handles tagging the image_), and it'll download/extract into the current working directory.

UX will vary by repo as some projects manage releases a bit more broadly with assets attached or bad naming conventions that a tool like `eget` cannot infer which file to download without some additional filtering for guidance (_supported via `--asset`_).

As you can see though it becomes vastly simpler for other images and reduces boilerplate.

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+FROM alpine:latest as downloader
+
+# Version will be passed from GitHub Actions during release builds
+ARG JUST_VERSION
+# Docker automatically provides these for multi-platform builds
+ARG TARGETARCH
+ARG TARGETVARIANT
+
+# Install tools needed to download and extract the binary
+RUN apk add --no-cache wget tar
```
polarathene: These deps aren't needed with Alpine, they're already available by default.

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+    tar -xzf just.tar.gz; \
+    chmod +x just
+
+# Use scratch for minimal final image - no OS, just the binary
+# This results in a ~10MB image vs ~50MB+ with a full OS
+FROM scratch
+COPY --from=downloader /just /usr/local/bin/just
+
+# Set working directory for when just is run
+WORKDIR /work
```
polarathene: Perhaps something a little more appropriate if you're choosing a custom location?

```suggestion
WORKDIR /just
```

Is there a scenario where you would use this implicitly? (_no mounts to this location?_)

I suppose it provides a small convenience at runtime for those that are aware of it, otherwise it can be a tad annoying to inherit a WORKDIR from a base image when extending (_CoreDNS had this issue when they switched from Debian to Google Distroless as their base image which broke users with a confusing failure about their configs not being read as the change was not documented in any manner_).

If you do think this is worth having as a default, you should add a mention about this in the README docs you've added (_which only seem to touch on using the image as a distribution mechanism where the `WORKDIR` here is irrelevant in that `COPY` context_).

README.md
```diff
@@ -4678,4 +4678,14 @@ your computational endeavors!
...
 
+## Container Usage
+
+Copy `just` into your multistage build (automatically selects correct architecture):
+
+```dockerfile
+FROM ghcr.io/casey/just:latest as just
+FROM your-base-image
+COPY --from=just /usr/local/bin/just /usr/local/bin/just
+```
```
polarathene: Doesn't need to be multi-stage, and implicitly matching the same architecture should be assumed.

Prior to better registry support, projects would publish alternative arches with a platform tag as a workaround, and you still see some projects doing that, but it shouldn't be confused with best practice.

````suggestion
Copy `just` into your own `Dockerfile`:

```Dockerfile
COPY --link --from=ghcr.io/casey/just:latest /usr/local/bin/just /usr/local/bin/just
```
````

File: .github/workflows/container.yaml [hunk 9]
```diff
@@ -0,0 +1,71 @@
...
+        push: true
+        tags: ${{ steps.meta.outputs.tags }}
+        labels: ${{ steps.meta.outputs.labels }}
+        build-args: |
+          JUST_VERSION=${{ steps.version.outputs.version }}
+        cache-from: type=gha
+        cache-to: type=gha,mode=max
```
polarathene: Why? There's really nothing worth caching here? It'd likely be faster to skip all that.

.github/workflows/container.yaml
```diff
@@ -0,0 +1,71 @@
+name: Container Image
+
+on:
+  release:
+    types: [published]
```
polarathene: Some caution may be needed when relying on this event. Are you confident it's appropriate for this project?

Releases appear to be automated in CI:

https://github.com/casey/just/blob/c594c8e73d52b56f930d768feaf4c2948f61dd20/.github/workflows/release.yaml#L115-L123

That may be an issue, for example see: https://github.com/orgs/community/discussions/25281#discussioncomment-3300251

---

Typically it's better to chain a dependent workflow to the one producing the artifacts, but there can be some security considerations that also need to be taken into account there depending on how you approach it (_speaking from experience as I've introduced a vulnerability from doing it wrong before_ 😓).

I'm more iffy about this since you're relying on the event to retrieve the assets, otherwise manual dispatch trigger which is probably going to be necessary as a workaround.

polarathene (changes requested):
Some feedback worth considering (_sorry about the verbosity, short on time_).

.github/workflows/container.yaml
elasticdotventures: i could call it lastest .. whatever was published last; i probably won't use this and don't really have a strong feeling other than latest should point at something.

.github/workflows/container.yaml
elasticdotventures: 🤣 Lol, I originally wrote it that way; but the thread about this topic seemed so old and contenious I figured it'd be less controversial by keeping it separate & downstream in it's own file.   Didn't want to upset anybody who felt that the primary build should be separate, if there's no objections I'm happy to chain it in the main artifact build.

Dockerfile
elasticdotventures: yeah i saw your earlier note - clever.; my containers don't always have internet access or limited bandwidth (ex: while travelling on planes).  docker provides a nice prepopulated caching layer for my dev-x.

elasticdotventures:
Hi @polarathene,

Thank you for the thorough feedback on the container implementation\! I've addressed all the points you raised:

**Changes made:**
- ✅ Updated authentication to use `github.repository_owner` instead of `github.actor` for better consistency
- ✅ Removed ARMv7 platform support (`linux/arm/v7`) to focus on modern architectures (amd64/arm64 only)
- ✅ Simplified the Dockerfile download/extraction logic using the direct tar pipeline approach you suggested
- ✅ Improved tagging logic to only update `latest`, major, and minor version tags when building from the default branch

The container now follows the best practices you outlined while keeping the implementation clean and maintainable. The ARMv7 removal should help avoid the compatibility issues you mentioned, and the improved tagging prevents inappropriate updates to latest/major tags from older releases.

Thanks again for taking the time to provide such detailed and helpful feedback\!

Dockerfile
elasticdotventures: removed

.github/workflows/container.yaml
elasticdotventures: nice catch!

.github/workflows/container.yaml
polarathene: If the published release event is reliable that'd be great, I'm just not sure how reliably that will be.

Few alternatives below if it helps for resolving the concern, but they're all extra work/hassle.

---

The approach I've used before is to have re-usable workflows, so they're still separate workflow files like you have here, but they don't have the typical event trigger.

- [Calling side](https://github.com/docker-mailserver/docker-mailserver/blob/69e2b56defe5135aa25125c083d61f960d0b0016/.github/workflows/default_on_push.yml#L33-L39) triggered manually via push to default branch or tag pushed (_this project tags releases by publishing via GH releases UI_)
- [Called side](https://github.com/docker-mailserver/docker-mailserver/blob/69e2b56defe5135aa25125c083d61f960d0b0016/.github/workflows/generic_publish.yml#L62-L81) via `workflow_call` trigger.

It's an old workflow I contributed some time ago, quite a bit of the feedback I gave here reviewing your PR needs to be addressed at that project too 😅 but you can see that the trigger calling side is rather simple to grok at a higher level that just acts as glue. Something similar could be used for the existing release workflow to call into your new image publish workflow here. 

That would need something like [upload](https://github.com/docker-mailserver/docker-mailserver/blob/69e2b56defe5135aa25125c083d61f960d0b0016/.github/workflows/docs-preview-prepare.yml#L68-L74) + [download](https://github.com/docker-mailserver/docker-mailserver/blob/69e2b56defe5135aa25125c083d61f960d0b0016/.github/workflows/docs-preview-deploy.yml#L66-L74) steps on each end for the artifacts though, which isn't great.

---

The other option is turning the workflow into an internal composite action, so that it's a single step on the `release.yml` workflow :shrug:

Or building from source again with a separate builder stage.

README.md
```diff
@@ -4678,4 +4678,14 @@ your computational endeavors!
 
 😸
 
+## Container Usage
+
+Copy `just` into your multistage build (automatically selects correct architecture):
+
+```dockerfile
+FROM ghcr.io/casey/just:latest as just
```
ulgens: This can be further simplified by the removal of the aliasing. It's not a requirement for a working implementation.

.github/workflows/container.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - name: Extract metadata
+      id: meta
+      uses: docker/metadata-action@v5
+      with:
+        images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+        tags: |
+          type=semver,pattern={{version}},value=${{ steps.version.outputs.version }}
+          type=semver,pattern={{major}}.{{minor}},value=${{ steps.version.outputs.version }},enable={{is_default_branch}}
+          type=semver,pattern={{major}},value=${{ steps.version.outputs.version }},enable={{is_default_branch}}
+          type=raw,value=latest,enable={{is_default_branch}}
```
polarathene: Only triggering the last 3 tags based on if it's a release event instead of a manual trigger is fine I think 🤔

Although I'm not quite sure how reliably that is in this scenario as tags aren't associated to branches AFAIK, I assume the release event commit is associated to a tag rather than a specific branch with the commit too? (_maybe it's documented publicly somewhere, otherwise you won't know until you try 😭)

I've shared a snippet from one of my own projects workflows, where I do want to highlight a bug with `latest` tag. Since the `raw` type doesn't understand `latest`, it's just a string, there's also the default `flavor` setting `latest=auto` that under some conditions can infer adding a `latest` tag and updating that implicitly, where **`enable=false` for the `raw` tag type would be ignored** in that case. [More info](https://github.com/docker/metadata-action/issues/461#issuecomment-2686558629) + fix below (`flavor.latest=false`).

---

If you ever need to determine if a version is the latest, I have written a step for figuring that out (was a bit tricky):

```yaml
    - name: 'Find the latest release tag (determined by semver sort)'
      id: latest-semver
      if: startsWith(github.ref, 'refs/tags/v')
      env:
        REPO_URL: https://gha:${{ inputs.token }}@github.com/${{ github.event.repository.full_name }}
        # Filter for semver tags like: `v1.2.34` (Ignores pre-release tags)
        TAG_FILTER: '^refs/tags/v[0-9]+(\.[0-9]+)?{2}$'
        TAG_REF: ${{ github.ref_name }}
      shell: bash
      run: |
        TAG=$(
          git ls-remote --exit-code --refs --tags --sort='-v:refname' "${REPO_URL}" \
          | awk '{print $2}' | grep -E "${TAG_FILTER}" | head -n 1
        )
        echo "tag=${TAG}" >> "${GITHUB_OUTPUT}"
```

Then using it like:

```yaml
    - name: 'Prepare image tags'
      id: docker-metadata
      uses: docker/metadata-action@v5
      env:
        is-latest: ${{ github.ref == steps.latest-semver.outputs.tag }}
        is-not-major-zero-ver: ${{ !startsWith(github.ref, 'refs/tags/v0.') }}
      with:
        images: |
          ghcr.io/${{ github.repository_owner }}/${{ inputs.project-name }}
        # WORKAROUND: Prevents accidentally overwriting `latest` + semver tags:
        # https://github.com/docker/metadata-action/issues/461#issuecomment-2686558629
        flavor: |
          latest=false
        # `edge` tag for development images in-between official releases:
        # Release tags (implicitly adds `latest` tag unless tag belongs to an older release series):
        tags: |
          type=edge,branch=${{ github.event.repository.default_branch }}
          # NOTE: Append `,value=${{ inputs.version }}` if not relying on a tag event from this repo
          type=semver,pattern={{major}},enable=${{ fromJSON(env.is-latest) && fromJSON(env.is-not-major-zero-ver) }}
          type=semver,pattern={{major}}.{{minor}}
          type=semver,pattern={{major}}.{{minor}}.{{patch}}
          type=raw,value=latest,enable=${{ env.is-latest }}
```

.github/workflows/container.yaml
```diff
@@ -0,0 +1,71 @@
...
+      with:
+        context: .
+        platforms: linux/amd64,linux/arm64
+        push: true
+        tags: ${{ steps.meta.outputs.tags }}
+        labels: ${{ steps.meta.outputs.labels }}
+        build-args: |
+          JUST_VERSION=${{ steps.version.outputs.version }}
+        cache-from: type=gha
+        cache-to: type=gha,mode=max
```
polarathene: Feedback was ignored?

Might as well drop the redundant caching?

```suggestion
```

Dockerfile
polarathene: Feedback was ignored?

README.md
polarathene: Feedback was ignored?

polarathene (changes requested):
Thanks for applying some of the earlier feedback! 😁

I'm a bit unsure why earlier feedback was marked as resolved when the suggestions were dismissed without any context?

Regarding the tag handling, if you're going to add some precautions like that you should probably include `flavor.latest=false` as an example I shared demonstrates.

I think while it's probably a hassle this publish workflow should probably get some confirmation that it's triggered and successful when a separate workflow triggers a published release like `release.yaml` does.

README.md
polarathene: This was suggested earlier but dismissed for some reason: https://github.com/casey/just/pull/2851#discussion_r2253628411

ulgens: It can be dismissed again, I'm okay with that. None of us is giving a blocking review here; they are just recommendations. 

Sorry for the duplication.

.github/workflows/container.yaml
elasticdotventures: lgtm.

README.md
elasticdotventures: @polarathene omg no! I started doing the commits here in github interface, then flipped to my editor for the bigger ones;  and I think I just dropped this and overwrote with a subsequent force push.   It was late, I wasn't expecting to make that many changes when I started.  Apologies.

elasticdotventures: also fwiw, this was actually my favorite comment.

README.md
elasticdotventures: I absolutely meant to make the change, it was late, it got lost when i flipped from github.com to my editor.

elasticdotventures: this was an oversight on my part. i agree with @polarathene and everybody, it was really late.

.github/workflows/container.yaml
elasticdotventures: cool. I'm keen to smoke test this.

elasticdotventures:
> I'm a bit unsure why earlier feedback was marked as resolved when the suggestions were dismissed without any context?

@polarathene Doh! My bad.  .. I had started to do them in github.com, but at a few in, I was like omg, these comments are huge .. and I flipped to the branch on my lappy, retyped the changes, it was late, force pushed, missed one.

.github/workflows/container.yaml
elasticdotventures: @polarathene /agree - workflow_call is more reliable.  updated.

.github/workflows/container.yaml
```diff
@@ -0,0 +1,92 @@
...
+        images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+        # WORKAROUND: Prevents accidentally overwriting `latest` + semver tags
+        # https://github.com/docker/metadata-action/issues/461#issuecomment-2686558629
+        flavor: |
+          latest=false
+        tags: |
+          type=semver,pattern={{version}},value=${{ steps.version.outputs.version }}
+          type=semver,pattern={{major}}.{{minor}},value=${{ steps.version.outputs.version }},enable={{is_default_branch}}
+          type=semver,pattern={{major}},value=${{ steps.version.outputs.version }},enable=${{ fromJSON(env.is-latest) && fromJSON(env.is-not-major-zero-ver) }}
+          type=raw,value=latest,enable=${{ env.is-latest }}
```
polarathene: > ```yaml
> if: github.event_name == 'release'
> ```

I am not sure about that condition?

Most of the time I guess it is probably correct to assume that the manual trigger via `workflow_dispatch` would be for older releases, although I'm not quite sure when this manual approach would be needed given this workflow is only intended to retrieve GH release artifact from a tag and publish an image with that?

Triggering for the latest tagged release would presumably already have a `latest` image pushed, but that would still warrant updating the `latest` tag then, so the conditional here seems it'd technically introduce a bug in publishing an image again?

A manual trigger probably makes the most sense when something about the image build itself went wrong (which isn't uncommon to experience), so a correction to the `Dockerfile` would resolve that as you're not checking out the associated source from git for that release (_assumed intentional_).

---

> ```yaml
> ,enable={{is_default_branch}}
> ```

[As expressed previously](https://github.com/casey/just/pull/2851#discussion_r2254152229), I'm not sure how well that works in practice?

Looking at the [action source for this feature](https://github.com/docker/metadata-action/blob/c1e51972afc2121e065aed6d45c65596fe445f3f/src/meta.ts#L444-L463), it's still a bit unclear what will happen at a glance. It makes note of it not being reliable when the event is a pushed tag, and it seems to query the branch name by removing the `refs\heads\` prefix to the repo's configured default branch on GH.

At a guess it would be `false` for the releases trigger (_or similar pushed tag event trigger_), while the manual dispatch regardless of version input, would be running this workflow from the default branch on the latest commit, so it would be expected to be `true`.

I think `<major>.<minor>` tag update was relatively safe (_assuming you weren't publishing an older patch release_), so you shouldn't need to guard against that (_for reference, [see this example of tagging concerns](https://github.com/docker/metadata-action/issues/407#issuecomment-2680650110)_).

polarathene: So it might look more like this:

```suggestion
    - name: 'Find the latest release tag (determined by semver sort)'
      id: latest-semver
      env:
        # Filter for semver tags (Ignores pre-release tags)
        TAG_FILTER: '^refs/tags/[0-9]+(\.[0-9]+){2}$'
        TAG_REF: ${{ steps.version.outputs.version }}
      shell: bash
      run: |
        TAG=$(
          git ls-remote --exit-code --refs --tags --sort='-v:refname' origin \
          | awk '{print $2}' | grep -E "${TAG_FILTER}" | head -n 1
        )
        echo "latest-tag=${TAG#refs/tags/}" >> "${GITHUB_OUTPUT}"
        echo "Latest semver tag found: ${TAG#refs/tags/}"

    - name: Extract metadata
      id: meta
      uses: docker/metadata-action@v5
      env:
        is-latest: ${{ steps.version.outputs.version == steps.latest-semver.outputs.latest-tag }}
        gh-release: ${{ steps.version.outputs.version }}
      with:
        images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
        # WORKAROUND: Prevents accidentally overwriting `latest` + semver tags
        # flavor+raw: https://github.com/docker/metadata-action/issues/461#issuecomment-2686558629
        # semver: https://github.com/docker/metadata-action/issues/407#issuecomment-2680650110
        flavor: |
          latest=false
        tags: |
          type=semver,pattern={{version}},value=${{ env.gh-release }}
          type=semver,pattern={{major}}.{{minor}},value=${{ env.gh-release }}
          type=semver,pattern={{major}},value=${{ env.gh-release }},enable=${{ env.is-latest }}
          type=raw,value=latest,enable=${{ env.is-latest }}
```

But realistically given how simple the Docker image and release management process for this project is, you could assume that any trigger is always for the latest release, which the maintainer may prefer:

```suggestion
    - name: Extract metadata
      id: meta
      uses: docker/metadata-action@v5
      with:
        images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
        # NOTE: The `semver` tag type below with the default `flavor.latest=auto` action setting,
        # will implicitly append a `latest` tag to publish:
        # https://github.com/docker/metadata-action/issues/461#issuecomment-2680849083
        tags: |
          type=semver,pattern={{version}},value=${{ steps.version.outputs.version }}
          type=semver,pattern={{major}}.{{minor}},value=${{ steps.version.outputs.version }}
          type=semver,pattern={{major}},value=${{ steps.version.outputs.version }}
```

Otherwise, the more verbose approach provides better resilience if an image for a prior release needs to be built + published again.

.gitignore
```diff
@@ -15,3 +15,4 @@
 /test-utilities/Cargo.lock
 /test-utilities/target
 /tmp
+._b00t_.toml
```
polarathene: Context?

.github/workflows/container.yaml
polarathene: If you made a change to use `workflow_call` that is not present in the diff?

It would swap the release event trigger, requiring `release.yaml` to call `container.yaml` like in the two reference workflows I linked (_calling vs called parts_).

---

Actually, instead of `workflow_call`, you might be able to get away with `workflow_run` (_like [shown here](https://github.com/docker-mailserver/docker-mailserver/blob/69e2b56defe5135aa25125c083d61f960d0b0016/.github/workflows/docs-preview-deploy.yml#L3-L8)_).

That should avoid any need to modify `release.yaml`, and the upload/download advice for passing the built artifacts to your workflow here shouldn't matter if your approach to pull assets from GH Releases works.

```yaml
on:
  # This workflow runs off the primary branch which provides access to the `secrets` context:
  # Event triggered upon completion of the `release.yaml` workflow
  workflow_run:
    workflows: ['Release']
    types:
      - completed
  # Manual trigger
  workflow_dispatch:

# ...

jobs:
  container:
  if: ${{ github.event.workflow_run.conclusion == 'success' || inputs.version }}
```

I think the conditional should support bailing early if `release.yaml` fails, but still allow `workflow_dispatch` to be usable. You might want to confirm that the fallback being unset (null?) is treated as a falsey value, I haven't written workflows in a while and I know GHA expressions are sometimes iffy 😅

.github/workflows/container.yaml
```diff
@@ -0,0 +1,92 @@
...
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
```
polarathene, line 41: Have you tested this workflow? I believe you'll need to add QEMU for running `docker build` for ARM64 (_**NOTE:** This can technically be avoided by modifying the `Dockerfile`_).

```suggestion
    - name: 'Set up QEMU'
      uses: docker/setup-qemu-action@v3
      with:
        platforms: arm64

    - name: Set up Docker Buildx
      uses: docker/setup-buildx-action@v3
```

The other option is using native GH runner for ARM64, but that adds complexity that isn't worth it for a simple image build.

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# Download the appropriate just binary for the target architecture
+# Map Docker's platform names to just's release naming convention
+RUN set -eux; \
+    case "$TARGETARCH" in \
+        "amd64") ARCH="x86_64-unknown-linux-musl" ;; \
+        "arm64") ARCH="aarch64-unknown-linux-musl" ;; \
+        *) echo "Unsupported architecture: $TARGETARCH" && exit 1 ;; \
+    esac; \
+    wget -O - "https://github.com/casey/just/releases/download/${JUST_VERSION}/just-${JUST_VERSION}-${ARCH}.tar.gz" \
+        | tar --directory /usr/local/bin --extract --gzip --file - --no-same-owner just
```
polarathene: Instead of QEMU action, when your `Dockerfile` can produce builds of each platform without any arch specific dependency, you can run those earlier stages as the same platform as the `docker build` host itself (_thus native without any emulation overhead_).

This is done by hinting what platform to pull an image from, where the hint is the BuildKit ARG `BUILDPLATFORM` for `FROM --platform`:

```Dockerfile
# syntax=docker/dockerfile:1

# NOTE: ARGs `BUILDPLATFORM` + `TARGETARCH` are implicitly defined by BuildKit (but must be declared after `FROM`):
# https://docs.docker.com/reference/dockerfile/#automatic-platform-args-in-the-global-scope
FROM --platform=${BUILDPLATFORM} alpine AS downloader
```

`TARGETARCH` will still be correct for your script to later match on, so it'll download the correct arch for each, and the final `scratch` image will then be pushed for each separate platform built 👍

If instead the project needed to build from source or rely on linking to system libraries via the package manager, it's still possible but gets a tad more complicated to support. The build and package installs would be notably faster without involving QEMU however.

polarathene: ## For reference

[Similar to the prior review feedback comment](https://github.com/casey/just/pull/2851#discussion_r2253538406) where I showed a useful technique for more complex `Dockerfile`s needing to map archs, the `FROM --platform` by default will match the target (_in this case from the `docker/build-push-action` setting, `platforms`_).

Just sharing for any reader that might be curious, `downloader-amd64` / `downloader-arm64` could omit their `--platform` below as it's implicit. `downloader-${TARGETARCH}` will select the appropriate earlier stage to match the one of the build host arch, and use that for all platform being built/published.

```suggestion
# syntax=docker/dockerfile:1

# NOTE: The ARG `TARGETARCH` is implicitly defined by BuildKit:
# https://docs.docker.com/reference/dockerfile/#automatic-platform-args-in-the-global-scope
# NOTE: BuildKit supplied ARGs use convention amd64 / arm64 instead of the desired x86_64 / aarch64
# https://itsfoss.com/arm-aarch64-x86_64
#
# Map arch naming conventions from BuildKit to Rust (TARGETARCH => RUST_ARCH):
FROM --platform=linux/amd64 alpine AS downloader-amd64
ARG RUST_ARCH=x86_64
FROM --platform=linux/arm64 alpine AS downloader-arm64
ARG RUST_ARCH=aarch64

# Fetch the expected version of `just` via GH Releases:
FROM downloader-${TARGETARCH} AS downloader
SHELL ["/bin/ash", "-eux", "-o", "pipefail", "-c"]
# This ARG will be set via GitHub Actions during release builds
ARG JUST_VERSION
ARG RELEASE_URL="https://github.com/casey/just/releases/download/${JUST_VERSION}/just-${JUST_VERSION}-${RUST_ARCH}-unknown-linux-musl.tar.gz"
RUN wget -O - "${RELEASE_URL}" \
    | tar --directory /usr/local/bin --extract --gzip --no-same-owner just
```

---

**UPDATE:** Slight mistake on my part.
- This example assumes QEMU for platforms not native to the build host, it's only a convenience for `RUST_ARCH` being set based on the target platform to build for, allowing it to be referenced via `ARG` across multiple instructions (_not just `RUN` where `$(uname -m)` can be used_).
- To switch to using only a native platform base image to avoid using QEMU (_and thus faster non-native builds when possible_) you would change both `--platform` to enforce using the build host platform with `--platform=${BUILDPLATFORM}`. Otherwise use the example as shown above with QEMU or native platform runners when you need the platform specific base image (_eg: for platform/arch specific libraries/packages_).

polarathene:
> omg, these comments are huge .. and I flipped to the branch on my lappy, retyped the changes, it was late, force pushed, missed one.

No worries, apologies for the excessive feedback, I rarely have the time to condense it (_I should probably learn how to use LLMs to help with that at some point_ 🤔).

Anyway, here's another round of review, but I've tried to save you the trouble with a checklist in this comment and snippets in the collapsed overview that you can roughly copy/paste to apply.

The individual review comments provide more justification to support the change requests.

---

**Actionables:**
- [ ] `WORKDIR` removal from `Dockerfile` (preferred), or documentation added to `README.md` (_if verified to have a valid working use-case_).
- [ ] Apply one of the change suggestions for semver tag config, the smaller variant is preferred.
- [ ] Switch from `on.release` to `on.workflow_run`.
- [ ] Add QEMU setup action to workflow, or modify `Dockerfile` to build only with native execution (preferred).
- [ ] Remove [`_b00t_.toml` from `.gitignore`](https://github.com/casey/just/pull/2851/commits/ff8e6f4e7f8ca174e16965b0a4d5ac8e2fe94590), or provide context to why it was added.

<details>
<summary>Change request overview (click to view)</summary>

You should either remove the `WORKDIR` instruction, or [document it in the README section as requested](https://github.com/casey/just/pull/2851/files#r2253605170) for those that would use the published container image to run `just` (_**EDIT:** [won't be functional](https://github.com/casey/just/issues/1497#issuecomment-1382322473) though?_), rather than only as convenient form of packaging for other images to pull from.

**UPDATE:** [Since a shell is required in the image to use `just`](https://github.com/casey/just#prerequisites), it seems better to remove `WORKDIR` and let others extend the `Dockerfile` to manage that as they see fit (_a static shell binary like NuShell could be added + configured_).

Another valid documentation example is instead of the `COPY` instruction, if you don't need `just` at runtime but only during the image build, it could be temporarily sourced via the published image (_like I [demonstrated how a hypothetical `eget` image could](https://github.com/casey/just/pull/2851#discussion_r2253598877)_):

```Dockerfile
FROM alpine

# Run `just` from the official image with a local `justfile` from your build context without needing a `COPY`:
RUN \
  --mount=type=bind,from=ghcr.io/casey/just:latest,source=/usr/local/bin/just,target=/usr/local/bin/just \
  --mount=type=bind,source=./justfile,target=./justfile \
    just do-it
```

From `scratch` image example (_could alternatively extend the `just` image_), with NuShell added:

```Dockerfile
# NOTE: Instead of `scratch`, alternatively extend the `just` image for a simpler `RUN` statement
FROM scratch
COPY --link --from=ghcr.io/nushell/nushell /usr/bin/nu /usr/local/bin/nu
SHELL ["/usr/local/bin/nu", "-c"]

WORKDIR /example
COPY <<HEREDOC /example/justfile
set shell := ["/usr/local/bin/nu", "-c"]

do-it:
  echo 'hello world'
HEREDOC

RUN --mount=type=bind,from=ghcr.io/casey/just:latest,source=/usr/local/bin/just,target=/usr/local/bin/just \
    just do-it
```

Build it and it'll execute that `justfile` via `RUN` successfully:

```bash
docker build --progress plain --tag localhost/example .

# ...
echo 'hello world'
hello world
```

---

I think there was a bit of a misunderstanding with [this feedback](https://github.com/casey/just/pull/2851#discussion_r2254152229) where I demonstrate how to be more thorough with semver concerns I've seen in the wild 😅 

As `just` is a much smaller project and is highly unlikely to maintain multiple release branches for backports AFAIK, the project is probably quite safe from those concerns (_unlike projects such as Tyk which had several mishaps with their release pipeline and GH releases `latest` / tagging concerns that affected their image tags incorrectly_). `just` only ever seems to have sequential releases at a glance.

Likewise, as I was providing an existing workflow snippet I have as reference material, it included another condition for major zero-ver releases which is no longer applicable to `just`, so that is redundant (_assuming no images would be built for these old releases via the manual `workflow_dispatch` trigger_).

To get a maintainer to feel more comfortable with approving a PR like this, it would be better to minimize what is being offloaded to them. I'm partly to blame here for not communicating more clearly in the review feedback 😓

I have provided two suggestions for semver tag handling:
- More resilient (_adjusted to this repo_) with fixes for the bugs introduced.
- Simple (_no identifying the latest semver tag_), should be suitable for this project 👍 (_unless the maintainer prefers the added complexity_)

Simple variant:

```yaml
    - name: Extract metadata
      id: meta
      uses: docker/metadata-action@v5
      with:
        images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
        # NOTE: The `semver` tag type below with the default `flavor.latest=auto` action setting,
        # will implicitly append a `latest` tag to publish:
        # https://github.com/docker/metadata-action/issues/461#issuecomment-2680849083
        tags: |
          type=semver,pattern={{version}},value=${{ steps.version.outputs.version }}
          type=semver,pattern={{major}}.{{minor}},value=${{ steps.version.outputs.version }}
          type=semver,pattern={{major}},value=${{ steps.version.outputs.version }}
```

---

Consider swapping the `on.release` workflow trigger for `on.workflow_run` instead:

```yaml
on:
  # Event triggered upon completion of the `release.yaml` workflow
  workflow_run:
    workflows: ['Release']
    types:
      - completed

  # Manual trigger
  workflow_dispatch:
  # ...

# ...

jobs:
  container:
  if: ${{ github.event.workflow_run.conclusion == 'success' || inputs.version }}
```

I think that'd be more reliable for triggering correctly after a release is published via the `release.yaml` workflow (_tag push event trigger_).

---

Presumably the workflow hasn't been verified to build/publish the ARM64 image? At least at a glance I think you'll need to either:

Add QEMU to your workflow setup to support emulating ARM64 image execution:

```yaml
    - name: 'Set up QEMU'
      uses: docker/setup-qemu-action@v3
      with:
        platforms: arm64
```

or modify the `Dockerfile` to build ARM64 image from an AMD64 base image (_since the final stage is `scratch` with only the platform specific binary_):

```Dockerfile
# syntax=docker/dockerfile:1

# NOTE: ARGs `BUILDPLATFORM` + `TARGETARCH` are implicitly defined by BuildKit (but must be declared after `FROM`):
# https://docs.docker.com/reference/dockerfile/#automatic-platform-args-in-the-global-scope
FROM --platform=${BUILDPLATFORM} alpine AS downloader
```

</details>

polarathene (changes requested):

.gitignore
elasticdotventures: ah yeah .. well, this is specific to my workflow; the lack of this line was why "DOH" happened, and your changes are overwritten.  i committed to prevent it from happening again. 

once the pr is finalized, or .. whatever then I can remove it - .. the problem is it's created on my side with git hooks, and it gets updated frequently.  ..

._b00t_.toml is a semi-stealth project that tracks my personal metadata about the repo, time spent, etc.  it should never be committed, it's like a .env file, but tracks metadata about a dev session. 

my problem is that .gitignore don't work until you commit them.  ;-) .. so I can't actually ignore a file without committing it to the .gitignore, .. after writing this i'm wondering if I could move it to .git/ and it wouldn't be tracked?  i'll try that and remove this.

polarathene: I think in that case what you actually want is a [global `.gitignore` configured](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/) rather than adding per-repo `.gitignore`?

elasticdotventures:
@polarathene 
plz review

.github/workflows/container.yaml
```diff
@@ -0,0 +1,85 @@
+name: Container Image
+
+on:
+  # Event triggered upon completion of the `release.yaml` workflow
+  workflow_run:
+    workflows: ['Release']
+    types:
+      - completed
```
polarathene: You are using `workflow_call` so no need for `workflow_run`.

```suggestion
```

---

That wouldn't work as-is btw as it lacks any input for acquiring the tag/version that you need later on for `wget`.

I believe I mentioned that would need to use the complimentary upload/download actions to volley over that information from `release.yaml`, so `workflow_call` was more appropriate anyway.

.github/workflows/container.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+  # Manual trigger
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to build (e.g., 1.42.4)'
+        required: true
+        type: string
+
+  workflow_call:
```
polarathene, line 17: Maybe won't hurt to add a bit of context to readers who might be less familiar with how `workflow_call` works at a glance?

```suggestion
  # `release.yaml` calls this workflow after it has published a new release and attached files.
  # NOTE: As `release.yaml` is triggered via a tag being pushed, this workflow inherits
  # that push event context (github.{event_name, event, ref_type, ref_name, ...})
  workflow_call:
```

.github/workflows/container.yaml
```diff
@@ -0,0 +1,85 @@
+name: Container Image
+
+on:
+  # Event triggered upon completion of the `release.yaml` workflow
+  workflow_run:
+    workflows: ['Release']
+    types:
+      - completed
+
+  # Manual trigger
```
polarathene: Likewise, while at it this might be obvious in some parts or helpful in others, (_the maintainer can trim the comment once they review, but it should otherwise make it easy to quickly grok during review_):

```suggestion
  # Manual trigger via GH Actions page, runs this workflow from a selected tag or branch,
  # which also affects the git clone for the `Dockerfile` used, and `github.ref_type` (branch / tag)
```

.github/workflows/container.yaml
```diff
@@ -0,0 +1,85 @@
...
+        required: true
+        type: string
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  container:
+    if: ${{ github.event.workflow_run.conclusion == 'success' || inputs.version }}
```
polarathene: Dropping `workflow_run`, this won't be necessary.

```suggestion
```

.github/workflows/container.yaml
```diff
@@ -0,0 +1,85 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
```
polarathene, line 33: I would need to check, but I think `permissions` are dependent upon what is permitted top-level? Not sure if it matters in this workflow, but when `release.yaml` calls this workflow via `workflow_call`, `release.yaml` will enforce it's `permissions` as the baseline, so you can only reduce the scope of permissions from what is granted there. In some cases `packages` is not `write` by default.

On newer repos, AFAIK the default is read-only permissions but either the `permissions` for `casey/just` are legacy or were explicitly relaxed by the maintainer, as can be [seen here](https://github.com/casey/just/actions/runs/16509993867/job/46689828155#step:1:19) with all the `write` permissions granted:

> image

So for this project specifically, I think these `permissions` should be valid (_if they instead defaulted to `read-only`, both permissions would need to be declared at `release.yaml` too_).

---

**UPDATE:** I checked with the `permissions` at the job scope. It's fine so long as the workflow has granted/inherited those permissions required.

.github/workflows/container.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: 'Set up QEMU'
+      uses: docker/setup-qemu-action@v3
+      with:
+        platforms: arm64
```
polarathene: You have the `Dockerfile` already building both AMD64 + ARM64 images with the native build platform of the build host, no emulation is needed for the `Dockerfile` in it's current state, so this can be dropped:

```suggestion
```

.github/workflows/container.yaml
```diff
@@ -0,0 +1,85 @@
...
+    - name: Build and push Docker image
+      uses: docker/build-push-action@v6
+      with:
+        context: .
+        platforms: linux/amd64,linux/arm64
+        push: true
+        tags: ${{ steps.meta.outputs.tags }}
+        labels: ${{ steps.meta.outputs.labels }}
+        build-args: |
+          JUST_VERSION=${{ steps.version.outputs.version }}
```
polarathene: With your changes to drop `on.release`, since you only have `inputs.version` to source the version from now, just use that directly?

```suggestion
    - name: Prepare image metadata
      id: metadata
      uses: docker/metadata-action@v5
      with:
        images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
        # NOTE: The `semver` tag type below (with the default `flavor.latest=auto` action setting),
        # will implicitly append a `latest` tag to publish:
        # https://github.com/docker/metadata-action/issues/461#issuecomment-2680849083
        tags: |
          type=semver,pattern={{major}},value=${{ inputs.version }}
          type=semver,pattern={{major}}.{{minor}},value=${{ inputs.version }}
          type=semver,pattern={{major}}.{{minor}}.{{patch}},value=${{ inputs.version }}

    - name: Build and push Docker image
      uses: docker/build-push-action@v6
      with:
        context: .
        platforms: linux/amd64,linux/arm64
        push: true
        tags: ${{ steps.metadata.outputs.tags }}
        labels: ${{ steps.metadata.outputs.labels }}
        build-args: |
          JUST_VERSION=${{ inputs.version }}
```

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    | tar --directory /usr/local/bin --extract --gzip --no-same-owner just
+
+# Use scratch for minimal final image - no OS, just the binary
+# This results in a ~10MB image vs ~50MB+ with a full OS
+FROM scratch
+COPY --from=downloader /usr/local/bin/just /usr/local/bin/just
+
+# Default to running just with help if no arguments provided
+ENTRYPOINT ["just"]
+CMD ["--help"]
```
polarathene, line 30: This decision to set a default `CMD` would prevent the expectation for [`default` recipe](https://github.com/casey/just#the-default-recipe) functionality. So perhaps `--help` as an implicit arg when no input is given is not ideal?
- I guess it's not a major issue since `just default` (_or whatever the first recipe is named_) would work just as well?
- Additionally since (_as explained in [this comments collapsed section](https://github.com/casey/just/pull/2851#issuecomment-3166541992)_) the `scratch` image lacks a shell it would not be able to run an actual recipe as-is... `--help` would work, but might also imply the image is meant to be usable standalone?

Without `CMD ["--help"]`:

```console
$ docker run --rm -it localhost/just
error: No justfile found

$ docker run --rm -it -v ./justfile:/justfile:ro localhost/just
error: Recipe `example` could not be run because just could not find the shell: No such file or directory (os error 2)
```

Those seem to be more informative? The README might need to still clarify the lack of a shell in the image to avoid users opening support requests about it? 😓

---

Alternatively, you could avoid using a `scratch` image for the final stage, and instead retain the Alpine image base (8.3MB) which isn't that much more when the `just` binary is 4.9MB (_I'm not sure why it appears to be closer to 10MB for you as noted in your `Dockerfile` comment_).

BTW, it's also possible to get a minimal glibc base with `bash` as the shell too that only weights 6.6MB before `just` is added:

<details>
<summary>Minimal glibc image with bash</summary>

```Dockerfile
# ...

FROM alpine AS get-chisel
ARG CHISEL_VERSION=1.2.0
ARG TARGETARCH
ARG CHISEL_RELEASE="https://github.com/canonical/chisel/releases/download/v${CHISEL_VERSION}/chisel_v${CHISEL_VERSION}_linux_${TARGETARCH}.tar.gz"
# NOTE: `--no-same-owner` used as `chisel` release has ownership of `1001:128`
RUN wget -qO - "${CHISEL_RELEASE}" | tar -xz --no-same-owner -C /usr/local/bin chisel

FROM get-chisel AS root-fs
WORKDIR /root-fs
RUN --mount=type=cache,target=/.cache/chisel,id="chisel-cache" \
  chisel cut --release ubuntu-24.04 --root /root-fs base-files_base bash_bins && \
  ln -s bash /root-fs/usr/bin/sh

FROM scratch
COPY --link --from=root-fs /root-fs /
COPY --link --from=downloader /usr/local/bin/just /usr/local/bin/just
ENTRYPOINT ["/usr/local/bin/just"]
```

</details>

But that is only marginally smaller than Alpine and it lacks common commands one might expect to be available, along with not providing a package manager. So not really viable for `just`.

Alpine would make more sense for those wanting to use the `just` image as-is or alternatively as a base image to extend with their own `Dockerfile` rather than `COPY` the binary from it.

polarathene, line 30: Assuming you agree with the prior `Dockerfile` comment on `CMD` + `scratch`, this will improve usability of the image directly at the expense of going from 5MB to 13MB (_less in network overhead due to compression at the image registry_), seems like a reasonable tradeoff:

```suggestion
ENTRYPOINT ["just"]
```

Otherwise update the README to inform users about the image not being intended for direct usage, so they don't burden the maintainer with support requests.

.github/workflows/release.yaml
```diff
@@ -207,3 +207,12 @@ jobs:
         github_token: ${{secrets.GITHUB_TOKEN}}
         publish_branch: gh-pages
         publish_dir: www
+
+  container:
+    uses: ./.github/workflows/container.yaml
+    needs:
+      - checksum
+      - prerelease
```
polarathene: This is incorrect.
- You're not dependent upon the `checksum` job (_which uploads checksums after `package` job completes successfully_), but you are dependent upon `package` completing.
- Likewise for `prerelease` job, except images should probably only be for proper releases, so this would be fine if you added a conditional guard on this jobs output.

```suggestion
    needs:
      - package
      - prerelease
    # Avoid publishing images for pre-releases (NOTE: `value` is a string, not a boolean)
    if: ${{ needs.prerelease.outputs.value == 'false' }}
```

.github/workflows/release.yaml
```diff
@@ -207,3 +207,12 @@ jobs:
         github_token: ${{secrets.GITHUB_TOKEN}}
         publish_branch: gh-pages
         publish_dir: www
+
+  container:
+    uses: ./.github/workflows/container.yaml
```
polarathene, line 215: No action required, just informative for those unaware.

---

**For reference:** This will run the `container.yaml` workflow associated to the commit of this workflows trigger (_the tagged commit_), just as `release.yaml` itself does.
- For this project in particular and the way this `container` job is triggered, this isn't a concern since it's unlikely to have a release with an older `Dockerfile` checkout being used.
- In other projects where the trigger isn't the latest commit on the default branch being tagged, but some other release series branch, you'd have to keep in mind the risk of divergence (_**note:** there is a similar caveat with `workflow_dispatch` at `container.yaml` when selecting to run from a different ref instead of the default branch offered_).
- This can be worth being mindful of when `workflow_run` is involved, since it instead operates from the latest commit (head ref) at the default branch, regardless of the commit that the workflow it depends on was using (_`release.yaml` in this case_).

However if it was desirable to fix `container.yaml` or the `Dockerfile` as a CI fix in a commit that landed after the commit to be tagged for release, _before_ that release was triggered via the maintainer pushing the tag... then it'd be useful to ensure this were adjusted.

That said, in other contexts doing so can be problematic when other branches are used such as a PR update for a workflow that is triggered by PRs, or breaking changes affecting other long-lived branches or forks (_if neither patched workflows that weren't restricted to the default branch to trigger_).

The alternative with a specific git ref would look like this (_the owner + repo prefix is mandatory however and expressions cannot be used for the benefit of forks or copy/paste_):

```yaml
 container:
    uses: casey/just/.github/workflows/container.yaml@master
```

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+ARG BUILDARCH_RUST=x86_64
+FROM --platform=linux/arm64 alpine AS downloader-arm64
+ARG BUILDARCH_RUST=aarch64
+
+# Fetch the expected version of `just` via GH Releases:
+FROM downloader-${BUILDARCH} AS downloader
+SHELL ["/bin/ash", "-eux", "-o", "pipefail", "-c"]
+# This ARG will be set via GitHub Actions during release builds
+ARG JUST_VERSION
+ARG RELEASE_URL="https://github.com/casey/just/releases/download/${JUST_VERSION}/just-${JUST_VERSION}-${BUILDARCH_RUST}-unknown-linux-musl.tar.gz"
```
polarathene: I made a mistake here as [updated in comment for the original suggestion you applied](https://github.com/casey/just/pull/2851#discussion_r2261888590).

```suggestion
# Map arch naming conventions from BuildKit to Rust (TARGETARCH => RUST_ARCH):
FROM --platform=${BUILDPLATFORM} alpine AS downloader-amd64
ARG RUST_ARCH=x86_64
FROM --platform=${BUILDPLATFORM} alpine AS downloader-arm64
ARG RUST_ARCH=aarch64

# Fetch the expected version of `just` via GH Releases:
FROM downloader-${TARGETARCH} AS downloader
SHELL ["/bin/ash", "-eux", "-o", "pipefail", "-c"]
# This ARG will be set via GitHub Actions during release builds
ARG JUST_VERSION
ARG RELEASE_URL="https://github.com/casey/just/releases/download/${JUST_VERSION}/just-${JUST_VERSION}-${RUST_ARCH}-unknown-linux-musl.tar.gz"
```

polarathene:
<details>
<summary>Collapsed as outdated response</summary>

**NOTE:** Below was a response I was preparing before I started a review, you later pushed changes and I did not wrap up this comment in time prior so I switched to a review.

You can probably can ignore this entire comment. But might be of some value so I've retained it.

---

Ok... so I'm not entirely sure how you're approaching this PR on your end, but [my attempt at a checklist](https://github.com/casey/just/pull/2851#issuecomment-3166541992) didn't seem to get read (_presumably even less likely for the more verbose feedback_), and you just applied the first/all suggestions as-is from my review feedback? 🤔 

> image

The arrow points out a collapsed area that would have all the _preferred_ snippets available for an easy copy/paste in your editor.

For example, on the checklist (_and mentioned in individual feedback comments_) I suggested either [QEMU action for the workflow](https://github.com/casey/just/pull/2851/commits/740751243ccfc0a5e3789c70682dbe92d8091c08) or a [`Dockerfile` change](https://github.com/casey/just/pull/2851/commits/9f8c9392e581f5a7982ca3b90b35586818d08513), since only one should have been necessary. You applied both suggestions (_with the `Dockerfile` one [clearly under a "For reference" section header](https://github.com/casey/just/pull/2851#discussion_r2261888590), with a much simpler alternative shown.. but hey both work, so I guess that's fine_ 🤷‍♂️).

---

If you weren't wanting to manually make changes locally via your editor, you could also do this fully in the browser via the GH web editor (_there's a full workspace editor to modify multiple files, or you can use the other web UI to edit per file_):

> image

I am not sure if you're new-ish to PR review processes, but seeing as you applied each suggestion via the web UI individually, if you click the "Files changed" tab at the top of this PR page, it would take you to the view that I took the screenshot above from. In that view you can batch commit multiple suggestions within a single commit.

However you should really try grok what the suggestion was about rather than apply it blindly, which is all I can infer given the lack of interaction and how you went about it 😅 (_I'm not berating you btw, I'm just a terrible communicator.. there's lots to learn and do as a dev so I'm just assuming these are areas you're inexperienced in which is totally okay!_)

---

Since concerns were raised about inconsistent workflow config and potential for bugs but no verification was done, I've done so myself.

Regarding `on.release` event trigger:
- I can confirm that with how this repo is setup pushing a tag to trigger `release.yaml` to publish a release will not trigger `container.yaml` via `on.release`. The bottom of the release actions docs also [references this](https://github.com/actions/create-release/issues/71) to raise awareness about that concern.
- Additionally if publishing a release via the GH release to create the tag that way, the `on.release` trigger for `container.yaml` is also not feasible as it would attempt to run before `release.yaml` completes to update the published release with the assets `container.yaml` wants to retrieve for the `Dockerfile` build.

</details>

polarathene (changes requested):
Hey sorry for the late review, I had most of this prepared when you requested but have been caught up elsewhere, I can't quite recall the state of this feedback but I think I covered most things (_Might still need some verification_ 😅).

I'll hopefully have my own tasks elsewhere wrapped up so we can try get this PR in a good state 💪 

---

**NOTE:** The iterations of this PR perhaps better clarifies the concerns I expressed in the associated issue, despite your ["bloody easy" expectation](https://github.com/casey/just/issues/1497#issuecomment-3150921671) and [offer to help ease the maintenance burden](https://github.com/casey/just/issues/1497#issuecomment-3154829366)... this kinda thing can get tricky when it's not our own personal projects 😝 (_speaking from my own experiences since when a maintenance issue does prop up, I often have to get familiar with all this stuff again since it's an area you ideally don't touch much again once setup, but it does happen_ 😓)

polarathene (approved):
LGTM 👍 @casey should be good to take over review from here.

---

I'd still [suggest avoiding the multi-stage to `scratch` image](https://github.com/casey/just/pull/2851#discussion_r2268522474) and just publish with the Alpine base image, it's not much more weight but makes it actually usable (_**EDIT:** actually this would require modifying `FROM scratch` to `FROM alpine` due to the first stage avoiding emulation via `FROM --platform=${BUILDPLATFORM}`, publishing that for a non-native arch would have the wrong base_).

If you're really against doing so, then you really should update the README for [reasons covered in prior review feedback](https://github.com/casey/just/pull/2851/files#r2268517666).

---

Alternatively the `container.yaml` workflow could also just handle the retrieval of the release archive instead (_if this `Dockerfile` would only be intended for CI usage_), then the `Dockerfile` would be simplified to something like:

```Dockerfile
FROM alpine
ARG TARGETARCH
COPY ./releases/${TARGETARCH}/just /usr/local/bin/just
ENTRYPOINT ["just"]
```

With `container.yaml` adding an extra step to fetch the two archives and extract to `./releases/{amd64,arm64}/just`.

The current `Dockerfile` is more convenient though as-is.

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    | tar --directory /usr/local/bin --extract --gzip --no-same-owner just
+
+# Use scratch for minimal final image - no OS, just the binary
+# This results in a ~10MB image vs ~50MB+ with a full OS
+FROM scratch
+COPY --from=downloader /usr/local/bin/just /usr/local/bin/just
+
+# Default to running just with help if no arguments provided
+ENTRYPOINT ["just"]
+CMD ["--help"]
```
polarathene, line 30: As [previously detailed in feedback](https://github.com/casey/just/pull/2851/files#r2268517666).

---

5MB => 13MB increase, but the image can be used directly.

```suggestion
FROM alpine
COPY --from=downloader /usr/local/bin/just /usr/local/bin/just
ENTRYPOINT ["just"]
```

No need to update README to mention the lack of a shell if adopting this change.

elasticdotventures:
@ulgens i think your review is blocking.

.github/workflows/container.yaml
```diff
@@ -0,0 +1,72 @@
...
+  container:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
```
ulgens, line 38: Extra new line.

.github/workflows/container.yaml
```diff
@@ -0,0 +1,72 @@
...
+    - name: Build and push Docker image
+      uses: docker/build-push-action@v6
+      with:
+        context: .
+        platforms: linux/amd64,linux/arm64
+        push: true
+        tags: ${{ steps.metadata.outputs.tags }}
+        labels: ${{ steps.metadata.outputs.labels }}
+        build-args: |
+          JUST_VERSION=${{ inputs.version }}
```
ulgens, line 72: Missing file ending.

Dockerfile
ulgens, line 30: Missing file ending.

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+SHELL ["/bin/ash", "-eux", "-o", "pipefail", "-c"]
+# This ARG will be set via GitHub Actions during release builds
+ARG JUST_VERSION
+ARG RELEASE_URL="https://github.com/casey/just/releases/download/${JUST_VERSION}/just-${JUST_VERSION}-${RUST_ARCH}-unknown-linux-musl.tar.gz"
+RUN wget -O - "${RELEASE_URL}" \
+    | tar --directory /usr/local/bin --extract --gzip --no-same-owner just
+
+# Use scratch for minimal final image - no OS, just the binary
+# This results in a ~10MB image vs ~50MB+ with a full OS
+FROM scratch
```
ulgens, line 25: Was the decision not to specify a version intentional?

ulgens (changes requested):
Added my comments.

.github/workflows/container.yaml
```diff
@@ -0,0 +1,72 @@
...
+        required: true
+        type: string
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  container:
+    runs-on: ubuntu-latest
```
ulgens, line 30: Was the decision not to specify a version intentional?

Dockerfile
polarathene: ? Why would `scratch` have a version?

ulgens: For reproducible builds? Not only should it have a version, but it's also important to have a pinned hash. If that wasn't the decision here, I'm okay - just wanted to clarify what is the choice.

polarathene: Please show me a `Dockerfile` where `scratch` has a version/digest associated to it? It's literally an empty base, which as implied is "from scratch" (no base content).

Even the comment context above it points out that only the `just` binary will be present in the image (_I've discouraged this and suggested the PR author keep the `alpine` base, but that feedback has been ignored twice_).

ulgens: I think you are misinterpreting my message. This wasn't a change request; I was just asking. I don't normally work with the said image so I don't know about its content.

> Was the decision not to specify a version intentional?

It seems intentional - so okay.

ulgens:
I have one small question left about `ubuntu-latest` and I believe it's better to have the fix file endings before the merge. Other than that, looks good.

Dockerfile
polarathene: > I think you are misinterpreting my message. This wasn't a change request; I was just asking. 

You seemed to be rather confident about your questioning with:

> _For reproducible builds? Not only should it have a version, but it's also important to have a pinned hash._

I was just asking for you to clarify why you believed that statement so confidently as if you knew what you were talking about rather than parroting.

I know you meant well, I was just a tad surprised that you know about those concerns but not what `scratch` is. Apologies if my communication style comes off a bit blunt/harsh 😅 

---

> I don't normally work with the said image so I don't know about its content.

It's nothing specific to this image (or content), it's literally `scratch`, this is common/basic `Dockerfile` knowledge IMO 🤷‍♂️ (_at least if you've ever written an multi-stage image where the final artifact is a static binary, such that it needs no runtime dependencies_)

If the binary fetching from GH releases was not handled in the `Dockerfile` itself we'd have only this very simple 2 line `Dockerfile`:

```Dockerfile
FROM scratch
COPY ./just /usr/local/bin/just
```

<details>
<summary>Rant</summary>

Which is all for the convenience of [avoiding a few lines](https://github.com/casey/just/issues/1497) by offloading the burden of maintaining the contributed files this PR provides to the `just` maintainer, which if there last stance and lack of engagement in this PR is anything to go by, they're not that interested.

It was believed by the PR author that this contribution would be simple/easy, yet as can be seen from past review cycles that was not the case, and so to believe that it will be no burden offloading future maintenance to someone else who may likewise lack the expertise to identify/resolve problems like were caught in review here, is not all that nice of an expectation.

Granted it's more ideal for such an image to be released / maintained through an official channel than someone in the community.

</details>

---

> It seems intentional - so okay.

No version or digest exists for `scratch`. I guess you learned something new today :)

ulgens: @polarathene Thanks, but you don't need to spend your precious time educating me, especially after I already said I got my answer. I do understand you have more knowledge about the topic, and you seem extremely proud of it —good for you.

Thanks for your time.

casey:
Implemented in #3174. I really had to do this myself to understand exactly what the purpose was, and the best most minimal way to do it, since i don't have much experience with Docker.

PR #2852, opened 2025-08-06
Print executed commands if run in verbose mode

tgross35:
It is currently difficult to debug what went wrong in cases such as OS exec errors, because verbose output prints the file but does not say how it is run. Update spawning to print executed commands with loquacious and above verbosity. The result when run with `-v` is something like the following:

    + cd "/home/user/workspace" && "/run/user/1000/just/just-kynaKD/configure"

The `+` syntax is meant to match shell scripts run with `-x`. This gets coloring to differentiate it.

Environment variables are not printed unless `-vv` is set:

    + cd "/home/user/workspace" && RUST_BACKTRACE="1" CARGO_HOME="/cargo-home" "/run/user/1000/just/just-r2rLtg/configure"

### DIFF
--- a/src/command_ext.rs
+++ b/src/command_ext.rs
@@ -11,11 +11,11 @@ pub(crate) trait CommandExt {
 
   fn export_scope(&mut self, settings: &Settings, scope: &Scope, unexports: &HashSet<String>);
 
-  fn output_guard(self) -> (io::Result<process::Output>, Option<Signal>);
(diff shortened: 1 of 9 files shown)


casey (changes requested):
Sorry for letting this sit! I think this is good, but I think that it's noisy enough that it should bet its own flag `--trace`, and not be part of `-v`. I would suggest using `-vv`, but `-vv` prints scripts, which is very annoying.

casey:

PR #2855, opened 2025-08-08
feat: replace fzf with skim as the default chooser

### LINKED ISSUE #939: Default to using `skim` as library when invoking chooser, to avoid dependency on `fzf` being installed
[`skim`](https://github.com/lotabout/skim) is a popular fuzzy finder written in rust with a library interface. We should see if we can use that by default instead of calling `fzf`.

bestgopher:
Closes: #939

https://github.com/user-attachments/assets/8774dc7b-d7e6-4593-a666-68c92ed6a459

### DIFF
--- a/src/chooser.rs
+++ b/src/chooser.rs
@@ -1,0 +1,127 @@
+use crate::config::Config;
+use crate::error::Error;
+use crate::justfile::Justfile;
+use crate::recipe::Recipe;
+use crate::search::Search;
+use skim::options::SkimOptionsBuilder;
+use skim::prelude::SkimItemReader;
+use skim::Skim;
+use std::ffi::OsString;
+use std::io::{Cursor, Write};
+use std::process::Stdio;
+
+pub enum Chooser<'a, 's: 'a> {
+  Command {
+    chooser: OsString,
+    justfile: &'a Justfile<'s>,
+    search: &'a Search,
+    config: &'a Config,
+  },
+  Skim,
+}
+
+impl<'a> Chooser<'_, 'a> {
+  fn run_with_command(self, recipes: Vec<&Recipe>) -> Result<Vec<String>, Error<'a>> {
+    let Self::Command {
+      chooser,
+      justfile,
+      search,
+      config,
+    } = self
+    else {
+      panic!("must be command chooser")
+    };
+
+    let result = justfile
+      .settings
+      .shell_command(config)
+      .arg(&chooser)
+      .current_dir(&search.working_directory)
+      .stdin(Stdio::piped())
+      .stdout(Stdio::piped())
+      .spawn();
+
+    let mut child = match result {
+      Ok(child) => child,
+      Err(io_error) => {
+        let (shell_binary, shell_arguments) = justfile.settings.shell(config);
+        return Err(Error::ChooserInvoke {
+          shell_binary: shell_binary.to_owned(),
+          shell_arguments: shell_arguments.join(" "),
+          chooser,
+          io_error,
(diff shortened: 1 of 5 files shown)


bestgopher:
I don't know how to fix some test case errors.

bestgopher:
@casey When you have a chance, could you take a look.

src/chooser.rs
```diff
@@ -0,0 +1,127 @@
+use crate::config::Config;
```
casey, line 1: You can use `use super::*` and get rid of a lot of these imports.

src/chooser.rs
```diff
@@ -0,0 +1,127 @@
...
+    justfile: &'a Justfile<'s>,
+    search: &'a Search,
+    config: &'a Config,
+  },
+  Skim,
+}
+
+impl<'a> Chooser<'_, 'a> {
+  fn run_with_command(self, recipes: Vec<&Recipe>) -> Result<Vec<String>, Error<'a>> {
+    let Self::Command {
```
casey, line 25: I would pass these arguments in instead of taking self.

src/chooser.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    let recipes = stdout
+      .split_whitespace()
+      .map(str::to_owned)
+      .collect::<Vec<String>>();
+
+    Ok(recipes)
+  }
+
+  fn run_with_skim(recipes: Vec<&Recipe>) -> Result<Vec<String>, Error<'a>> {
```
casey, line 90:
```suggestion
  fn choose_with_skim(recipes: Vec<&Recipe>) -> Result<Vec<String>, Error<'a>> {
```

src/chooser.rs
```diff
@@ -0,0 +1,127 @@
...
+    chooser: OsString,
+    justfile: &'a Justfile<'s>,
+    search: &'a Search,
+    config: &'a Config,
+  },
+  Skim,
+}
+
+impl<'a> Chooser<'_, 'a> {
+  fn run_with_command(self, recipes: Vec<&Recipe>) -> Result<Vec<String>, Error<'a>> {
```
casey, line 24:
```suggestion
  fn choose_with_command(self, recipes: Vec<&Recipe>) -> Result<Vec<String>, Error<'a>> {
```

src/chooser.rs
```diff
@@ -0,0 +1,127 @@
...
+      .collect::<Vec<String>>();
+
+    Ok(recipes)
+  }
+
+  fn run_with_skim(recipes: Vec<&Recipe>) -> Result<Vec<String>, Error<'a>> {
+    let options = SkimOptionsBuilder::default()
+      .height(String::from("100%"))
+      .multi(true)
+      .preview(Some("just --unstable --color always --show {}".to_string()))
```
casey, line 94: I'd like to avoid a recursive call here. Can we pass in both the items and their previews?

src/chooser.rs
```diff
@@ -0,0 +1,127 @@
...
+      recipes
+        .iter()
+        .map(|r| r.spaced_namepath())
+        .collect::<Vec<_>>()
+        .join("\n"),
+    ));
+
+    // `run_with` would read and show items from the stream
+    let selected_items =
+      Skim::run_with(&options, Some(items)).map_or_else(Vec::new, |out| out.selected_items);
```
casey, line 111: We should not swallow errors.

casey (changes requested):
Sorry for letting this sit. See comments!

casey:

PR #2858, opened 2025-08-15
docs: fix PATH_SEP value in README for Windows

mabdelaal86:
PATH_SEP value for Windows "/" is not showen in the documentation as the / is not skipped.


casey:
Thanks for the PR! Changes to this table need to be made in two places, since there's a test which ensures that the table is kept in sync with the actual constants. This PR doesn't allow me to push changes to it, I believe because it's the master branch, so I made the change in #2859.

Issue #2406, opened 2024-10-02
Improve dynamic completions


casey:
Just previously used clap's statically generated completion scripts. These scripts would complete flags and options, but wouldn't complete recipes, since that requires invoking just itself to find and parse the justfile. To remedy this, we started patching the generated completion scripts.

This slowly turned into a maintainance nightmare, as the patches grew in number and complexity, and had to be manually adapted every time the base completion scripts changed.

Clap has a new dynamic completion engine, tracked in https://github.com/clap-rs/clap/issues/3166.

In #3167 I switched to the new completion engine. Strictly speaking, this was a regression, since we lost the ability to complete recipes, but I didn't want to try to do everything in one PR, so I thought it was better to merge it and add back the missing functionally in follow-up PRs.

To do:
- [ ] Bash splits completion arguments on `:`, which breaks completion of `::`-separate recipe paths. We should upstream a fix to `clap_complete`, since it's basically never correct to split on `:`. I have a draft PR open for this: https://github.com/clap-rs/clap/pull/6311
- [x] Complete arguments with variable names (suffixed with `=`, recipes, and file paths (since they could be recipe arguments)
- [ ] Complete arguments more intelligently. I'm not sure how much we can do here, but it would be ideal to only complete overrides if we haven't gotten a recipe name or an argument, and to only complete recipes when we're in a position where we're expecting a recipe (i.e., we're not in position where we're completing a recipe argument). This may be tough. We don't know where exactly in the positional arguments we're completing, so we don't necessarily know what we're completing. However, this enables a bunch of stuff, like showing the parameter name when completing a recipe parameter, or completing flags and options.
- [x] Complete variable names for `--set`
- [ ] Support proper completions for nushell
- [ ] Don't complete variable names in value position for `--set`, i.e., `--set foo <tab>`
- [x] Complete recipes for `--show`
- [ ] Look through all arguments for opportunities to improve completions (e.g., only complete files or directories, or add a new custom completer)
- [x] Add instructions for each supported shell for how to load shell completion scripts, including how to lazy load them
- [x] Close all PRs and issues related to the old completion scripts
- [ ] The patched zsh completion script had a nice feature where if pressed tab after completing a recipe, it would show you the recipe as a hint message. This was nice, we should try to replicate it.
- [ ] When completing recipes, should we show the docs, the parameters, or both?
[...]

PR #2861, opened 2025-08-16
fix fish completions with words containing `-f`

### LINKED ISSUE #2686: Completions not working in fish shell
## Steps to reproduce

When I have a justfile like this:

```just
run-frontend:
  echo "just frontend"
```

Then I typed `just run-fr` and pressed tab to complete, some errors occured:

```bash
just run-frerror: Failed to read justfile at `/xxx/tmp/r`: No such file or directory (os error 2)
```

But if I typed `just run-f` or less letters, completions work fine. Other similar commands such as `run-backend` will not trigger this problem.

## Environments

System: MacOS 15.3.2
Shell: fish 4.0.1
just: 1.40.0

azarmadr:
Fixes #2686


azarmadr:
It is a small change. only adding a regex to filter out `-f` in between words

casey:

I think this should work with the new engine.

PR #2886, opened 2025-09-18
Add loongarch64 support

SkyBird233:
The [loongarch64 binary](https://github.com/SkyBird233/just/releases/tag/v0.9.4-1) built in [this workflow](https://github.com/SkyBird233/just/actions/runs/17821154019/job/50663740131) has been smoke tested.


eatradish:
ping?

casey:
Thanks for the PR! I removed the TODO, since I prefer to use issues instead. What's the difference between `gcc-14-loongarch64-linux-gnu` and `gcc-loongarch64-linux-gnu`?

SkyBird233:
> Thanks for the PR! I removed the TODO, since I prefer to use issues instead. What's the difference between `gcc-14-loongarch64-linux-gnu` and `gcc-loongarch64-linux-gnu`?

Hi @casey. `gcc-loongarch64-linux-gnu` is a metapackage that points to a specific version of GCC, e.g., in Ubuntu 24.10/25.04, it points to `gcc-14-loongarch64-linux-gnu`. However, the latest Ubuntu version of GitHub Actions is currently 24.04, and the metapackage `gcc-loongarch64-linux-gnu` doesn't exist in this version, so we have to use `gcc-$VERSION-loongarch64-linux-gnu` instead.

Also, GCC 14 is the first version to introduce vector support for LoongArch targets, as required by [Software Development and Build Convention for LoongArch Architectures](https://github.com/loongson/la-softdev-convention/blob/master/la-softdev-convention.adoc#112-compiler-vector-instruction-support-detection)

casey:
@SkyBird233 Got it, thanks! I created #2932 to track updating.

PR #2889, opened 2025-09-19
Support `[private]` attribute on modules

### LINKED ISSUE #2561: Allow `[private]` attribute on modules
To hide them from `--list`.

Scott-Guest:
Fixes #2561, allowing modules to be marked `[private]` so that they are hidden from `--list` and `--summary`.


casey:
Merged, finally! I made one behavior change, I didn't filter private modules when the user specifically requests one using `--list PATH`.

Issue #2616, opened 2025-02-03
Add `just_version` function to get the version of the `just` tool


nyurik:
While #2290 (minimal required just version) is being discussed, we may also want a separate path to ensure that the current runner meets our needs.  Let's add a new function `just_version()` that returns a semver of the current just tool - this way recipes could use [`semver_matches(just_version(), "...")`](https://just.systems/man/en/functions.html#semantic-versions) to see if just requirements are met for the project. Moreover, this could be added as some "prereq" recipe that most other recepes depend on - which will either succeed or fail with instructions to update just.

```just
[private]
validate_just:
    if [ {{semver_matches(just_version(), ">1.38")}} != 'true' ]; then \
        echo "Your just runner is too old, please update" ;\
        exit 1 ;\
    fi

foo: validate_just
    echo "running foo"
```

In the mean time, we could already use something like `{{semver_matches(shell(just_executable(), '--version'), ">1.38")}}` as  a temporary workaround until a new function is added

laniakea64:
Per https://github.com/casey/just/pull/2892#issuecomment-3366732435 :

The use case for a `just_version()` function is reporting or logging information about the execution environment for the purpose of making debugging easier.  The `just` version needs to be made available internally (e.g. a `just` function or builtin constant) because on Windows and in some CI environments, the currently-available method of getting `just` version by doing `{{quote(just_executable())}} --version` does not reliably work.

casey:
Added `just_version()` in #3497.

Issue #2290, opened 2024-08-02
Set minimal required `just` version for a justfile


nyurik:
`just` is rapidly evolving, frequently adding new features. `justfiles` are frequently shared between users, and those users may have different Just versions installed, which means in some cases some newer features might not be available in the older `just` versions.

We need a way to specify the minimum required version for `just`, similar to how `cmake` and many other similar tools do it.  The actual syntax is up for a discussion:

### Naming is hard

Some ideas...
```justfile
set mjv := '1.32.0'
set minimum-just := '1.32.0'
set require-just-at-least := '1.32.0'
set min-required-just := '1.32.0'
set min-just-version := '1.32.0'
```

hustcer:
set min-just-version := '1.32.0'

casey:
This is actually quite a tricky problem. Let's say there's a new function `foo()` in `just` 1.50.0, so when using it, you put `set minimum-just-version := '1.50.0'`:

```just
set minimum-just-version := '1.50.0'

foo:
  echo {{ foo() }}
```

However, a call to a new function in `just` 1.49.0 is a compilation error, so when you're producing an error message, you do not have a `Justfile` whose settings you can inspect for a `minimum-just-version` to produce a better error message.

This is a somewhat trivial example, because a missing function error could be deferred until the end of compilation, so you *did* have a `Justfile` you could get the minimum just version from, however in the case of, for example, new grammatical constructs, there's no way you can parse the justfile at all, so you can't extract the minimum just version to improve the error message.

There are a few solutions, but none of them seem very nice:

- Have a separatefile, like `.justversion`, which is a file which only contains the minimum version, and refer to that when an error happens.
- Try to grep through the file if you get an error, and just search for a line that matches `set min-just-version := '.*'`
- Have a special form for directives embedded in comments, so they don't depend on parsing anything else `#! just-version 1.50.0`

I'm not convinced though that any of these are worth it though.

Spatenheinz:
Just a thought,
Would it make sense to only allow the minimum version as the very first tokens of a Justfile, but still as optional?

casey:
> Just a thought, Would it make sense to only allow the minimum version as the very first tokens of a Justfile, but still as optional?

This is possible, although I'd like to avoid adding special case parsing logic just for this feature.

windl1n:
what if add version into filename? like
```
.a.b.c.justfile
a.b.c@justfile
```
which can be created by `just --init`
and in justfile, when add new function,
```
[...]

PR #2892, opened 2025-09-26
feat: implement `just_version()` function

nyurik:
implement #2616

Note that this also allows #2290 

```just
_required_just_ver := if semver_matches(just_version(), '>=1.43.0') == 'true' {
  'yes'
} else {
  error('just version 1.43.0 or greater is required, found ' + just_version())
}
```


nyurik:
@casey hi, are there any blockers or concerns with this? Thx!

laniakea64:
Sorry if I'm missing something, but so far the proposed use cases for this function revolve around detecting whether the running `just` version has the features the justfile requires.  But when a too-old `just` version is presented with a syntax or function that requires a later `just` version, it will fail with a compile error **before the justfile's version check is evaluated**.  So as casey said in https://github.com/casey/just/issues/2290#issuecomment-2266033579 , a `just` function cannot fulfill those use cases.

Are there any other use cases for `just_version()`?

nyurik:
ironically, my biggest use case is NOT to do version check, but rather to reliably report the version from inside.  All my `just`-using projects have this section. I need it to work in CI before anything else - just so that I can debug bigger issues.

```just
# Print environment info
env-info:
    @echo "Running for '{{main_crate}}' crate {{if ci_mode == '1' {'in CI mode'} else {'in dev mode'} }} on {{os()}} / {{arch()}}"
    {{just_executable()}} --version
    rustc --version
    cargo --version
    rustup --version
```

What I recently discovered is that this breaks on Windows - because `just_executable()` returns Windows path, but if the script runs using bash, it creates all sorts of problems.  So if we have {{just_version()}}, I will have more flexibility in the future.

Do note that with every function we will have the "before" and "after" -- so anyone who wants to use any new function will have to ensure that the version of the just executable will handle it. Not much we can do about it "in retrospect" other than to create some external wrappers.

examples/kitchen-sink.just
```diff
@@ -5,6 +5,12 @@ set positional-arguments
 set dotenv-load
 set export
 
+_required_just_ver := if semver_matches(just_version(), '>=1.43.0') == 'true' {
```
casey, line 8: We should remove this, since this is a fragile way to enforce a particular just version, since the justfile may fail to compile at all with the older version.

src/function.rs
```diff
@@ -392,6 +393,10 @@ fn just_pid(_context: Context) -> FunctionResult {
   Ok(std::process::id().to_string())
 }
 
+fn just_version(_context: Context) -> FunctionResult {
+  Ok(env!("CARGO_PKG_VERSION").to_string())
```
casey, line 397:
```suggestion
  Ok(env!("CARGO_PKG_VERSION").into())
```

tests/functions.rs
```diff
@@ -6,26 +6,28 @@ fn test_os_arch_functions_in_interpolation() {
     .justfile(
       r"
 foo:
-  echo {{arch()}} {{os()}} {{os_family()}} {{num_cpus()}}
+  echo {{arch()}} {{os()}} {{os_family()}} {{num_cpus()}} {{just_version()}}
```
casey, line 9: Let's add a single extra test for this, instead of making these tests more complicated.

README.md
```diff
@@ -1794,12 +1794,13 @@ All functions ending in `_directory` can be abbreviated to `_dir`. So
...
   `"powerpc64"`, `"s390x"`, `"sparc"`, `"wasm32"`, `"x86"`, `"x86_64"`, and
   `"xcore"`.
-- `num_cpus()`<sup>1.15.0</sup> - Number of logical CPUs.
+- `num_cpus()`<sup>1.15.0</sup> — Number of logical CPUs.
 - `os()` — Operating system. Possible values are: `"android"`, `"bitrig"`,
   `"dragonfly"`, `"emscripten"`, `"freebsd"`, `"haiku"`, `"ios"`, `"linux"`,
   `"macos"`, `"netbsd"`, `"openbsd"`, `"solaris"`, and `"windows"`.
 - `os_family()` — Operating system family; possible values are: `"unix"` and
   `"windows"`.
+- `just_version()`<sup>master</sup> — The version of `just` being run, e.g. `"1.42.5"`.
```
casey, line 1803: I would leave off the `e.g`.

```suggestion
- `just_version()`<sup>master</sup> — The version of `just` being run.
```

casey (changes requested):
See comments!

casey:
Added in #3497.

PR #2904, opened 2025-09-30
Implemented wildcard import for justfiles

### LINKED ISSUE #1885: support wildcard imports in justfiles

```
import '.just/git.just'
import '.just/poetry.just'
import '.just/precommit.just'
...
...


```
import '.just/*.just'

...

BoolPurist:
## Issue  

Closes #1885

## Needed Refactoring

I needed to change the type of the `absolute` field from `Option<PathBuf>` to Vec<PathBuf> in the Variant `Import` of the Enum `Item` in file `src/item.rs`.  Otherwise  the Ast would always try to match the globing path  something like `*.just` from a parsed Import Item against absolute paths of the loaded/imported files.

Note: Just will always  try to check if this path given to import, is an existing file before any gobbing matching is done. 

## Glob added as dependency

I  included the dependency [glob](https://docs.rs/glob/latest/glob/) to handle the globbing matching the syntax of something like import `*.justfile`. This crate also handles Window Paths. 

If this dependency is not wanted please let me know for another solution for having the globbing  pattern matching.

## My checks

Since this is my 1. code  contribution to this codebase

I tested with the following commands
```
cargo t --all
just ci
```

I hope this is alright  for you.

## Backwards compatability 
This feature breaks backwards compatability.
 
However this should be fine, according to a [comment](https://github.com/casey/just/issues/1885#issuecomment-1925468685) from casey on the isuue in question.


BoolPurist:
I have finally noticed that there is already a related [PR](https://github.com/casey/just/pull/2376/files) going on much longer. Unfortunately I did not notice this previous PR because my PR and the previous one are referencing different Issues.
Mine pointed to this [one](https://github.com/casey/just/issues/1885). The previous PR pointing to this [issue](https://github.com/casey/just/issues/1641).

I close my duplicate PR in favor of other [PR](https://github.com/casey/just/pull/2376/files). Since the previous PR progressed much further in the review process and also incorporates an opt-in option so It does not break backward compatibility compared to mine.

PR #2906, opened 2025-10-03
Remove legacy hack forcing justfile to be handled as Makefiles

alerque:
Both VIM and Neovim ship with native ftdetect rules to detect justfiles
and proper syntax highlighting for them. This modline hack is absolete
and actually makes these files *less* accessible in modern editors.


laniakea64:
Nice catch.

Previously, casey has preferred to outright remove outdated suggestions about `just` support in editors.  So instead of warning that using `make` modeline is bad for current Vim/Neovim, maybe best entirely remove the suggestion about `make` modeline?

alerque:
I would support that. In fact I started nuking that section, then restored it when I found some mentions of modern editor support above thinking that might be more controversial than the actual cleanup.

I just removed it entirely.

casey:
Better late than never ;) Thank you for the PR!

PR #2907, opened 2025-10-03
Add lazy variable evaluation with `lazy` keyword

### LINKED ISSUE #953: Make assignment evaluation lazy
I'd like to be able to define variables with backtick substitutions that are only run if the variable is referenced (I'd be fine if *all* the variables used in any commands are run before any of the commands run).

The use case I have is having a justfile that runs in multiple contexts, and some of the backtick substitutions can only be run in one of the contexts:

```make
nix_volume_path := `podman volume inspect --format '{{.Mountpoint}}' nix`

build:
    podman exec nix just /src/_build

_build
   nix build ...
```

where the `podman volume inspect ` does not succeed in the inner context.

MasterNayru:
I have been checking #953 for many years now waiting for this feature to land and getting all frustrated and bent out of shape, and after having to write another Makefile, I have started hate writing Rust to put some simple implementation forward rather than suffer in silence.

This PR implements lazy evaluation for variable assignments using a new `lazy` keyword. This is something that had been discussed as a potential way forward by someone, and as we all know if two people on GitHub give it a tick that is all the green lights we need. 

The main motivation for this is that over the years I have been unable to really suggest just as an alternative because I often need to get things like credentials to log into AWS services or some such, and having them execute on every invocation just gets way too expensive.

I am not sure what I need to update in terms of the documentation, grammar, etc. but I thought I would at least put this forward and let everyone throw tomatoes at it so at least I gave addressing the issue a try.

I have tried to write tests for this that, hopefully, cover all of the bases and edge cases, but my experience with the more exotic features of just is somewhere between zero and very limited.

```
Usage:
  lazy credentials := `aws sts get-session-token`

  # credentials are only evaluated if this recipe runs
  deploy:
    ./deploy.sh {{credentials}}
```

Fixes #953

### DIFF
--- a/GRAMMAR.md
+++ b/GRAMMAR.md
@@ -59,7 +59,7 @@ alias         : 'alias' NAME ':=' target eol
 
 target        : NAME ('::' NAME)*
 
-assignment    : NAME ':=' expression eol
+assignment    : 'lazy'? NAME ':=' expression eol
 
 export        : 'export' assignment
 
--- a/README.md
+++ b/README.md
@@ -1491,6 +1491,19 @@ braces:
   echo 'I {{ "{{" }}LOVE}} curly braces!'
 ```
 
+#### Lazy Evaluation
+
+By default, variables are evaluated when they are defined. If you would like a
+variable to only be evaluated when it is used for the first time, you can use the
+`lazy` keyword:
+
+```just
+lazy password := `aws ecr get-login-password --region us-west-2`
+```
+
+This is useful for values that are expensive to compute, or that may not be
+needed in every invocation of `just`.
+
 ### Strings
 
 `'single'`, `"double"`, and `'''triple'''` quoted string literals are
--- a/README.md
+++ b/README.md
@@ -1498,11 +1498,27 @@ variable to only be evaluated when it is used for the first time, you can use th
 `lazy` keyword:
 
 ```just
-lazy password := `aws ecr get-login-password --region us-west-2`
+lazy aws_account_id := `aws sts get-caller-identity --query Account --output text`
+```
+
+Once a lazy variable has been evaluated, its value is the same for the rest of
+the invocation of `just`, even if it is used multiple times:
+
+```just
+lazy timestamp := `date +%s`
+
+foo:
+  # The value is computed here
+  echo The time is {{timestamp}}
+  sleep 1
+  # The same value is used here
+  echo The time is still {{timestamp}}
 ```
 
 This is useful for values that are expensive to compute, or that may not be
-needed in every invocation of `just`.
+needed in every invocation of `just`. It also saves you from having expensive
+values being recomputed even for simple invocations of `just` that don't
+actually use them, like `just --list`.
 
 ### Strings
 
--- a/src/assignment.rs
+++ b/src/assignment.rs
@@ -9,6 +9,10 @@ impl Display for Assignment<'_> {
       writeln!(f, "[private]")?;
     }
 
+    if self.lazy {
+      write!(f, "lazy ")?;
+    }
+
     if self.export {
       write!(f, "export ")?;
     }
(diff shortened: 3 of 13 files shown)


PHANTOMKATHERINE (approved):
A lazy approval from a lazy mate

BoolPurist:
Cool feature. 
Looking at your tests though. I do not find any asserting if a variable assignment is only done once. 

Maybe a test case where you produce a random number and see whether printing the variable remains the same ?

MasterNayru:
@BoolPurist That's an excellent idea, I hadn't considered that. I have added a test that generates a timestamp and ensures that it is the same value when referenced more than once. It does only evaluate the value once.

BoolPurist:
@MasterNayru Nice, your approach of the timestamp for testing if evaluated only once is even better/simpler compared to generating a random number.

BoolPurist:
> I am not sure what I need to update in terms of the documentation, grammar

In my opinion the usage of this feature might be best put in its own chapter aka under its own MD heading within the README.
Maybe after the section with MD Heading called "Expressions and Substitutions".

There is also a grammar file where this new keyword `lazy` should be inserted. My guess would be that it belongs into grammar rule `assignment` under MD Heading `grammar`,

Maybe replace 
```
assignment    : NAME ':=' expression eol
```
with this
```
assignment    : 'lazy'? NAME ':=' expression eol
```

MasterNayru:
I can certainly push a change up that does a best-guess at this. I was just reluctant to push a bunch of stuff that was wrong that some poor maintainer then needs to correct. This is my first contribution to a project of this size, and I don't have much experience in Rust as I learnt just enough to write this change. I was hoping that the change would be small enough that getting this over the line wouldn't involve me playing guessing games as to what should go where.

If I see someone munge another Makefile of mine with macros and function calls, my eyeballs will roll out of my head and run away, so I just want to do whatever it takes to get the feature in as this feature not being there has meant that I couldn't introduce it at the last 3 companies that I've worked at.

BoolPurist:
> I can certainly push a change up that does a best-guess at this. I was just reluctant to push a bunch of stuff that was wrong that some poor maintainer then needs to correct. 

Fair enough. To me, you did good work so far. I appreciate that you followed my suggestion.

About my suggestion for adding the documentation. 
It is my belief that your implementation of this suggestion of adding this documentation at these places is helpful for the maintainer even if the maintainer needs to correct things.  If these changes now cause more trouble than It is worth then I am in the wrong, sorry about that and need to rethink what is useful for the maintainer in the future. 

I hope this feature will be merged. Here is my informal approval of this PR.

casey:
Sorry for letting this sit! Check out my comment here and let me know what you think https://github.com/casey/just/issues/953#issuecomment-3855815862. I'm interested in the trade-offs between the two approaches.

Also, I was testing this branch, and I found an issue, this will print two different numbers for `just a b`:

```just
lazy foo := `echo $RANDOM`

a:
  echo {{ foo }}

b:
  echo {{ foo }}
```

I think lazily evaluated assignments, when evaluated, are getting cached in an inner scope, not in the outer scope in which they were declared.

MasterNayru:
@casey I pushed a commit that I believe addresses the issue regarding variable scope. I don't know whether it's the most efficient way of handling a problem like this but I made sure to also include a test that has a Justfile that includes two other Justfiles imported with `mod` to make sure that use case was covered as well.

In terms of performance impact, the consequences of my decisions on the maintenance of the project, etc. I haven't given that much thought mostly because I'm not comfortable enough with Rust to spot a footgun like that. When I saw that the issue was marked initially as a "good first issue", I was quite surprised because the mechanics of a change like this would be quite profound.

Having said that, my only concern at this stage would be around how easily someone could get an existing Justfile, turn a global setting on and go through the process of finding what things were depending only eager assignment evaluation. You mentioned it in your PR, but I do imagine you would want to have an eager keyword at some point. Whether you see the need to do that now, or whether you let that need organically surface is something you are much better placed to answer.

At the end of day, I just think lazy evaluation is a vital feature for a project like this and I really want the project to be able to do this. I don't mind whether it's this PR that gets it over the line or not, I just want to be able to stop using Make.

casey:
I wound up going with the setting approach, implemented in #3083. It's currently unstable, see #3124 for more details.

PR #2917, opened 2025-10-12
Import remote just files via git

gmcquillan:
This is probably a bad idea because it introduces the concept of dependencies, which `just` is blissfully devoid of currently.

However, I really wanted teams to have the ability to link to a shared set of functions or tasks that could help keep our Just files in each repo no only shorter, but also more consistent between projects, without lots of copy and paste.

My thinking was the easiest way to do that is to have `just` clone repositories into a local cache, and handle versioning by using the sha1 in the import path of the git repository to check out specifically that version, but to otherwise default to the current HEAD of the repository.

Rust is not a primary language for me, so feel free to ask for changes or close this PR. 

I just wondered if something in this area was interesting to anybody else.


BoolPurist:
Fetching recipes from remote resource over the internet was already discussed at [this post](https://github.com/casey/just/issues/1799).

In short right now the community does not want to handle network related stuff. This git functionality is however network stuff.

Sharing just file through a remote Git repository can be also be archived by writing a just recipe invoking git commands to clone/get the recipe into the local project directory of a user.

Shared repository could be managed as git sub-modules in one's project.

woutervh:
My plan is to solve such use-cases with wildcard support.
see https://github.com/casey/just/issues/1885

casey:
I think this is a desirable feature, but I'm just not sure it's worth the complexity. There are some good workarounds, for example, you could have an init script which downloads submodule justfiles:

```bash
#!/usr/bin/env bash
curl -O https://foo.com/foo.just
```

Or similar for git.

PR #2921, opened 2025-10-16
Update README.md with uv tool install command

Wesztman:
I saw that you had a prior PR coming in which added `uv` to the readme, which you rejected in June.

I would argue that going into October 2025, `uv` is one of the most prominent and used package manager for Python packages out there.

I believe there are equally many or even more ppl installing tools using `uv tool install` compared to using `pipx` at this point.

With that said I believe this is a must to provide in the readme 😊


Wesztman:
@casey I'm closing this since another PR got merged that included this :)

PR #2925, opened 2025-10-21
add macos install option through homebrew to README

HalisCz:


casey:
Thanks for the PR! Homebrew is already listed in the cross-platform section.

Issue #2002, opened 2024-04-12
Omitting "else" from an if statement


alshdavid:
Hi all,

Please close this issue if this is already possible - it doesn't seem to work currently

When using an `if` statement, an else block must be present

```c
foo myarg:
  echo {{ if myarg != "" { "myarg is present" } else { "" } }}
```

Would it be possible to default to `""` is an `else` block is omitted?

```c
foo myarg:
  echo {{ if myarg != "" { "myarg is present" } }}
```

Thanks!

casey:
I think this is probably reasonable. I don't really like defaulting to `""`, it just seems inelegant, but I don't see a better alternative.

chrass:
I'm taking a look at this issue because it looks relatively simple to implement, where exactly is the implementation for conditionals?

casey:
It's in a few places, but you can start with `Parser::parse_conditional`, and make `Expression::Conditional::otherwise` an `Option`.

casey:
Implemented in #3401, now that we have a proper false value, the empty list `[]`, `if` without `else` evaluates to `[]` if its condition is false.

PR #2933, opened 2025-10-27
Make "else" clause optional in if/else statements (#2002)

nfogh:
Essentially does what you suggested in https://github.com/casey/just/issues/2002

### DIFF
--- a/README.md
+++ b/README.md
@@ -2413,6 +2413,21 @@ $ just bar
 abc
 ```
 
+If the `else` part is omitted, it defaults to "":
+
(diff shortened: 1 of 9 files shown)


casey:
I'm hesitant to merge at the moment. The empty string is the current preferred "false" value, but long-term, it's not what we want. Long term, I want to make all variables in just lists of strings, and use the empty list as the false value. This may wind up being a breaking change for places which can produce false values, so I'm hesitant to add more places which can produce false values.

PR #2936, opened 2025-11-01
Update README.md

### LINKED ISSUE #2935: Expected behavior of `env('FOO') || 'DEFAULT_VALUE'`
The documentation at https://github.com/casey/just?tab=readme-ov-file#environment-variables (and https://just.systems/man/en/functions.html#environment-variables) provides the following code snippet.
```
set unstable

foo := env('FOO') || 'DEFAULT_VALUE'
```

However, running that code verbatim with just 1.43.0 results in an error.
```
error: Call to function `env` failed: environment variable `FOO` not present
 ——▶ justfile:3:8
  │
3 │ foo := env('FOO') || 'DEFAULT_VALUE'
  │        ^^^
```

Modifying it slightly seems to make it work as intended (or at least causes it not to error out).
```
set unstable

foo := env('FOO', '') || 'DEFAULT_VALUE'
```

Could this possibly be an error in the documentation / implementation, or did I misunderstand how this should work?

laniakea64:
Fix https://github.com/casey/just/issues/2935

### DIFF
--- a/README.md
+++ b/README.md
@@ -1721,7 +1721,7 @@ A default can be substituted for an empty environment variable value with the
 ```just
 set unstable
 
-foo := env('FOO') || 'DEFAULT_VALUE'
+foo := env('FOO', '') || 'DEFAULT_VALUE'
 ```
 
 #### Executables


casey:
Merged, thank you!

Issue #2260, opened 2024-07-18
Stabilize `[script(…)]` attribute


casey:
A `[script(…)]` attribute was added in #2259. See #1479 for discussion.

The attribute is currently unstable.

Before stabilization:
- Make sure it's tested and has the right semantics.
- Is `script` the right name?
- Consider renaming "shebang recipes" in readme to "script recipes" and consider both shebang recipes and recipes with the script attribute as cases of script recipes.
- Add a `script-interpreter` setting and allow an empty `[script]` attribute.

linux-china:
Is it possible to add ext name support for `script` annotation: 

```
[script("pluto")]
[extension: 'lua']
foo:
  print(os.getenv("NICK"))
```

And following code is clean. 
```
[script("pluto", "lua")]
foo:
  print(os.getenv("NICK"))
```

If you use JavaScript/Types, lots of extension names:  js, cjs, mjs, ts, mts. 

```
[script("bun", "ts")]
foo:
  let name: string = "jackie"
  console.log(name)
```

casey:
I think that would probably be confusing, and using an additional attribute for the extension isn't much of a burden. But note that an extension is only required if the interpreter requires one. Most don't, in fact the only interpreters I know if that will fail without the correct extension is `cmd.exe` and PowerShell.

casey:
Also, one more reason for not stabilizing the `[script]` annotation right away: Since this is a new attribute, while it is unstable it is possible in a backwards compatible way to change the semantics of recipes with that attribute. In #1413 I discussed some shortcomings with shebang recipes, and discussed a lot of somewhat radical ideas for changing the semantics of recipes with a `[script]` attribute. Ultimately I don't think any of those were great ideas, and I don't have any particular ideas for changing the semantics of recipes with the `[script]` attribute, but I think it's a good idea to let it stew for a bit in case anyone can come up with some.

g9wp:
I wonder if there's a way to conditionally includes settings or variables to recipt like the [`cfg_attr`](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute) in rust.

```just
  SH := "sh"

  [cfg_attr(unix, ```
     set shell := ["bash", "-c"]   // setting 
     SH := bash                    // variable
  ```)]
  [cfg_attr(windows, ```
     set shell := ["cmd", "/c"]
     SH := cmd
  ```)]
  [cfg_attr(cygwin, ```
[...]

PR #2944, opened 2025-11-09
Add windows-script-interpreter setting

cspotcode:
Discussed here: https://github.com/casey/just/issues/2260#issuecomment-2741516734

I did my best to copy coding conventions from `windows-shell`, but I'm not a rust expert.

Worth noting that, although this does solve my use-case -- pointing to a different `sh` on Windows than Linux -- it doesn't let you `[script('linux-shell')]` and `[windows-script('windows-shell')]`. If you're specifying shell inline within the attribute, you can't specify a different option for Windows and Linux. For that, you're still better off using `#!{{SHEBANG}}` and setting that variable conditionally.

Tests are failing locally for me, but I suspect that's because my `bash.exe` is WSL2, but tests try to use it as a Windows bash shell. I'm opening this PR to get some initial feedback; don't want to waste time fixing my local testing issue if this PR can't be merged for other reasons.

### DIFF
--- a/src/analyzer.rs
+++ b/src/analyzer.rs
@@ -169,6 +169,10 @@ impl<'run, 'src> Analyzer<'run, 'src> {
       unstable_features.insert(UnstableFeature::ScriptInterpreterSetting);
     }
 
+    if settings.windows_script_interpreter.is_some() {
+      unstable_features.insert(UnstableFeature::WindowsScriptInterpreterSetting);
+    }
+
     let source = root.to_owned();
     let root = paths.get(root).unwrap();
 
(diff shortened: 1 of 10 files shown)


casey:
I haven't reviewed the code, but I think adding `windows-script-interpreter` is probably fine. I like allowing platforms attributes on all top-level items, as you proposed in #2955, much better, but I wouldn't want the perfect to be the enemy of the good.

casey:
I'm working on #2955, allowing attributes on all top level items, including settings, which I think is probably a better option than adding more windows-specific settings.

PR #2946, opened 2025-11-09
Document path_exists() as returning strings

cspotcode:
This follows the same convention as for the documentation of `is_dependency()`.

This avoids confusion where a reader might assume that `just` supports boolean values and is returning true or false booleans which are compatible with `if` conditionals. The string `false` is considered truthy in `if` conditionals, which can be a surprising gotcha.


casey:
LGTM! Thank you for your patience 🙏

Issue #2826, opened 2025-07-16
Bug: `just` does not use `PATHEXT` environment variable when looking for an executable file on Windows, leads to `program not found` errors


hrumhurum:
I created [`gnu-tk`](https://github.com/gapotchenko/gnu-tk) tool that provides seamless access to GNU toolkits on non-GNU operating systems. I provided more details about it in [this discussion](https://github.com/casey/just/issues/2766#issuecomment-3078209372).

When verifying `gnu-tk` with `just`, I found a bug in `just` that occurs when it searches for executable files using `PATH` environment variable.

Consider `justfile` with a custom shebang line:

```
run:
    #!gnu-tk -l bash
    echo Hello from cross-platform GNU scripting
```

If `gnu-tk` is installed as NuGet package (`dotnet tool install -g Gapotchenko.GnuTK`), everything works as expected when `just run` command is issued:

```
Hello from cross-platform GNU scripting
```

But if `gnu-tk` is installed as NPM package (`npm install -g @gapotchenko/gnu-tk`), `just run` results in the following error:

```
error: Recipe `run` with shebang `#!gnu-tk -l bash` execution error: program not found
```

Both NuGet and NPM package managers install `gnu-tk` tool and it is available in the command shell of the system:

```
> gnu-tk --version
GNU-TK 2025.2.5
```

but they do it differently.

**NuGet** installs `gnu-tk.exe` file at `C:\Users\Tester\.dotnet\tools` directory which is a part of `PATH` environment variable. `just` finds it there and everything works.

**NPM** installs `gnu-tk.cmd` file at `C:\Users\Tester\AppData\Roaming\npm` directory which is also enlisted in `PATH` environment variable. But the file created by NPM has `.cmd` extension as opposed to `.exe`, and this time `just` cannot find `gnu-tk` program. This leads to the `program not found` error.

This is a common issue that arises when developers attempt to replicate the semantics of `which`/`where` utility in their code. The right way to do it on Windows is to use `PATHEXT` environment variable that enumerates file extensions that should be treated as launchable programs. Here is an [example implementation of an executable file lookup in Windows](https://github.com/gapotchenko/Gapotchenko.FX/blob/4619057cf9f9a71e29f794b6546caee58ccfff03/Source/Modules/Catalog/Diagnostics/Gapotchenko.FX.Diagnostics.CommandShell/CommandShell.cs#L91).

## Environment

```
Version: just 1.42.1
OS: Windows 11 x64
```

[...]

PR #2950, opened 2025-11-11
Fix Windows PATH resolution to respect PATH order instead of System32…

### LINKED ISSUE #2926: `which()` does not use PATHEXT on Windows
_Subtly different from #2826_

Reproduced on just 1.43.0.

The built-in `which()` and `require()` commands fails to find executables on Windows because they do not follow Windows conventions for executable lookup: use PATH and PATHEXT.

For example:

```justfile
# Suppose I am writing a cross-platform justfile

set unstable

# require_git_path_fails_windows := require('git')
# Logs error:
#    error: Call to function `require` failed: could not find executable `git`

# This succeeds on Windows but fails on Linux
require_git_path_succeeds_windows := require('git.exe')

git_path_empty_string_on_windows := which('git')
git_path_found_on_windows := which('git.exe')

# Is forcing extra syntax to be cross-platform desirable?
# What value comes from forcing this extra syntax, as opposed to finding the desired path on all systems without it?
# On Windows, executables are invoked without explicit extension and have extension on the filesystem.
git_path_cross_platform := if os() == "windows" { which("git.exe") } else { which("git") }

@default:
    echo "require_git_path_succeeds_windows = {{require_git_path_succeeds_windows}}"
    echo "git_path_empty_string_on_windows = {{git_path_empty_string_on_windows}}"
    echo "git_path_found_on_windows = {{git_path_found_on_windows}}"
    echo "git_path_cross_platform = {{git_path_cross_platform}}"
```

Some people may feel that checking `$PATHEXT` is a windows-specific hack that `just` shouldn't implement. But I disagree.

Just as filesystem logic must use `\`s on Windows, and the $PATH parser must split on `;` instead of `:`, so too must implementations of `which` look at `$PATHEXT`. This is baked into the way `$PATH` works on Windows: it only makes sense paired with `$PATHEXT`.  You will never find a file named `git` on the Windows path. It will never exist. You will find `git.exe`.

### LINKED ISSUE #2947: `just` prioritizes System32 over `$Path`; is impossible to put `bash` on `$Path` if you have WSL2 installed
This is a doozy, but I've done my best to explain it concisely.

# Bug

`just` on Windows, by virtue of `std::process::Command`, does not respect `$Path` they way we want when resolving shell or interpreter executables. This means, when you try to use `bash` and you have a `bash.exe` on your `$Path` (perhaps it's Git Bash), and you coincidentally have WSL2 installed, `just` ignores your preferred `bash.exe` and instead calls `C:\Windows\System32\bash.exe`, which launches a shell inside the WSL2 guest virtual machine, *not* a bash shell on your host.

This breaks integration tests on Windows, because `test.rs` sets `--shell bash`.

It's also a bug for anyone who wants to use Git Bash or any other `bash.exe` as their `just` shell.

# Fix

My proposed fix is for `just` to resolve executables to absolute paths using ["which"](https://crates.io/crates/which) or ["pathsearch"](https://docs.rs/pathsearch/latest/pathsearch/) crates. It can use this implementation consistently across:
- `which()` function
- `set shell`
- `[script()]` / `set script-interpreter`

We already want `which()` functionality, so using it to resolve shells and interpreters to absolute paths, fixing a bug in the process, and fixing the integration tests on Windows: seems like a win all around.

# Reproduction

You can reproduce it quickly if you have WSL2 installed:

```
# Put C:\Program Files\git\bin at the top of $Path
# Open any shell to confirm this works:
$ bash -c 'if wslinfo --version >/dev/null 2>&1; then echo BAD we are in WSL2 ; else echo GOOD, we are in Git Bash! ; fi'
GOOD, we are in Git Bash!
```
Now create this justfile:
```
set shell := ["bash", "-c"]
@default:
    if wslinfo --version >/dev/null 2>&1; then echo BAD we are in WSL2 ; else echo GOOD, we are in Git Bash! ; fi
```
...and run it in the shell:
```
$ just
BAD we are in WSL2
```

I've created a more exhaustive reproduction here: https://github.com/cspotcode/repros2/tree/just-path-behavior-windows

# Q&A

### Why in the world does std::process::Command prioritize System32?

_EDIT previous link was outdated_

These [rust release notes](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0/#reduced-windows-command-search-path
) explain it clearly.

It checks in this order: child's `$PATH` if it's been explicitly set, then System32, then parent's `$PATH`
We want it to check `$PATH` *before* System32.  (note that `$PATH` always contains a System32 entry, too)

~~It makes [some (buggy, undocumented) attempts](https://github.com/rust-lang/rust/blob/ac968c466451cb9aafd9e8598ddb396ed0e6fe31/src/libstd/sys/windows/process.rs#L133-L148) to resolve the executable *only when a PATH env var has been explicitly set.* In `just`'s case, it doesn't resolve the executable, leaving it to `CreateProcess` which checks a bunch of places before `$Path`: the local directory, `System32/`, etc.~~

~~https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa~~

## But isn't default resolution the correct (albeit strange) behavior, since it's the rust standard library?

I argue: no, it's not what we want. The proof is that *other shells intentionally don't behave this way!* They resolve the target executable according to `$Path` *before* the System32 fallback, which matches user expectations. They do not give `C:\Windows\System32` special treatment.

## What's the problem with using WSL2 bash.exe as your shell?

It might work sometimes, break in others. For example, the C: drive is often mounted at a different location in WSL. **Ultimately, `just` should respect `$Path` the same way that other shells do both on Windows and Linux.**

## Related tickets

#2926
#2826
https://github.com/rust-lang/rust/issues/15149
https://github.com/rust-lang/rust/issues/37519

nooscraft:
Fix Windows PATH resolution to respect PATH order instead of System32 priority

This commit fixes a critical bug on Windows where `just` would prioritize
C:\Windows\System32 over entries in the PATH environment variable when
resolving executables. This caused issues when users wanted to use executables
like Git Bash's `bash.exe` but had WSL2 installed, as Windows' CreateProcess
would find System32\bash.exe first and launch WSL2 instead of the user's
preferred bash.

The fix implements proper PATH-based executable resolution that:
- Respects PATH order (first match wins, not System32 priority)
- Checks PATHEXT extensions on Windows (.exe, .bat, .cmd, etc.)
- Works consistently across shell, script-interpreter, and shebang resolution
- Provides clear error messages when executables are not found

Changes:

Core Implementation:
- Add `resolve_executable()` function in `src/which.rs` that:
  * Searches PATH entries in order
  * On Windows, checks PATHEXT extensions for each PATH entry
  * Handles absolute paths, relative paths, and simple command names
  * Returns proper error messages when executables aren't found
- Add `ExecutableNotFound` error variant with helpful error messages

Platform-Specific Updates:
- Update `src/platform/windows.rs`:
  * Use `resolve_executable()` for shebang interpreters without forward slashes
  * Maintains existing cygpath behavior for Unix-style paths (with '/')
- Update `src/executor.rs`:
  * Use `resolve_executable()` for command interpreters
  * Ensures script interpreters respect PATH order
- Update `src/settings.rs`:
  * Use `resolve_executable()` for shell command resolution
  * Pass working directory for proper relative path resolution
- Update `src/evaluator.rs`:
  * Pass working directory to `shell_command()` for proper resolution
- Update `src/justfile.rs`:
  * Pass working directory to `shell_command()` for proper resolution

Testing:
- Add comprehensive Windows-specific tests in `tests/windows_shell.rs`:
  * Test PATH order respect (not System32 priority)
  * Test PATHEXT extension checking
  * Test script-interpreter PATH resolution
  * Test shebang interpreter PATH resolution
- Add cross-platform tests in `tests/shell.rs`:
  * Test PATH order respect
  * Test executable not found error messages
  * Test absolute path handling

Addressing Q&A from issue #2947:

Q: Why does std::process::Command prioritize System32?
A: Rust's `std::process::Command` doesn't resolve executables itself when no
   explicit PATH is set. It delegates to Windows' CreateProcess API, which
   checks multiple locations before PATH: current directory, System32, etc.
   This is documented Windows behavior but doesn't match user expectations
   or how other shells (bash, PowerShell, cmd.exe) behave.

Q: Isn't CreateProcessA's resolution the correct behavior?
A: No. While it's built-in Windows behavior, it doesn't match user
   expectations. Other shells intentionally resolve executables according to
   PATH order, giving users control over which executables are used. `just`
   should behave consistently with these shells on both Windows and Unix.

Q: What's the problem with using WSL2 bash.exe as shell?
A: WSL2 bash runs in a Linux virtual machine with different filesystem
   semantics. Paths like C:\ are mounted differently, which can break
   integration tests and scripts expecting Windows path behavior. Users
   should be able to choose Git Bash or other Windows-native shells via PATH.

Related Issues:
- Fixes #2947 (Windows PATH resolution)
- Fixes #2926 (PATHEXT extensions)
- Related to #2826 (Windows shell issues)
- Related to rust-lang/rust#15149 and rust-lang/rust#37519

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -94,6 +94,10 @@ pub(crate) enum Error<'src> {
   ExcessInvocations {
     invocations: usize,
   },
+  ExecutableNotFound {
+    name: String,
+    suggestion: Option<Suggestion<'src>>,
+  },
   ExpectedSubmoduleButFoundRecipe {
     path: String,
   },
@@ -409,6 +413,12 @@ impl ColorDisplay for Error<'_> {
       ExcessInvocations { invocations } => {
         write!(f, "Expected 1 command-line recipe invocation but found {invocations}.")?;
       },
+      ExecutableNotFound { name, suggestion } => {
+        write!(f, "Could not find executable `{name}` in PATH")?;
+        if let Some(suggestion) = suggestion {
+          write!(f, "\n{suggestion}")?;
+        }
+      }
       ExpectedSubmoduleButFoundRecipe { path } => {
         write!(f, "Expected submodule at `{path}` but found recipe.")?;
       },
--- a/src/evaluator.rs
+++ b/src/evaluator.rs
@@ -267,11 +267,16 @@ impl<'src, 'run> Evaluator<'src, 'run> {
   }
 
   pub(crate) fn run_command(&self, command: &str, args: &[&str]) -> Result<String, OutputError> {
+    let working_dir = self.context.working_directory();
     let mut cmd = self
       .context
       .module
       .settings
-      .shell_command(self.context.config);
+      .shell_command(self.context.config, &working_dir)
+      .map_err(|error| OutputError::Io(io::Error::new(
+        io::ErrorKind::NotFound,
+        format!("{}", error.color_display(Color::never())),
+      )))?;
 
     cmd
       .arg(command)
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -15,7 +15,11 @@ impl Executor<'_> {
   ) -> RunResult<'src, Command> {
     match self {
       Self::Command(interpreter) => {
-        let mut command = Command::new(&interpreter.command.cooked);
+        // Resolve executable to absolute path using PATH
+        let working_dir = working_directory.unwrap_or_else(|| Path::new("."));
+        let resolved = which::resolve_executable(&interpreter.command.cooked, working_dir)?;
+
+        let mut command = Command::new(resolved);
 
         if let Some(working_directory) = working_directory {
           command.current_dir(working_directory);
--- a/src/justfile.rs
+++ b/src/justfile.rs
@@ -153,7 +153,7 @@ impl<'src> Justfile<'src> {
         binary, arguments, ..
       } => {
         let mut command = if config.shell_command {
-          let mut command = self.settings.shell_command(config);
+          let mut command = self.settings.shell_command(config, &search.working_directory)?;
           command.arg(binary);
           command
         } else {
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -27,8 +27,14 @@ impl PlatformInterface for Platform {
 
       Cow::Owned(cygpath.output_guard_stdout()?)
     } else {
-      // …otherwise use it as-is.
-      Cow::Borrowed(shebang.interpreter)
+      // …otherwise resolve using PATH to respect PATH order
+      let working_dir = working_directory.unwrap_or_else(|| Path::new("."));
+      let resolved = which::resolve_executable(shebang.interpreter, working_dir)
+        .map_err(|error| OutputError::Io(io::Error::new(
+          io::ErrorKind::NotFound,
+          format!("{}", error.color_display(Color::never())),
+        )))?;
+      Cow::Owned(resolved.to_string_lossy().to_string())
     };
 
     let mut cmd = Command::new(command.as_ref());
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -296,10 +296,11 @@ impl<'src, D> Recipe<'src, D> {
         continue;
       }
 
-      let mut cmd = context.module.settings.shell_command(config);
+      let working_dir = context.working_directory();
+      let mut cmd = context.module.settings.shell_command(config, &working_dir)?;
 
-      if let Some(working_directory) = self.working_directory(context) {
-        cmd.current_dir(working_directory);
+      if let Some(recipe_working_directory) = self.working_directory(context) {
+        cmd.current_dir(recipe_working_directory);
       }
 
       cmd.arg(command);
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -102,14 +102,21 @@ impl<'src> Settings<'src> {
     settings
   }
 
-  pub(crate) fn shell_command(&self, config: &Config) -> Command {
+  pub(crate) fn shell_command(
+    &self,
+    config: &Config,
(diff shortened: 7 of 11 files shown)


cspotcode:
@nooscraft Looks AI-generated, yeah?

nooscraft:
> @nooscraft Looks AI-generated, yeah?

WDYM? The commit message? Yes, it is, but the changes, nah, not so AI.

cspotcode:
Ok yeah, I read the commit message and ticket description, and they are *waaay* too verbose, outright copying and rephrasing sections from #2947. The implementation also has some AI hallmarks. I'm wary of reviewing in-detail, since babysitting someone else's AI output isn't a great use if my time.

Nothing wrong with using AI, I use it too, but I wonder if you should do another human review of its output first.

nooscraft:
> Ok yeah, I read the commit message and ticket description, and they are _waaay_ too verbose, outright copying and rephrasing sections from #2947. The implementation also has some AI hallmarks. I'm wary of reviewing in-detail, since babysitting someone else's AI output isn't a great use if my time.
> 
> Nothing wrong with using AI, I use it too, but I wonder if you should do another human review of its output first.

I don’t want to argue. The changes were properly reviewed before submitting the PR. The added tests support the modifications. The commit message was intentionally verbose and partially AI-assisted to highlight the key aspects. I could’ve easily rewritten it to sound more developer-like or less AI-generated, but I chose not to!

casey:
This is resolved in in #3160, which adds a new method `CommandExt::resolve`, which respects both PATH and PATHEXT on Windows, and uses it wherever commands are launched.

PR #2957, opened 2025-11-18
Add env attribute to set environment variables for recipes

neunenak:
The env attribute accepts two arguments (env_var, value) and can be used multiple times per recipe to set environment variables without using the existing `$parameter` syntax:

```just
[env('API_KEY', 'secret')]
[env('LOG_LEVEL', 'debug')]
deploy:
    ./deploy.sh
```

Addresses https://github.com/casey/just/issues/2825

### DIFF
--- a/README.md
+++ b/README.md
@@ -2139,6 +2139,7 @@ change their behavior.
 | `[confirm(PROMPT)]`<sup>1.23.0</sup> | recipe | Require confirmation prior to executing recipe with a custom prompt. |
 | `[default]`<sup>1.43.0</sup> | recipe | Use recipe as module's default recipe. |
 | `[doc(DOC)]`<sup>1.27.0</sup> | module, recipe | Set recipe or module's [documentation comment](#documentation-comments) to `DOC`. |
+| `[env(ENV_VAR, VALUE)]` <sup>1.44</sup> | recipe | Set env vars that apply only to this recipe. |
 | `[extension(EXT)]`<sup>1.32.0</sup> | recipe | Set shebang recipe script's file extension to `EXT`. `EXT` should include a period if one is desired. |
 | `[group(NAME)]`<sup>1.27.0</sup> | module, recipe | Put recipe or module in [group](#groups) `NAME`. |
 | `[linux]`<sup>1.8.0</sup> | recipe | Enable recipe on Linux. |
@@ -2514,6 +2515,17 @@ Parameters prefixed with a `$` will be exported as environment variables:
 
 ```just
 test $RUST_BACKTRACE="1":
+  # will print a stack trace if it crashes
+  cargo test
+```
+
+Or you can use the `[env(env_var, value)]` attribute to set environment variables that
+apply only to this recipe:
+
+```just
+
+[env("RUST_BACKTRACE", "1")]
+test:
   # will print a stack trace if it crashes
   cargo test
 ```
--- a/README.md
+++ b/README.md
@@ -2139,7 +2139,7 @@ change their behavior.
 | `[confirm(PROMPT)]`<sup>1.23.0</sup> | recipe | Require confirmation prior to executing recipe with a custom prompt. |
 | `[default]`<sup>1.43.0</sup> | recipe | Use recipe as module's default recipe. |
 | `[doc(DOC)]`<sup>1.27.0</sup> | module, recipe | Set recipe or module's [documentation comment](#documentation-comments) to `DOC`. |
-| `[env(ENV_VAR, VALUE)]` <sup>1.44</sup> | recipe | Set env vars that apply only to this recipe. |
+| `[env(ENV_VAR, VALUE)]` <sup>master</sup> | recipe | Set env vars that apply only to this recipe. |
 | `[extension(EXT)]`<sup>1.32.0</sup> | recipe | Set shebang recipe script's file extension to `EXT`. `EXT` should include a period if one is desired. |
 | `[group(NAME)]`<sup>1.27.0</sup> | module, recipe | Put recipe or module in [group](#groups) `NAME`. |
 | `[linux]`<sup>1.8.0</sup> | recipe | Enable recipe on Linux. |
@@ -2519,7 +2519,7 @@ test $RUST_BACKTRACE="1":
   cargo test
 ```
 
-Or you can use the `[env(env_var, value)]` attribute to set environment variables that
+You can also use the `[env(NAME, VALUE)]` attribute to export environment variables to a specific recipe:
 apply only to this recipe:
 
 ```just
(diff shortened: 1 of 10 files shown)


neunenak:
Also as a general note, I initially generated this PR with LLM coding tools and then manually audited+tweaked some things. I dunno if `just` as a project does or should have some kind of stated policy on that, but I wanted to mention that for the sake of transparency. 

I'm personally pretty bullish on AI-assisted coding, I do it all the time myself, and I take the position that in general a human developer is responsible for the code they submit to a project, but the project shouldn't care exactly how they generated that code.

neunenak:
@casey it'd be great if you could take a look at this, I think this is a pretty simple change that yields a pretty useful feature

src/attribute.rs
```diff
@@ -179,6 +181,20 @@ impl<'src> Attribute<'src> {
       AttributeDiscriminant::Confirm => Self::Confirm(arguments.into_iter().next()),
       AttributeDiscriminant::Default => Self::Default,
       AttributeDiscriminant::Doc => Self::Doc(arguments.into_iter().next()),
+      AttributeDiscriminant::Env => {
+        let [key, value]: [StringLiteral; 2] =
+          arguments
+            .try_into()
+            .map_err(|arguments: Vec<StringLiteral>| {
```
casey: I think this should be impossible to hit, because of argument range check above, so you can just `.unwrap()`

src/attribute.rs
```diff
@@ -243,7 +249,7 @@ impl<'src> Attribute<'src> {
   pub(crate) fn repeatable(&self) -> bool {
     matches!(
       self,
-      Attribute::Arg { .. } | Attribute::Group(_) | Attribute::Metadata(_),
+      Attribute::Arg { .. } | Attribute::Env(_, _) | Attribute::Group(_) | Attribute::Metadata(_),
```
casey, line 252: These should be alphabetically sorted.

src/attribute.rs
```diff
@@ -179,6 +181,10 @@ impl<'src> Attribute<'src> {
       AttributeDiscriminant::Confirm => Self::Confirm(arguments.into_iter().next()),
       AttributeDiscriminant::Default => Self::Default,
       AttributeDiscriminant::Doc => Self::Doc(arguments.into_iter().next()),
+      AttributeDiscriminant::Env => {
+        let [key, value]: [StringLiteral; 2] = arguments.try_into().unwrap();
```
casey, line 185: Just a note, I made this change myself, we can unwrap here because we check the argument count above.

tests/attributes.rs
```diff
@@ -323,3 +323,80 @@ fn shell_expanded_strings_can_be_used_in_attributes() {
...
 }
+
+#[test]
+fn env_attribute_single() {
+  Test::new()
+    .justfile(
+      "
+        [env('MY_VAR', 'my_value')]
+        foo:
+          echo $MY_VAR
```
casey: You can make this a quiet line with `@`, since checking the stderr is not very useful, and make the tests noisier.

tests/attributes.rs
```diff
@@ -323,3 +323,80 @@ fn shell_expanded_strings_can_be_used_in_attributes() {
...
+
+#[test]
+fn env_attribute_multiple() {
+  Test::new()
+    .justfile(
+      "
+        [env('VAR1', 'value1')]
+        [env('VAR2', 'value 2')]
+        foo:
+          echo $VAR1 $VAR2
```
casey: This should be a quiet line, so we don't have to check stderr.

tests/attributes.rs
```diff
@@ -323,3 +323,80 @@ fn shell_expanded_strings_can_be_used_in_attributes() {
...
+      ",
+    )
+    .stdout("value1 value 2\n")
+    .stderr("echo $VAR1 $VAR2\n")
+    .run();
+}
+
+#[test]
+fn env_attribute_1_arg() {
```
casey:
```suggestion
fn env_attribute_too_few_arguments() {
```

tests/attributes.rs
```diff
@@ -323,3 +323,80 @@ fn shell_expanded_strings_can_be_used_in_attributes() {
...
+  1 │ [env('MY_VAR')]
+    │  ^^^
+",
+    )
+    .status(EXIT_FAILURE)
+    .run();
+}
+
+#[test]
+fn env_attribute_3_args() {
```
casey:
```suggestion
fn env_attribute_too_many_arguments() {
```

README.md
```diff
@@ -2158,6 +2158,7 @@ change their behavior.
 | `[confirm(PROMPT)]`<sup>1.23.0</sup> | recipe | Require confirmation prior to executing recipe with a custom prompt. |
 | `[default]`<sup>1.43.0</sup> | recipe | Use recipe as module's default recipe. |
 | `[doc(DOC)]`<sup>1.27.0</sup> | module, recipe | Set recipe or module's [documentation comment](#documentation-comments) to `DOC`. |
+| `[env(ENV_VAR, VALUE)]` <sup>1.44</sup> | recipe | Set env vars that apply only to this recipe. |
```
casey: Since this will be initially unreleased it should be master:

```suggestion
| `[env(ENV_VAR, VALUE)]` <sup>master</sup> | recipe | Set env vars that apply only to this recipe. |
```

casey:
```suggestion
| `[env(NAME, VALUE)]` <sup>1.44</sup> | recipe | Export environment variable `NAME` with `VALUE` to this recipe. |
```

README.md
```diff
@@ -2537,6 +2538,17 @@ test $RUST_BACKTRACE="1":
   cargo test
 ```
 
+Or you can use the `[env(env_var, value)]` attribute to set environment variables that
```
```suggestion
You can also use the `[env(NAME, VALUE)]` attribute to export environment variables to a specific recipe:
```

casey (changes requested):
This looks good. Some minor tweaks.


A couple bigger questions:

Should exported environment variables should be available to recipe parameters?

```
[env("foo", "bar")]
baz x=`echo ${foo}.txt`:
```

Also, should they be available to `env` functions?

```
[env("foo", "bar")]
baz:
  echo {{ env("foo") }}.txt
```

I think the answer to both is probably yes.

As far as AI goes, I have no problem at all with LLM and AI generated code, as long as you the human coder "vouches" for the code to the same degree as code you wrote yourself, which requires reviewing the whole thing in depth before submitting it.

casey:
Actually, maybe they shouldn't be available to `env` functions. The `env` functions don't see exported top-level assignments or parameters, so the `env` attribute shouldn't be an exception. The question about parameters remains.

src/attribute.rs
neunenak: Am I missing something here, the Attribute members are in alphabetical order here

neunenak:
> Actually, maybe they shouldn't be available to `env` functions. The `env` functions don't see exported top-level assignments or parameters, so the `env` attribute shouldn't be an exception. The question about parameters remains.

This is actually potentially confusing behavior with `env()`, although altering this would be a breaking change. In any case, I implemented making environment variables set via `env` accessible in parameters but not via `env()`. I think it's fine to deter thinking about what the right way for `env()` to behave is to another issue.

neunenak:
@casey could you take a look at this?

casey (changes requested):
Looks good! Sorry for not thinking of this earlier, but I think that a duplicate variable name in an `[env]` attribute should be an error, since it's almost certainly a mistake.

casey:
Nice, merged!

PR #2958, opened 2025-11-18
Fix a typographical error in the rustbook docs

neunenak:
The `modules` anchor link was specified accidentally as `#modules1190` instead of `#modules`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -771,7 +771,7 @@ Available recipes:
     lint
 ```
 
-Recipes in [submodules](#modules1190) can be listed with `just --list PATH`,
+Recipes in [submodules](#modules) can be listed with `just --list PATH`,
 where `PATH` is a space- or `::`-separated module path:
 
 ```


laniakea64:
Not sure that's an error?

On the website, that link is here: https://just.systems/man/en/listing-available-recipes.html#listing-available-recipes

and https://just.systems/man/en/modules.html is 404

neunenak:
Ah my bad, this is what I get for trying to submit code at 3 am ._. the thing i was trying to fix was [1.6.38. Modules1.19.0](https://just.systems/man/en/modules1190.html) - so I changed the wrong string, but also it's werid that this html filename has the version string concatenated weirdly in it anyway. Will update this PR when I have a chance.

neunenak:
Ok I don't know why this didn't work either. Honestly the `Modules1.19.0` thing in the docs sidebar was just bugging me and though it would be easy to fix but I guess this involves debugging mdbook tooling.

casey:
I think having version references in headings (and thus anchor links) is just generally insane, so I removed them in #3153.

PR #2959, opened 2025-11-18
modify tests in positional.rs to use new style

neunenak:
Get rid of the code in positional.rs that defines a module-specific test macro, and replace it with the Test::new() infrastructure used elsewhere, for the sake of consistency with the rest of the codebase.

### DIFF
--- a/src/positional.rs
+++ b/src/positional.rs
@@ -96,133 +96,184 @@ mod tests {
 
   use pretty_assertions::assert_eq;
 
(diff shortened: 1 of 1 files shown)


neunenak:
Is this CI/pages thing maybe broken because of the cloudflare outage? Or is something else going on here @casey?

casey:
This seems like a good change, if you want to reopen the PR.

PR #2964, opened 2025-11-20
Update Readme with `brew install just` infos

JeanDavidDaviet:
From the readme it's not clear `just` can be installed with homebrew, but it can.


casey:
Thanks for the PR! It's a little confusing, but homebrew is cross-platform (it works on Linux too) so the instructions are in the cross-platform section.

PR #2966, opened 2025-11-21
Fix #2947

### LINKED ISSUE #2947: `just` prioritizes System32 over `$Path`; is impossible to put `bash` on `$Path` if you have WSL2 installed

cspotcode:
On Windows, explicitly set $PATH of spawned child processes so that executables resolve in $PATH order, and C:\Windows\System32 doesn't override it.

Fixes #2947

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,6 +93,11 @@ path = "src/main.rs"
 name = "just"
 test = false
 
+[[bin]]
+path = "src/bin/echoargs.rs"
+name = "echoargs"
+test = false
+
 # The public documentation is minimal and doesn't change between
 # platforms, so we only build them for linux on docs.rs to save
 # their build machines some cycles.
--- a/src/bin/echoargs.rs
+++ b/src/bin/echoargs.rs
@@ -1,0 +1,6 @@
+use std::env;
+
+fn main() {
+  let args: Vec<String> = env::args().skip(1).collect();
+  println!("{}", args.join(" "));
+}
--- a/src/command_ext.rs
+++ b/src/command_ext.rs
@@ -1,6 +1,8 @@
 use super::*;
 
 pub(crate) trait CommandExt {
+  fn with_explicit_path(&mut self) -> &mut Command;
+
   fn export(
     &mut self,
     settings: &Settings,
@@ -19,6 +21,18 @@ pub(crate) trait CommandExt {
 }
 
 impl CommandExt for Command {
+  fn with_explicit_path(&mut self) -> &mut Command {
+    // On Windows, set child's path explicitly so that spawned executable
+    // is located on path before System32 fallback, not the other way around.
+    // https://blog.rust-lang.org/2022/01/13/Rust-1.58.0/#reduced-windows-command-search-path
+    #[cfg(windows)]
+    if let Some(path) = std::env::var_os("Path") {
+      self.env("Path", path);
+    }
+
+    self
+  }
+
   fn export(
     &mut self,
     settings: &Settings,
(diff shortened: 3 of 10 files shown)


casey:
This is resolved in in #3160, which adds a new method `CommandExt::resolve`, which respects both PATH and PATHEXT on Windows, and uses it wherever commands are launched.

PR #2968, opened 2025-11-22
Improve bash completion with alias extraction and submodule support

tomers:
Enhanced the bash completion script with the following improvements:

- Extract both recipes and aliases from 'just --list' output instead of only using 'just --summary', providing more complete completions
- Add support for submodule recipe completion when typing 'just <module> <recipe>'
- Automatically detect module justfiles in common locations (, /justfile, modules/, modules//justfile)
- Parse aliases from both [alias: ...] and [aliases: ...] patterns in list output
- Add fallback to --summary if --list fails for better compatibility
- Improve recipe name extraction to handle edge cases better

These changes make tab completion more comprehensive and useful for projects using just modules and aliases.


Spatenheinz:
I tried this change out. Really some nice completion going on

tomers:
@casey can you please have a look?

casey:
Thanks for the PR!

The new completion engine is a complete overhaul, new features need to be added in Rust, which makes them work across shells, so this no longer applies. Please comment in #2406 if any of the improvements here still make sense with the new engine.

PR #2973, opened 2025-11-26
Fix `just --list` submodules example in readme

neodejack:
the example written forgot to pass in `--list`


casey:
Nice, good catch!

Issue #605, opened 2020-03-12
Consider named parameter overrides


steveej:
When dealing with multiple parameters with default options, it's currently necessary to set all parameters preceding the one which is intended to be overridden. This could be improved by allowing named parameter overrides, similar to how variable overriding works.

#### Example

Consider the recipe

```
rcp one="1" two="2" x="3":
    echo {{one}} {{two}} {{x}}

```

One has to override _x_ like so:

```
$ just rcp "1" "2" "4"
```

It would be nice to just do

```
# just rcp x="4"
```

#### Downsides
* It would become invalid to give a value beginning with `{{p}}=` to a parameter without giving its name first, where _p_ is any parameter name except the one given.

--- 

#476 might be an alternative to this.

casey:
Thanks for the well-written feature request!

I think the downside you mentioned is potentially an annoying one. I don't think it would come up often, but it would be pretty unexpected when it does.

Do you have a specific use-case for this feature?

steveej:
> Do you have a specific use-case for this feature?

The example demonstrates the use-case. The current workaround I'm using to do this is to move the parameters in question to global variables, because they can be overridden. I'd prefer to not need this workaround.

casey:
> The example demonstrates the use-case.

I'm curious about the specific use case, as in, what are you trying to do in the just recipe, to see if there's another way to accomplish what you're trying to do. Of course, it's entirely possible that there isn't!

I think the way that I would be inclined to support this is to add a token that can be used in the recipe signature arguments list, beyond which arguments can only be supplied values via overriding.

For example, using the `--` character:

```
foo a -- b="hello":
  echo {{a}} {{b}}
```

```
$ just foo 1 b=2
echo 1 2
```

This is a bit clunkier, and certain recipes can only be overridden, not positional, but it avoids the ambiguity of whether `X=Y` is an argument value or an override.

colearendt:
A bit late, but a use case that caused me to look for this same thing:
- just targets to build particular versions of different products (i.e. `just build-one version_one=1` or `just build-one 1`, `just build-two version_two=2` or `just build-two 2`)
- a helper target that builds both (i.e. `just build version_one=1 version_two=2`) with optional selecting one or the other, or using the respective defaults. Positional parameters here is hard to grok as a user.

It's possible to do this with global variables: `just version_one=9 build`, but that feels quite unnatural 🤷 Any other recommendations would be most welcome!

TheNamlessGuy:
Another example use-case:

[...]

PR #2974, opened 2025-11-26
Named parameters

Spatenheinz:
This PR is an attempt at solving an issue I and others have with wanting "python-like" keyword-arguments / named-parameters.
See e.g. 

https://github.com/casey/just/issues/605

I think it is an overall neat solution as it does not break any usage of normal recipes, its completely opt-in.


casey:
Thanks for the PR! I ultimately decided to go with a long and short option syntax, `--foo` and `-f`, implemented starting with #3026. My reasoning was that long and short options are more common on the command line, so users would be more familiar with them.

PR #2981, opened 2025-12-03
Add set no-cd and align path resolution

### LINKED ISSUE #2291: [Feature Request] An option to disable change directory for all recipes in justfile
Hello. `just` is a great tool and I really like it. However this automatic change directory behavior and using `[no-cd]` in all my recipes is a bit annoying (but still acceptable). 
Would you consider adding some setting to disable this behaviour? Something like `set no-cd` and new attribute `[do-cd]` or something like this? 

### My use case. 

All projects I'm working on are in `~/Projects`. 
```
~/Projects/my-project-1
~/Projects/my-project-2
...
```
I want to reuse one `justfile` located in `~/Projects/justfile`. All those projects contain terraform configuration that I'm working on. When I want to change infrastructure of some project I will:
1. Go into `~/Projects/my-project-2` 
2. Make changes in infra config
3. Run `validate`, `fmt`, `plan` 
4. If all is okay then `apply` infrastructure

There are some other common actions I do on all projects also but this is not important. 

To work with `terraform` I need to be inside the directory that contains configuration. To make it work with `just` each recipe needs to have this `[no-cd]` added.

My working `justfile` looks like similar to this:
```
set quiet := true

[private]
[positional-arguments]
run-tf *args='':
    # this is actually more complicated command using docker with volumes and whatnot
    # but for this example I simplified it
    terraform "$@"

[no-cd]
init:
    just run-tf init

[no-cd]
upgrade:
    just run-tf init -upgrade

[no-cd]
validate:
    just run-tf validate

[no-cd]
plan:
    just run-tf plan

[no-cd]
apply:
    just run-tf apply

[no-cd]
destroy:
    just run-tf destroy

[no-cd]
cli:
    just run-tf bash

[no-cd]
format:
    just run-tf fmt -recursive .

[no-cd]
output *args='':
    just run-tf output "$@"

alias console := cli
```

All is working fine but would be great if I could just use something like `set no-cd` and remove all those `[no-cd]` lines. If I wanted to change dir to jsutfile dir I could then add `[do-cd]` in recipe and it would work the standard way.

Would you consider adding setting like this? 

Thanks

### LINKED ISSUE #2754: Feature request: setting `[no-cd`] for entire submodule file
The `[no-cd]` attribute can be set on particular attributes, but not on entire submodules.

What I want to do is to set the [working directory](https://just.systems/man/en/working-directory.html) to the justfile's directory:

```just
set working-directory := justfile_directory()
```

But it looks like I can't call `justfile_directory()` like this.

It would be convenient if there had been a `set no-cd := true` setting available at the submodule source file.

### LINKED ISSUE #2822: [no-cd] attribute ignored by the absolute_path/canonicalize functions
Not sure if this is a bug, but the result surprised me:
With the just script (below, and) located in `~/bin/pathtest`, the `no-cd` attribute has no effect on the `absolute_path`/`canonicalize` functions.

``` just
#!/usr/bin/env -S just --justfile
# -*- just -*-

cd FILE="./":
    @echo -e "\$HOME:\t\t$HOME"
    @echo -e "\$PWD:\t\t$PWD"
    @echo -e "pwd:\t\t$(pwd)"
    @echo -e "absolute_path:\t{{ absolute_path(FILE)}}"
    @echo -e "canonicalize:\t{{ canonicalize(FILE)}}"

[no-cd]
no-cd FILE="./":
    @echo -e "\$HOME:\t\t$HOME"
    @echo -e "\$PWD:\t\t$PWD"
    @echo -e "pwd:\t\t$(pwd)"
    @echo -e "absolute_path:\t{{ absolute_path(FILE)}}"
    @echo -e "canonicalize:\t{{ canonicalize(FILE)}}"
```

## Output:

```
$ /usr/bin/env just --version
just 1.42.1

$ echo $HOME
/home/dwa

$ echo $PWD
/home/dwa

$ which pathtest
~/bin/pathtest

$ pathtest cd
$HOME:		/home/dwa
$PWD:		/home/dwa/bin
pwd:		/home/dwa/bin
absolute_path:	/home/dwa/bin
canonicalize:	/home/dwa/bin

$ pathtest no-cd
$HOME:		/home/dwa
$PWD:		/home/dwa
pwd:		/home/dwa
absolute_path:	/home/dwa/bin
canonicalize:	/home/dwa/bin
```

### LINKED ISSUE #2940: The  `[no-cd]` attribute doesn't respect the `working-directory` of dependents
Hello there! Thanks a lot for creating this project 😄

It would be incredibly helpful to allow recipes with an explicit `working-directory` to maintain that location for the duration of its lifecycle. I thought adding `[no-cd]` to dependencies would work like that but it does not. Here's an example:

```just
[working-directory: "docs"]
@foo: dependency && dependency
  python -c "import os;print('caller in:', os.getcwd())"

[no-cd]
@dependency:
  python -c "import os;print('dependency in:', os.getcwd())"
```

```
❯ just foo
dependency in: C:\Users\ofek\Desktop\code\msgspec
caller in: C:\Users\ofek\Desktop\code\msgspec\docs
dependency in: C:\Users\ofek\Desktop\code\msgspec
```

Would this require new functionality or is the current behavior unintentional?

hans-d:
## Summary

Feature: Add module-local set no-cd to skip automatic cd before recipes/backticks; mutually exclusive with set working-directory; recipe attributes still override.

Opt-in behavior: Add set no-cd-strict so when no-cd is on, backticks/shell/path helpers resolve relative to the invocation directory instead of the module dir. Default remains the old behavior for compatibility.

- fixes #2291
- fixes #2754 
- fixes #2940 (partially?)
- fixes #2822


casey (changes requested):
Thanks for the PR!

This mostly looks good, but I would remove `set no-cd-strict` from this PR and leave it for later, to make it simpler.

elijahfhopp:
Would love to see this make it in. @hans-d I can offer a patch to resolve the changes Casey needs on this PR if you'd like.

To clarify about removing `no-cd-strict`, we want to have it `on` or `off` (while removing the setting)? Just trying to understand the desired behaviour.

Thanks, all!

hans-d:
this update is bringing the pr up to match master, and to have all test run locally.

re the no-cd-strict, it's similar to the no-cd but applied more strictly for certain functions (see linked issues). Added as it's new/changed behaviour compared to the original of no-cd.

will spilt the no-cd-strict into a new pr.

src/execution_context.rs
```diff
@@ -41,15 +41,20 @@ impl<'src: 'run, 'run> ExecutionContext<'src, 'run> {
 
   pub(crate) fn working_directory(&self) -> PathBuf {
     let base = if self.module.is_submodule() {
-      &self.module.working_directory
+      self
```
casey, line 44: I'm not sure I understand why this function is changing.

casey (changes requested):
This seems mostly okay, but I'm not sure why ExecutionContext::working_directory is changing, see comment.

src/execution_context.rs
hans-d: Likely lingering from teh previous bigger change. Reverted it and spotted a few other things that needed tightening.

casey (approved):
LGTM!

PR #2984, opened 2025-12-05
Revise VS Code extension details in README.中文.md

A-moment096:
A small modification on `README.中文.md`, just found that the information on VS Code's available extensions in Chinese version is outdated, and the Chinese version is not coordinate with the English version. So I just translated it to update.

Maybe we should change the description of this part on the github pages as well, but I don't know how to. Also I'm not sure if it's okay to delete the command line extension installation instruction.


casey:
Nice, thank you!

PR #2985, opened 2025-12-05
Add f'{format}' strings

### LINKED ISSUE #11: Allow interpolation inside of backticks and strings

casey:
Fixes #11.


alerque:
It doesn't look like this made it into the docs ... that or the website manual needs rebuilding or something.

casey:
Thanks for the report! I'll check it out.

casey:
Fixed! The published docs should now be up-to-date.

PR #2992, opened 2025-12-06
Allow newlines in interpolations and `}` to abut interpolation `}}`

casey:


casey:
@laniakea64 I removed the restricting on `}` abutting `}}` in an interpolation. I'm not sure why it existed, it was super easy to fix given that we track whether there are any open delimiters in the lexer. I'm wondering if it's a relic of before we did this.

PR #2999, opened 2025-12-09
support using variables for `working-directory`

jiezhuzzz:
This pull request enables the use of variables, concatenation, and path operations in `working-directory` settings.

**Testing**

- Added new tests to verify that the `working-directory` setting supports variables, string concatenation, and path joining expressions.

Issue #2309, opened 2024-08-10
Allow const expressions in settings


casey:
Currently, arbitrary expressions are not allowed in settings. This is because the value of expressions can, in the general case, depend on settings. An example is the `env_var()` function, which depends on environment variables, which may be loaded by `.env` related settings, or `shell()` and backticks, which depend on the shell.

This has come up a bunch of times, for example in #2307, where it would be useful to use a `just` variable to determine which env var is loaded.

I think the solution is to distinguish between "const" expressions, which cannot depend on settings or other operations, and can be evaluated statically, and non-"const" expressions, which can, and allow "const" expressions in settings.

Variables could be explicitly annotated as const `const foo := …`, or it could just be implicit.

casey:
Implemented in #3037. Non-boolean settings may now take any expression as their value, as long as that expression does not contain backticks or function calls, directly or transitively by reference, since the value of settings, like working directory, can affect the execution of functions and backticks.

alerque:
I think this issue is a candidate for closure now that this landed in a release.

casey:
Thanks for the reminder!

PR #3003, opened 2025-12-10
Redo #2999

jiezhuzzz:
The last PR is broken by the pull bot. This new PR repeats the code and provide the variable functionality to `working-directory`.


casey:
I don't think this will work in general, since the value of the working directory setting influences the behavior of backticks and calls to `shell(…)` which can appear in expressions:

```
set working-directory := `pwd`
```

Or:

```
set working-directory := shell("pwd")
```

I think this needs a more principled design, something like #2309.

PR #3035, opened 2025-12-31
Fix a leak of two pipes into child processes

### LINKED ISSUE #3034: Just leaks two file descriptors into subprocesses on unix
Since #2488, just has created a pipe pair for use in signal handling on unix, but it does not set `O_CLOEXEC` on them so they get leaked into child processes.  This is usually harmless, but it interacts badly with things that care about inherited FDs (for example, things that use the systemd socket activation protocol with `systemfd`).  It's a very easy fix though, instead of using `nix::unistd::pipe()` to create the pipe pair, use `nix::unistd::pipe2(nix::fnctl::OFlag::O_CLOEXEC)` instead.

rjmac:
Introduced in #2488, which opens a pipe pair without setting the O_CLOEXEC flag on them.

Fixes #3034

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ unicode-width = "0.2.0"
 uuid = { version = "1.0.0", features = ["v4"] }
 
 [target.'cfg(unix)'.dependencies]
-nix = { version = "0.30.1", features = ["signal", "user"] }
+nix = { version = "0.30.1", features = ["signal", "user", "fs"] }
 
 [target.'cfg(windows)'.dependencies]
 ctrlc = { version = "3.1.1", features = ["termination"] }
--- a/src/error.rs
+++ b/src/error.rs
@@ -194,6 +194,10 @@ pub(crate) enum Error<'src> {
     source: ctrlc::Error,
   },
   #[cfg(unix)]
+  SignalHandlerPipeCloexec {
+    io_error: io::Error,
+  },
+  #[cfg(unix)]
   SignalHandlerPipeOpen {
     io_error: io::Error,
   },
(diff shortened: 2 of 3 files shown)


casey (changes requested):
Great find and thank you for the fix! How did you happen to find this?

It looks like `pipe2` isn't supported on macOS. I think the alternative is to do:

```
nix::fcntl::fcntl(fd, FcntlArg::F_SETFD(FdFlag::FD_CLOEXEC))?;
```

On both ends of the pipe after opening it.

rjmac:
> How did you happen to find this?

systemd's socket-activation protocol expects to pass file descriptors into the launched service starting from FD 3.  I was using systemfd and being very confused when the file descriptor I received in that slot wasn't a socket.

>  I think the alternative...

Oh how annoying.  That is indeed the alternative (or close to it; O_CLOEXEC is the only flag defined for F_SETFD but you're not supposed to assume that), but it's not thread-safe, which is why `pipe2` exists.  Fortunately, in `just`'s case, it doesn't look like this matters, since this code is called before anything "interesting" happens in execution.

casey:
Nice, looks good to me. I added a `fcntl` helper function so the error handling isn't duplicated. Thanks again for finding this, if I got a bug report about this, without it having already been root caused, I would surely have never figured it out 😅

casey:
When bugs are found, I usually cut a new release immediately, but I have a few big new features in master that I want to wait on releasing for a little bit, to get feedback, but I should be doing a new release in a few days. (Very exciting stuff, recipes will be able to define and take flags and options, and many kinds of expressions will be allowed in settings, instead of just strings.)

PR #3052, opened 2026-01-13
Add glob pattern support for imports

### LINKED ISSUE #1885: support wildcard imports in justfiles

woutervh:
Implements wildcard pattern matching in import statements, allowing users to import multiple files at once using patterns like:

  import '.just/*.justfile'

Key features:
- Support for * wildcard in import paths
- Works with optional imports (import?)
- Files imported in deterministic alphabetical order
- Circular import detection still works correctly
- Preserves non-Unicode file path support on all platforms

Implementation uses custom byte-level pattern matching to avoid UTF-8 conversion issues that prevented previous attempts (PR #2376) from being merged. Pattern matching works at the OsStr level on both Unix (using byte arrays) and Windows (using UTF-16).

This feature is currently unstable and requires 'set unstable' in the justfile to use.

Fixes #1885

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+Unreleased
+----------
+
+### Added
+- Add glob pattern support for imports ([#1885](https://github.com/casey/just/issues/1885))
+  - Support `*` wildcard in import paths (e.g., `import '.just/*.justfile'`)
+  - Works with optional imports (`import? '*.just'`)
+  - Files imported in deterministic alphabetical order
+  - Preserves support for non-Unicode file paths on all platforms
+  - **Unstable**: Requires `set unstable` to use
+
 [1.46.0](https://github.com/casey/just/releases/tag/1.46.0) - 2026-01-01
 ------------------------------------------------------------------------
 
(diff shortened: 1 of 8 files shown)


casey:
I haven't really decided what my policy is for large AI written features like this. In another PR for a smaller feature which the author mentioned was partially written by AI, I said that I didn't really mind.

In this case though, it's a huge feature, and it's not clear to me that reviewing it makes sense. It's not really contributing anything, since I could just run the agent myself, and come up with same code. Additionally, I would be able to make the agent match the style of the project, and because I was involved in the back and forth with the agent in writing the code, the review would be easier.

I'm going to close this, because I don't really feel like putting effort into reviewing AI-generated code, when in fact, generating the code myself with AI would actually be easier and more fun.

About this PR I'll note:

- This PR contains many comments, which are not in the style of the codebase.
- It synthesizes new `Item::Import`s, which is unnecessary, `absolute` could be made a `Vec<PathBuf>`.
- It does not support wildcards in paths.

PR #3057, opened 2026-01-20
Add instructions to readme for setting up completions for oh-my-zsh

GradedJestRisk:
## Problem
https://github.com/casey/just/issues/618

## Solution
Use the directives found in [this comment ](https://github.com/casey/just/issues/618#issuecomment-1584657008)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4366,6 +4366,10 @@ $ just --completions zsh > just.zsh
 
 Please refer to your shell's documentation for how to install them.
 
+#### Zsh
+
+##### Homebrew (MacOS)
+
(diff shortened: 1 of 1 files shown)


GradedJestRisk:
Hi, is anyone interested into it ? I can close the PR if nobody is.

casey:
Thank you for the PR! I think the right thing to do here is to add a `just` plugin to ohmyzsh, since it's generally extended with plugins. A PR was opened but it stalled out, not sure what happened: https://github.com/ohmyzsh/ohmyzsh/pull/9287

GradedJestRisk:
Thanks, I'll carry over to oh-my-zsh plugin

PR #3061, opened 2026-01-22
fix issue where modules dotenv files are not loaded

### LINKED ISSUE #2712: dotenv settings are not respected when using mod feature
Thanks for the great tool! I've been setting up a hierarchical justfile structure with the mod/import feature.

I found that the dotenv settings are not respected when using `mod` feature. E.g., when a mod justfile stated that it requires a dotenv, it's not actually loaded.

I found it out when some of my recipes need some environment variables from `.env` but its not loaded when being executed as a mod

# Test env
A script to setup the test environment:
```sh
#!/bin/sh

mkdir subdir
echo -e 'mod subdir\n\ndefault:\n  echo root' > justfile
echo -e 'set dotenv-required := true\n\ndefault:\n  echo subdir' > subdir/justfile
```

Folder/file structure:
```sh
$tree
.
├── justfile
└── subdir
    └── justfile

$ cat justfile
mod subdir

default:
  echo root

$ cat subdir/justfile
set dotenv-required := true

default:
  echo subdir
```

# Behaviour
Running from root-level, the subdir (which requires **dotenv**) did NOT produce any error
```sh
$  just default
echo root
root

$ just subdir::default
echo subdir
subdir
```
But running in subdir, the correct behaviour is shown
```sh
$ cd subdir/
$ just default
error: Dotenv file not found
```	

This behaviour is unexpected and made some of the recipe that NEEDS certainly environment variables fails silently

chadac:
* currently if a module specifies a `dotenv-path`, this setting is not respected if the parent specifies a dotenv path of its own
* this adds that capability
* When a parent and module dotenv are present, I set it up by default to apply the module dotenv on top of the parents -- but I can change this to use only the module dotenv if preferred.

Resolves https://github.com/casey/just/issues/2712

### DIFF
--- a/src/justfile.rs
+++ b/src/justfile.rs
@@ -76,20 +76,51 @@ impl<'src> Justfile<'src> {
 
   fn evaluate_scopes<'run>(
     &'run self,
-    arena: &'run Arena<Scope<'src, 'run>>,
     config: &'run Config,
-    dotenv: &'run BTreeMap<String, String>,
+    dotenv_arena: &'run Arena<BTreeMap<String, String>>,
+    parent_dotenv: &'run BTreeMap<String, String>,
     root: &'run Scope<'src, 'run>,
-    scopes: &mut BTreeMap<String, (&'run Justfile<'src>, &'run Scope<'src, 'run>)>,
+    scope_arena: &'run Arena<Scope<'src, 'run>>,
+    scopes: &mut BTreeMap<
+      String,
+      (
+        &'run Justfile<'src>,
+        &'run Scope<'src, 'run>,
+        &'run BTreeMap<String, String>,
+      ),
+    >,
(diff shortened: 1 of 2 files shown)


casey:
Thanks for the PR! I think this needs design work before implementation. If a submodule specifies a dotenv, the dotenv file should probably only apply to things in that submodule, and not to the parent module. So the ultimate implementation will probably involve making each modules dotenv file independent, and only apply to things in that module, and its submodules.

chadac:
Thanks! That actually makes a lot more sense, I was thinking the method of modifying evaluate_scopes also didn't really look too clean. Would it be all right if I PR'd up a change for the alternative implementation? I could reach out on Discord too if you'd like to discuss in more detail

PR #3062, opened 2026-01-23
docs: add uv install

npikall:
Adding instructions on how to install just from PyPI with uv


casey:
Thanks for the PR! I think this list is in alphabetical order, so `uv` should be after `Snap`.

npikall:
@casey Thanks, I did not realize. Should be fixed now.

casey:
Nice, merged!

PR #3067, opened 2026-01-28
Add a real project justfile as a CI startup example

initdc:
Hello, I built my project with just for the major platforms, I'd like to share my Justfile.


casey:
Thanks for the PR! I think this probably isn't a good fit for an example, since those are mostly showing off specific features, and I'm not sure that a large Justfile like this makes a good reference.

PR #2497, opened 2024-12-02
Add initial support for OpenBSD with attribute openbsd. Closes #2496

vtamara:
I tested this change on OpenBSD 7.6beta1 and on Linux Ubuntu 24.04 and it worked in both.

Issue #1682, opened 2023-10-02
No way to distinguish between macOS and other Unix with attributes


benjamineskola:
Currently it seems like it's not possible to define a recipe for everything _but_ macOS. For example, I can write a recipe that runs only on macOS:

```make
[macos]
blah:
    echo blah
```

But it appears that there's no way to write the inverse, a recipe that does not run on macOS, because `linux` is false on BSDs (correctly so), but `unix` is true on macOS (also correctly).

(A `bsd` attribute wouldn't really help, since arguably that should also be true on Darwin anyway.)

Perhaps being able to negate an attribute:

```make
[macos]
blah:
  echo this is macOS

[no-macos]
blah:
  echo this is {{os()}}
```

Alternatively, or additionally, maybe there should be attributes for every possible value of `os()`? But I don't know if that function even has a complete list.

Edit:

It's possible to work around this by invoking a new instance of just:

```make
default:
   just only-{{ os() }}

only-macos:
  echo do something macOS-specific

only-openbsd:
  echo do something bsd-specific
```

But this way can't take advantage of the attribute system, and I suspect there are other downsides as well to a child process (e.g., dependencies may behave differently).

runeimp:
What is the use case that should work on all UNIX variants except Linux and macOS in a recipe?

benjamineskola:
>What is the use case that should work on all UNIX variants except Linux and macOS in a recipe?

I don't have such a use case. I don't care about excluding Linux; I care about excluding macOS.

Currently, there's no way to write a recipe for "all Unixes except macOS". I can write a macOS recipe, and a Linux recipe; but there's no way to write one for Free/Net/OpenBSD (or, I presume, Solaris or any other platforms — I have no idea how many are theoretically supported).

So if I have platform-specific tasks, where things are done one way on macOS and a different way on other Unixes, I can't use the attribute system to implement that. What I think _should_ work is something like:

```make
default: platform

[macos]
platform:
    # macOS specific here

[not-macos]
platform:
    # other-platform-specific here
```

Or:

```
[linux,freebsd,openbsd]
platform:
    # listing all the platforms would also be fine
```

runeimp:
[...]

PR #3072, opened 2026-01-29
Add attribute support for other BSD variants

jakewilliami:
Support BSD variants FreeBSD, Dragonfly, and NetBSD.  These are the remaining main BSD variants listed [in the Rust documentation for `target_os`](https://doc.rust-lang.org/reference/conditional-compilation.html#target_os).

See also #2497, which added support for OpenBSD.

This may also help towards a complete solution for #1682.


casey:
Please open this PR on a branch other than master. I wanted to make a small edit, but can't do so if a PR is opened from the master branch.

jakewilliami:
Thanks @casey, I have created #3115 using the same commit but moved to a different branch.

Issue #1294, opened 2022-07-28
List doc output too dark on Windows


runeimp:
The beautiful blue doc comments in the `--list` output on macOS is a primary blue on Windows in both PowerShell and CMD. It's hard to read against the black background in CMD and next to invisible on the default blue background of PowerShell. Could have default to a softer/brighter blue for the doc comments on Windows? 😀

casey:
Definitely! Could you post a screenshot?

flw-cn:
This should be related to the color scheme theme of the terminal. I'm using iTerm2 and I don't think it's a "beautiful blue", I think it's very hard to read and can easily damage the eyes.

Here's the effect I see here, I'm using iTerm2, and the standard xterm theme:

<img width="1061" alt="image" src="https://user-images.githubusercontent.com/5546718/181525764-fad35195-e502-4180-80f8-69e77b66f22c.png">

flw-cn:
If possible, I would like to be able to somehow configure the colors used by just, like LSCOLOR does. Or at least replace blue(34) with bright blue(94).

runeimp:
I guess the beautiful blue is only in KiTTY on macOS
![macOS - KiTTY - Screen Shot 2022-07-28 at 10 17 21](https://user-images.githubusercontent.com/1418110/181600185-b11be6b5-fb28-4c12-be5b-4f1bf4000f76.png)

Here is Terminal on macOS
![macOS - Terminal - Screen Shot 2022-07-28 at 10 19 20](https://user-images.githubusercontent.com/1418110/181600337-cbef55d3-f1a0-46b9-a579-8f1a07999084.png)

Windows Command Prompt
![Windows - Command Prompt - Screen Shot 2022-07-28 at 10 17 59](https://user-images.githubusercontent.com/1418110/181600408-b4b2f257-d101-4cf7-a75b-2611cf6b3e21.png)

Windows PowerShell
![Windows - PowerShell - Screen Shot 2022-07-28 at 10 21 16](https://user-images.githubusercontent.com/1418110/181600440-72e4e552-ca41-4e5a-95b3-b47e3e503d29.png)

In all cases of macOS it is Big Sur and
in all cases of Windows it is Windows 10

casey:
Nice, great screenshots, they definitely make the problem clear! I won't have time to hack on this until Saturday, but if someone wants to take a crack at it before then, I'm on board with using bright blue everywhere we're using normal blue.

[...]

PR #3075, opened 2026-01-31
allow color config for --list 

zachvalenta:
# why

Too dark output from `just -l` was a blocker to me adopting just, saw that it was covered by https://github.com/casey/just/issues/1294 and figured I'd have a go.

# before

before

# after

after

# how to

Can config via:

* env var
```sh
JUST_LIST_RECIPE_COLOR='#a6e3a1'
JUST_LIST_DOC_COLOR='#f9e2af'
JUST_LIST_ALIAS_COLOR='#cba6f7'
JUST_LIST_GROUP_COLOR='#f38ba8'
```

* flags
```
just 
    --list-recipe-color '#a6e3a1' \
    --list-doc-color '#f9e2af' \
    --list-alias-color '#cba6f7' \
    --list-group-color '#f38ba8' \
    --list
```


casey:
Thanks for the PR!

See my comment [here](https://github.com/casey/just/issues/1294#issuecomment-4122646119) I think this is really an issue with the terminal, and since terminal colors already have a mechanism to configure them, it's not clear that we should add that functionality to Just.

PR #3082, opened 2026-02-03
Improve PowerShell recipe completions

abdalmoniem:
Extend `PowerShell` completions to show recipe names, arguments, descriptions, and aliases, bringing them in line with `zsh` completions.

also moved `use [System.Management.Automation]` inside the `Register-ArgumentCompleter` block so users can do:

```pwsh
just --completions powershell >> $PROFILE
```

without getting errors about use statements not being before any other statements.

image

### DIFF
--- a/completions/just.powershell
+++ b/completions/just.powershell
@@ -1,6 +1,3 @@
-using namespace System.Management.Automation
-using namespace System.Management.Automation.Language
-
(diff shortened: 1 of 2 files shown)


casey:

I think most of this works with the new completion engine, but please comment in #2406 if anything is missing.

PR #3085, opened 2026-02-11
Modify Gentoo installation instructions for 'just'

DarthGandalf:
The package was moved from guru to main gentoo tree a while ago


casey:
Nice, moving up in the world! Thank you for the PR!

PR #3087, opened 2026-02-14
fmt: preserve indentation

### LINKED ISSUE #1684: `just --fmt` should preserve tabs
```just
default:
	just --list
```

Run `just --color always --fmt --unstable --check` on this file, you'll get this diff:

```diff
$ just --color always --check --fmt --unstable
 default:
-	just --list
+    just --list
```

**Version**: 1.14.0, obtained via nix-unstable channel

GeertJohan:
fixes #1684

### DIFF
--- a/src/justfile.rs
+++ b/src/justfile.rs
@@ -844,7 +844,7 @@ alias f := foo
       r"alias f := foo
 
 foo:
-    echo a",
+  echo a",
     );
   }
 
@@ -859,7 +859,7 @@ foo:
       r"alias f := foo
 
 foo:
-    echo a",
+  echo a",
     );
   }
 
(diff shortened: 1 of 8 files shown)


casey:
I'm actually not sure this is the right behavior. This will preserve each recipe's indentation, but shouldn't `just --fmt` use a single indentation style for all recipes?

GeertJohan:
@casey that would make it more analogous to `go fmt`, yes. Which could be a nice thing to have. I think it would  make sense to describe the preferred indentation in the docs as it helps setting expectations. It would also mean that users who rely on IDE formatting features cannot freely choose their indentation (see https://github.com/jackTabsCode/zed-just/issues/35)

casey:
I think the right way to solve this is to have some way of configuring what kind of indentation `just --fmt` uses. However, I'm totally unsure what the right way to do this would be 😅 I just opened #3136 for discussion.

Issue #3092, opened 2026-02-18
Migrate legacy `.bat` in favor of `.cmd` for shebang recipes


egasato:
It seems that `cmd.exe` support (in Windows) is not _fully integrated_ (take with a grain of salt).

Microsoft tends to be conservative when it comes to breaking changes[1], hence the officially recommended extension for batch scripts is still `.bat`. However, since Windows NT, `cmd.exe` was introduced as a replacement for `COMMAND.COM`, and support was added for a "stricter" execution mode, which is enabled only when using the `.cmd` file extension.

I believe Just should not rely on an MS-DOS-era interpreter and migrate the existing behaviour to default to the `.cmd` extension. If legacy support is really required, a specific shebang or attribute could be added to allow its configuration.

<sub>[1] Ignoring all the recent "Microslop" AI-assisted Windows updates that cause wreackage constantly, the proliferation of laggy React-native/Electron-based programs for basic functionality like the start menu or contextual menu, and the early deprecation of Windows 10 to pressure the public to use a not-so-stable OS.</sub>

casey:
Is this a breaking change? Some scripts could be written to depend on the `.bat` behavior, and would break if changed to `.cmd`.

egasato:
It could certainly be considered a breaking change, but that's up to you to decide.

Initially, I didn't know this was the case; when you write a recipe you don't encode anywhere the path, filename nor extension to be used by `just` when it writes the recipe contents into disk. You could consider this is an implementation detail, that some users might have relied on without being part of an explicit API, in which case this change is a bugfix and not a breaking change.

However, **if you agree with the take that all behaviour is part of an API**, be it explicit or implicit, then this is no longer a bug but a feature, in which case **this is a breaking change**.

Personally, I don't think this change would break most existing recipes (nor cause a lot of trouble if it happens to break them), because of the following reasons:

- The only difference is that when running `.bat` scripts, the `ERRORLEVEL` variable gets update only when a command fails, whereas on `.cmd` scripts it's always updated.

  - Error handling would require saving the `ERRORLEVEL` into a temporal variable to prevent other actions (such as printing an error message) from overriding the value and making the recipe fail "silently".
  - This behaviour can be disabled with `SETLOCAL DisableExtensions`.

[...]

PR #3093, opened 2026-02-18
Migrate `.bat` to `.cmd` (#3092)

egasato:
As described in the issue #3092, this PR changes the extension used when persisting batch scripts to disk from `.bat` to `.cmd`. Due to the stricter requirements of the latter, it also enforces CRLF encoding.


casey:
See my comment [here](https://github.com/casey/just/issues/3092#issuecomment-3999736078).

Issue #3094, opened 2026-02-19
Support UTF-8 with BOM for Windows PowerShell


egasato:
It seems that `powershell.exe` support (in Windows) is not fully integrated.

When writing shebang recipes that use non-ASCII characters encoded in UTF-8, the output gets mangled (or more like rendered as gibberish). To support UTF-8 on Windows PowerShell, the output encoding (or codepage) has to be changed in every recipe or the file can be encoded in UTF-8 with BOM. The former is not considered ideal due to its verbose nature, and due to unintended side-effects (changing the codepage affects how the Console Window Host behaves, which means a push/pop strategy has to be used to prevent the encoding change from affecting the "caller" process).

Here's an example:

```just
set quiet := true

default:
    #!powershell.exe -NoProfile
    Write-Host '🇪🇸 España ♥️💛♥️ just'
```

Here is the execution result of `just default`:

```text
En C:\Users\egasato\AppData\Local\Temp\just-D1BjHk\default.ps1: 5 Carácter: 51
+ Write-Host 'ðŸ‡ªðŸ‡¸ EspaÃ±a â™¥ï¸ðŸ’›â™¥ï¸ just'
+                                                   ~
Falta la cadena en el terminador: '.
    + CategoryInfo          : ParserError: (:) [], ParseException
    + FullyQualifiedErrorId : TerminatorExpectedAtEndOfString

error: Recipe `shebang` failed with exit code 1
```

Here's the expected result:

```text
🇪🇸 España ♥️💛♥️ just
```

brandon-avantus:
I'm experiencing the same issue. Here is my workaround until a fix is in:
```just
set quiet := true

[script('powershell.exe', '-NoProfile', '-Command', '& {Invoke-Expression (Get-Content -Path $args[0] -Encoding UTF8 -Raw)}')]
default:
    Write-Host '🇪🇸 España ♥️💛♥️ just'
```

The command can be shortened a bit using aliases:
```just
[script('powershell.exe', '-NoProfile', '-Command', '& {iex (gc -Path $args[0] -Encoding UTF8 -Raw)}')]
default:
    ...
```

casey:
Fixed in #3347, which includes a byte-order mark in all PowerShell scripts.

PR #3095, opened 2026-02-19
Force UTF-8 with BOM for Windows PowerShell shebang recipes

egasato:
As explained in #3094, Windows PowerShell support is not full due to encoding issues.

The default encoding (codepage) used by `conhost.exe` (Console Window Host) may vary depending on the display language and region of the system (e.g. 437 for US English, 850 for Western European languages), but it's certainly never 65001 for UTF-8 (still an experimental feature). Moreover, even if UTF-8 was the default codepage, correct execution should not depend on system-wide settings that can potentially be changed.

Forcing the UTF-8 Byte Order Mark to be present when the shebang recipe is for Windows PowerShell fixes most (if not all) output encoding issues.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,6 +42,7 @@ pub(crate) use {
     expression::Expression,
     format_string_part::FormatStringPart,
     fragment::Fragment,
+    fs::File,
     function::Function,
     interpreter::Interpreter,
     invocation::Invocation,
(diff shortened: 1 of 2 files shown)


casey (changes requested):
This seems fine to me. However, I would like to deduplicate the check for powershell in recipe.rs with the one in `Executor::script_filename`. They're quite similar. There should be a function in executor which returns an enum `ExecutorKind`, which has `Cmd`, `Powershell`, and `Pwsh` variants.. Also, the preamble can be concatenated with the script and written in one call to `fs::write`.

casey:
Fixed in #3347, which includes a byte-order mark in all PowerShell scripts.

PR #3097, opened 2026-02-23
list: show fallback recipes when fallback is enabled

### LINKED ISSUE #1500: "just --list" should list fallback recipes
Now that fallback is stable, it seems that `just --list` should be aware and display fallback recipes, perhaps under a "fallback" section.

veeceey:
When `set fallback` is active, `just --list` now walks up the directory tree and also displays recipes from parent justfiles. Each parent's recipes appear under a clearly labeled heading:

```
Available recipes:
    child-recipe # A child recipe

# Fallback recipes from /path/to/parent
Available recipes:
    build         # Another one
    parent-recipe # A parent recipe
```

It keeps walking as long as each parent also has `set fallback` enabled, and respects the same search config constraints as the run fallback logic (only works with directory-based search, not `--justfile`).

Changes:
- `Search` now derives `Clone` so we can iterate parent directories
- `list()` takes `loader` and `search` parameters to compile parent justfiles
- Added integration test

Manual test results:
```
$ cd /tmp/just-test/sub && just --list
Available recipes:
    child-recipe # A child recipe

# Fallback recipes from /private/tmp/just-test
Available recipes:
    build         # Another one
    parent-recipe # A parent recipe
```

All existing fallback (14) and list (45) tests pass.

Fixes #1500

### DIFF
--- a/src/search.rs
+++ b/src/search.rs
@@ -4,7 +4,7 @@ const DEFAULT_JUSTFILE_NAME: &str = JUSTFILE_NAMES[0];
 pub(crate) const JUSTFILE_NAMES: [&str; 2] = ["justfile", ".justfile"];
 const PROJECT_ROOT_CHILDREN: &[&str] = &[".bzr", ".git", ".hg", ".svn", "_darcs"];
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub(crate) struct Search {
   pub(crate) justfile: PathBuf,
   pub(crate) working_directory: PathBuf,
--- a/src/subcommand.rs
+++ b/src/subcommand.rs
@@ -107,7 +107,7 @@ impl Subcommand {
       }
       Dump { format } => Self::dump(compilation, *format)?,
       Groups => Self::groups(config, justfile),
-      List { path } => Self::list(config, justfile, path)?,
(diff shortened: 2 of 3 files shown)


veeceey:
just a gentle nudge on this when you have a moment!

casey (changes requested):
Thanks for the PR!

The fallback logic is complex enough that I don't want to duplicate it. It should be shared between list and run.

Also, not all recipes in parent justfiles are runnable.In particular, if you have a recipe `foo` in the justfile in the current directory, and a recipe `foo` in a fallback justfile in the parent directory, the parent `foo` is not runnable, since `just foo` will run `foo` in the current directory, so it should be filtered.

cbrown1234:
I've had a stab at implementing the shared logic, covering casey's suggestions and the other features noted in https://github.com/casey/just/issues/1500

You can see the diff here: https://github.com/cbrown1234/just/pull/1

I don't seem to be able to raise PR's to the project, even though the contributing section of the README seems to indicate I should, maybe things have changed and that's not been updated yet?

casey:
@cbrown1234 Sorry about that! PRs are currently closed. See #3227. I just added a note to the top of the contributing section of the readme in #3289.

casey:

PR #3099, opened 2026-02-24
Add flag to measure recipe execution time

GeorgeLS:
As part of this PR, I have added a new simple feature for printing elapsed time of recipes.
The output format is really simple: `<materialized_recipe> (<elapsed_time>)`.

Here's an example of two recipes and the output for those:

Recipe 1:

```
test:
    ...
```

Output:
```
test (Duration: <time>)
```

Recipe 2:
```
test arg:
    ...
```

Invocation:

```
just test server
```

Output:

```
test server (Duration: <time>)
```

The duration is reported as soon as the recipe finishes.

Please note that if you have recursive recipes where you basically invoke `just` inside the recipe, the time reporting won't be propagated to the recursive invocation. If you want to measure the inner-invoked recipe, just pass `-t/--time` to the inner `just` invocation.


src/config.rs
```diff
@@ -419,6 +408,14 @@ impl Config {
           .value_parser(value_parser!(PathBuf))
           .help("Save temporary files to <TEMPDIR>."),
       )
+      .arg(
+        Arg::new(arg::TIME)
+          .action(ArgAction::SetTrue)
+          .short('t')
```
casey: I would remove the short flag. They're a scarce resource, so I try to allocate them in response to popular demand.

```suggestion
```

completions/just.bash
```diff
@@ -30,7 +30,7 @@ _just() {
...
+            opts="-E -n -g -f -q -t -u -v -d -c -e -l -s -h -V --alias-style --ceiling --check --chooser --clear-shell-args --color --command-color --cygpath --dotenv-filename --dotenv-path --dry-run --dump-format --explain --global-justfile --highlight --justfile --list-heading --list-prefix --list-submodules --no-aliases --no-deps --no-dotenv --no-highlight --one --quiet --allow-missing --set --shell --shell-arg --shell-command --tempdir --time --timestamp --timestamp-format --unsorted --unstable --verbose --working-directory --yes --changelog --choose --command --completions --dump --edit --evaluate --fmt --groups --init --list --man --request --show --summary --usage --variables --help --version [ARGUMENTS]..."
```
casey: The completion scripts are now dynamically generated by calling into the binary, so this diff will go away when your rebase.

src/recipe.rs
```diff
@@ -16,6 +18,31 @@ fn error_from_signal(recipe: &str, line_number: Option<usize>, exit_status: Exit
   }
 }
 
+fn print_elapsed_recipe_time(
```
casey: I would just inline this.

src/recipe.rs
```diff
@@ -16,6 +18,31 @@ fn error_from_signal(recipe: &str, line_number: Option<usize>, exit_status: Exit
...
+    config.color
+  }
+  .stderr();
+
+  let prefix = color.prefix();
+  let suffix = color.suffix();
+
+  eprint!("{prefix}{recipe_name}");
+
+  for arg in recipe_args {
```
casey: I'm not sure about including recipe args, they can be quite long.

casey (changes requested):
See comments! There will be conflicts when rebasing, but the changes to the completion files won't be needed.

src/recipe.rs
GeorgeLS: Hmm. I think it can be useful since different arguments can yield different timings (e.g having a _build_ recipe where you pass `debug` or `release`) and it is convenient to see how each case affects the timings.

I didn't think of that being a problem since I usually don't use that many parameters or very long arguments.

Either way, if we would like to keep the arguments, how would the idea of having a threshold (e.g line length or number of arguments) and then omitting the rest of the args?

For example, print something like this:

```
<recipe> <arg1> <arg2> ...
```

Finally, we could print everything based on verbosity level.

casey: Let's just include them for now, regardless of length. We can figure out something clever later, if it's actually a problem in practice.

casey: Actually, I'm on the fence about this. If we want to print arguments, then I think perhaps we should print them before running the recipe, in this print statement:

```rust
    if context.config.verbosity.loquacious() {
      eprintln!(
        "{prefix}===> Running recipe `{}`...{suffix}",
        self.recipe_path(),
      );
    }
```

Put another way, if the user wants to see arguments, than I can't think of a good argument for why they should be in the timing print statement, and not in the initial print statement.

So I think it's better to leave off arguments for now, and deal with that as a separate issue.

GeorgeLS: Ok, let me change that real quick. We are going to leave arguments completely off the equation then.

src/recipe.rs
```diff
@@ -1,4 +1,5 @@
 use super::*;
+use std::time::Instant;
```
casey: This should be included in the `use` block in `lib.rs`.

src/recipe.rs
```diff
@@ -221,11 +222,30 @@ impl<'src> Recipe<'src> {
...
+      } else {
+        context.config.color
+      }
+      .stderr();
+
+      let prefix = color.prefix();
+      let suffix = color.suffix();
+      let recipe_name = self.name.lexeme();
+
+      eprintln!("{prefix}{recipe_name} (Duration: {elapsed:.2?}){suffix}");
```
casey: Add a four-character marker, so it doesn't get lost in other output, also, the display implementation shouldn't be used:

```suggestion
      eprintln!("{prefix}---> {recipe_name} completed in {:.3}s{suffix}", elapsed.as_secs_f64());
```

Three digits of precision seems good since milliseconds is a common unit.

casey (changes requested):
See comments! Also, this should have a test.

src/recipe.rs
GeorgeLS: You mean that I shouldn't use the `Debug` implementation.
The reason I used the `Debug` implementation, is because it is more human friendly.
It automatically converts to the closest unit of time that has elapsed.

For example, the output is going to be one of:
- seconds
- milliseconds
- nanoseconds
- microseconds

GeorgeLS: I'm leaving this open for discussion for now. I have added tests as well.

casey: I agree that it would be more user-friendly to do that, but the debug implementation is unstable, meaning that there are no guarantees that it might not change in the future.

GeorgeLS: Yeah from that perspective I agree. I will update the code as you suggested.

casey:
Nice, thank you!

Issue #3088, opened 2026-02-16
"--fmt --check" fails on variables when import is used


begoon:
file `Justfile`:
```
ABC := "ABC"

default: abc

import 'a.just'
```

file `a.just`:
```
abc:
    echo {{ ABC }}
```

execution works:
```
just

echo ABC
ABC
```

`--check` for the main `Justfile` works:
```
just --fmt --check --unstable
```

however:
```
just --fmt --check --unstable --justfile a.just

error: Variable `ABC` not defined
 ——▶ a.just:2:13
  │
2 │     echo {{ ABC }}
  │   
```

casey:
Fixed in #3103, and I added a test for this specific case in #3110.

PR #3103, opened 2026-02-28
Avoid compilation for justfile formatting

terror:
`--fmt` only uses the root AST's `Display` impl, but previously went through full compilation, including import and module resolution. This meant `--fmt` would fail with `MissingImportFile` or `MissingModuleFile` errors for files it never needed.

This diff moves `Format` handling to before the `Self::compile()` call, similar to `Edit`, so it loads, lexes, and parses only the root file. The unstable check inspects the AST directly for set unstable rather than going through `Config::require_unstable`, which requires a fully compiled Justfile.

### DIFF
--- a/src/compilation.rs
+++ b/src/compilation.rs
@@ -5,15 +5,10 @@ pub(crate) struct Compilation<'src> {
   pub(crate) asts: HashMap<PathBuf, Ast<'src>>,
   pub(crate) justfile: Justfile<'src>,
   pub(crate) root: PathBuf,
-  pub(crate) srcs: HashMap<PathBuf, &'src str>,
 }
 
 impl<'src> Compilation<'src> {
   pub(crate) fn root_ast(&self) -> &Ast<'src> {
     self.asts.get(&self.root).unwrap()
-  }
-
-  pub(crate) fn root_src(&self) -> &'src str {
-    self.srcs.get(&self.root).unwrap()
   }
 }
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -241,41 +241,6 @@ mod tests {
   use {super::*, temptree::temptree};
 
   #[test]
-  fn include_justfile() {
-    let justfile_a = r#"
-# A comment at the top of the file
-import "./justfile_b"
-
-#some_recipe: recipe_b
-some_recipe:
-    echo "some recipe"
-"#;
-
-    let justfile_b = r#"import "./subdir/justfile_c"
-
-recipe_b: recipe_c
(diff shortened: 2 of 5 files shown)


terror:
Related: https://github.com/casey/just/issues/3088

casey (changes requested):
This is definitely a good feature, although I'm not sure I like the duplication that it introduces. Lexing and parsing was previously only done in `Compiler::compile`, but now it's done both there and in `Subcommand::format`. Can this be de-duplicated? Perhaps by creating a helper function on `Compiler` which does lexing and parsing for a `Source`, which can be called in both places?

casey:
Is this ready for another look? I ignore notifications from pushed changes to a branch, since they're often just noise.

terror:
> Is this ready for another look? I ignore notifications from pushed changes to a branch, since they're often just noise.

No worries, yeah this should be ready. I lifted out a `Compiler::parse` to avoid duplicating lexing/parsing. We _could_ take it a step further and also load in this function, but 🤷 

I also don't know if this should live on `Compiler`. Seems better fitted for `Parser`?

casey (changes requested):
I agree that putting it on Parser is better. Ideally, we could replace `Parser::parse` with something that would take a `Source`, but if that doesn't work `Parser::parse_source` that takes a `Source` is fine.

casey:
Why did the `include_justfile` test go away?

terror:
> Why did the `include_justfile` test go away?

`Compilation::root_src` is now unused outside of tests, but I guess it could be a non-panic test, or make `root_src` test only 🤔

casey:
What was it testing? Are there other tests testing the same thing, or did we lose coverage by deleting it?

terror:
> What was it testing? Are there other tests testing the same thing, or did we lose coverage by deleting it?

It was testing that `Compiler::compile` successfully handles a chain of transitive imports across multiple files. From what I can tell, there are [integration tests](https://github.com/casey/just/blob/master/tests/imports.rs) that cover the same thing, i.e. [`nested_import_paths_are_relative_to_containing_submodule`](https://github.com/casey/just/blob/master/tests/imports.rs#L287), [`recipes_in_nested_imports_run_in_parent_module`](https://github.com/casey/just/blob/master/tests/imports.rs#L298), and [`nested_multiply_imported_items_do_not_conflict`](https://github.com/casey/just/blob/master/tests/imports.rs#L386).

casey (approved):
Dope, LGTM!

Issue #3070, opened 2026-01-28
What are `?` parameters?


solidsnack:
This may be a typo. My LLM was giving me a song and dance about Just's `?` parameters and what I could use them for.

https://github.com/casey/just/blob/982e8f18cbe1a853e88001a38bbd982a8f65682e/README.md?plain=1#L2853

However, there's no section for them in the docs; and this is the only reference I can find to them.

casey:
Thanks for the issue! There is definitely no such thing as `?` parameters, it should read "`*` parameters". Fixed in #3150.

Issue #2967, opened 2025-11-21
Add a skill for coding agents


PaulRBerg:
I've been using Just with Claude Code and think a [Claude Skill](https://code.claude.com/docs/en/skills) for Just would be really useful.

## The idea

Similar to the [Playwright Skill](https://github.com/lackeyjb/playwright-skill), a Just skill would let Claude automatically discover and run recipes without me having to manually explain what's in my `justfile` every time.

## What it could do

- Understand modern Just syntax and how to find Just docs
- Parse justfiles to find available recipes
- Run recipes with the right arguments
- Suggest relevant recipes based on what I'm trying to do

## Why this would be great

Claude often makes mistakes about Just features, e.g. it doesn't know about `[script]` or the predefined variables like `CYAN`.

A Claude Skill would make the workflow way smoother - less context switching, better discoverability of project tasks.

I suppose there is significant overlap between Just users and Claude Code users.

PaulRBerg:
I've rolled my own Agent Skill for Just - sharing here in case anyone wants to use it.

> [!NOTE]
>
> The skill is opinionated. You might want to customize the skill with your own preferences.

https://github.com/PaulRBerg/agent-skills/tree/d3f5540ed2fc0fa07f802bd925e06b9387cbe90f/skills/cli-just

You can install it using `skills` CLI:

```bash
npx skills add PaulRBerg/agent-skills --skill cli-just --global
```

casey:
This is really interesting! I'm definitely open to adding a skill for Just, but I want to understand what the goal would be.

What is it that agents don't do by default that you want them to do? Are these things specific to you or your workflow, or more general?

Is the goal to just get agents to use just, create justfiles, and write recipes? Or is it to make them do so more effectively?

What changes, both positive and negative, have you noticed after adding the skill you wrote?

I have some concerns about a just skill, which:

- Do skills cause agents to over-index on doing the thing described in the skill? Would it cause an agent to use just more than you would actually want it to, like, for example, writing lots of recipes that were only used once? Or creating a justfile in a project where it's never going to be used again?

- The contents of the skill have enormous overlap with the existing documentation, and as-is would mean maintaining essentially two copies of the documentation. If the core issue is that the readme is too long to put into an agent's context window, then I think a better approach is possibly a cheatsheet (see #3108), and the skill just tells the agent to run `just --cheatsheet` to put the cheatsheet into its context.

[...]

PR #3105, opened 2026-03-04
Add llms.txt for AI agent integration

### LINKED ISSUE #2891: Provide `llms.txt` for better integration with AI agents
`llms.txt` is a standard for AI-readable software documentation

https://llmstxt.org/

Bortlesboat:
Closes #2891

## Summary

Adds an `llms.txt` file to the `www/` directory so it will be served at `just.systems/llms.txt`.

[llms.txt](https://llmstxt.org/) is a proposed standard that helps AI agents and LLMs understand a project by providing a structured markdown overview with links to relevant documentation sections. This file gives AI tools a quick map of just's documentation — from getting started and installation through justfile syntax, recipes, expressions, execution, and editor support.

The file follows the llms.txt spec: a title, short description blockquote, expanded summary, and categorized links to the existing just manual pages at just.systems/man/en/.

## Placement

The file is placed in `www/` alongside the existing site assets (`index.html`, `CNAME`, etc.) so it will be served at the site root as `just.systems/llms.txt`.

### DIFF
--- a/www/llms.txt
+++ b/www/llms.txt
@@ -1,0 +1,95 @@
+# just
+
+> just is a command runner for project-specific tasks. Commands, called recipes, are stored in a file called justfile with syntax inspired by make. just avoids make's complexity and idiosyncrasies and works on Linux, macOS, and Windows.
+
+just is not a build system. It is a handy way to save and run project-specific commands.
+Recipes can accept command-line arguments, be written in arbitrary languages (Python,
+NodeJS, etc.), and load environment variables from .env files. just resolves errors
+statically, supports shell completion, and can be invoked from any subdirectory.
+
+## Getting Started
+
+- [Introduction](https://just.systems/man/en/introduction.html): Overview of just and its key features
+- [Quick Start](https://just.systems/man/en/quick-start.html): Minimal guide to get up and running
+- [Examples](https://just.systems/man/en/examples.html): Example justfiles demonstrating common patterns
+- [Backwards Compatibility](https://just.systems/man/en/backwards-compatibility.html): Stability guarantees and versioning policy
+
+## Installation
+
+- [Prerequisites](https://just.systems/man/en/prerequisites.html): System requirements for running just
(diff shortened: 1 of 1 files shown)


casey:
This is something which I want to understand why we're adding it and what we're hoping to accomplish. See my comment [here](https://github.com/casey/just/issues/2891#issuecomment-3999668157).

Bortlesboat:
Fair point. I should have led with the "why" instead of just dropping a file. And yeah, AGENTS.md makes more sense as the name.

  The reason I think this is worth considering: AI agents are already being used with just, and they're getting things wrong. #2967 is someone reporting that Claude doesn't know about [script] or color constants like CYAN. #3070 is someone whose LLM straight up hallucinated how ? parameters work. The README is thorough but it's ~18K words — agents either miss things in all that text or make stuff up when they can't find what they need.

  An AGENTS.md wouldn't duplicate the README. It'd be more like a cheat sheet for the stuff agents actually get wrong — things like "use {{var}} interpolation not $var", which features are version-gated, gotchas around modules and Windows shell behavior. Focused on real failure modes, not a summary of the manual.

  For measuring it — honestly hard to put a precise number on, but the proxy would be fewer issues caused by bad AI-generated advice about just syntax. #2967 and #3070 are early examples of that. You could also do before/after testing with a set of justfile tasks and see if agent output improves.

  If you'd rather write it yourself or skip it entirely, no hard feelings. Just wanted to make the case since I've seen the problem
  firsthand.

casey:
Sorry, I totally misunderstood the difference between an `llms.txt` and an `AGENTS.md` file. I think that an `llms.txt` files is only useful if agents spontaneously look at it or request it, and otherwise, there's enough overlap with what humans need that it doesn't make sense to have a separate file. I.e., if the value of an `llms.txt` is that it's a kind of short cheat sheet, then that's valuable to humans too, and we should just create a cheat sheet type doc for LLMs and humans.

PR #3111, opened 2026-03-08
Fix zsh completion in nested modules

### LINKED ISSUE #2912: Shell completion does not work with modules
[Shell completions](https://just.systems/man/en/shell-completion-scripts.html) do not pick up recipes in [modules](https://just.systems/man/en/modules1190.html).

E.g., given these files:

```Justfile
# Justfile
mod hello

help:
        @just --list

example:
        echo this is an example rule
```

```Justfile
# hello.just
help:
        @just --list hello

world:
        echo hello world
```

If you want to run 'just hello world', completions will help you enter 'hello' but not 'world'.

(I found a [couple](https://github.com/casey/just/issues/2252#issuecomment-2663644735) [mentions](https://github.com/casey/just/pull/2113#issuecomment-2140408083) of this in other threads, but I wanted to file this dedicated issue for folks to subscribe to)

kitsuyui:
Fixes #2912.

This updates zsh completion so it can continue completing commands inside modules, instead of stopping after the first module segment.

What changed:
- resolve the current module path from the words already typed by the user
- use `just --list <module>` to complete names at the current module level
- detect fully-qualified recipe paths with `just --summary`
- add a zsh regression test covering nested modules

Before this change, completion could stop after the first module name in cases like the one described in #2912, where a command should continue completing deeper module recipes.

### DIFF
--- a/completions/just.zsh
+++ b/completions/just.zsh
@@ -100,19 +100,7 @@ _just() {
             curcontext="${curcontext%:*}-${words[2]}:"
 
             local lastarg=${words[${#words}]}
-            local recipe
-
-            local cmds; cmds=(
-                ${(s: :)$(_call_program commands just --summary)}
-            )
-
-            # Find first recipe name
-            for ((i = 2; i < $#words; i++ )) do
-                if [[ ${cmds[(I)${words[i]}]} -gt 0 ]]; then
-                    recipe=${words[i]}
-                    break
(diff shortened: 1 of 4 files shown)


casey:

The new completion engine will complete `::` separated recipe paths. It doesn't work with space-separated paths, although there's an item in #2406 for that.

PR #3112, opened 2026-03-08
Extend nested module completion fixes to zsh, bash, fish, PowerShell, and Nushell

kitsuyui:
This PR builds on the work in #3111 and further addresses #2912 for other shells.

The original zsh fix in #3111 resolved nested module completion by tracking the current module path and completing against the current module level instead of stopping after the first segment. This branch applies the same idea to additional shells that showed the same behavior.

Shells updated in this PR:
- bash
- fish
- PowerShell
- Nushell

What changed:
- bash: complete nested module names and recipes based on the current module path
- fish: complete nested module names and recipes with module-aware runtime completion
- PowerShell: resolve the current module path during completion and avoid invalid justfile-path handling
- Nushell: traverse nested modules from the current completion context instead of only looking at top-level recipes

Tests:
- added or extended regression coverage for bash, fish, PowerShell, and Nushell
- verified with `cargo test completions -- --nocapture`

This is intended as a follow-up to #3111, using the same approach for shells that still exhibited the module-completion problem described in #2912.


casey:

The new completion engine partially fixes this by completing `::` separated recipe paths, and there's a note in the above issue that we should make it work with space-separated paths.

PR #3115, opened 2026-03-09
Add attribute support for other BSD variants

jakewilliami:
Support BSD variants FreeBSD, Dragonfly, and NetBSD.  These are the remaining main BSD variants listed [in the Rust documentation for `target_os`](https://doc.rust-lang.org/reference/conditional-compilation.html#target_os).

See also #2497, which added support for OpenBSD.

This may also help towards a complete solution for #1682.

Supersedes #3072.


crates-io-readme.md
```diff
@@ -19,6 +19,6 @@ test TEST: build
 `just` produces detailed error messages and avoids `make`'s idiosyncrasies, so
 debugging a justfile is easier and less surprising than debugging a makefile.
 
-It works on Linux, MacOS, Windows, and BSD variants OpenBSD, FreeBSD, Dragonfly, and NetBSD.
+It works on all operating systems supported by Rust.
```
jakewilliami, line 22: I don't think this is _strictly_ true as we don't support iOS, Android, and other weird architectures&mdash;but it's true enough for most people. [\[1\]](https://doc.rust-lang.org/reference/conditional-compilation.html#r-cfg.target_os), [\[2\]](https://doc.rust-lang.org/beta/rustc/platform-support.html)

jakewilliami:
Thanks heaps for merging this @casey.  Really helpful for me, as I use `just` on my OpenBSD machine!  :sweat_smile:

Issue #2922, opened 2025-10-16
List just-recipes only from a group


woutervhy:
using just 1.40.0

I usually group my just-recipes per command:

```
[group: 'foo']
foo-install:
    ...

```

with `just list`  you can list all commands.
with more and more recipes,  this output becomes longer and longer.

I'm looking for an elegant way to list only the commands from a specific group.

Ideally something like a group-option when using the list:

```
just --list --group foo
```

casey:
Implemented in https://github.com/casey/just/pull/3117.

Issue #2662, opened 2025-03-11
just --list --group <group>


jabalsad:
Hello,

I think it would be nice to add an argument for only listing recipes for a given group.

From the documentation, I couldn't see that it is currently supported.

Thanks!

casey:
Implemented in #3117.

PR #3117, opened 2026-03-10
Add `--group` flag to filter `--list` output by group

### LINKED ISSUE #3109: Allow filter by group for recipes in --list
There is an ability to mark recipes by group(s):

```
[groups: 'deploy']
deploy_foo:
  ...
[groups: 'deploy']
deploy_bar:
  ...
[groups: "somethingelse']
do_unexpected:
 ...
```

It would be very convenient if we could extract this list of recipes for a specific group from Justfile by a simple query. At the moment Just does not allow to filter `--list` by group, only by submodule.

Proposals:

1. Allow filter by a group name (`just --list --group deploy`).
2. Allow to output in the json notation  (`--format json`).
3. Allow to output in the space separated format (`--format space`)

Expected output for `just --list --group deploy --format json`:

```
["deploy_foo", "deploy_bar"]
```
At the moment, the same result can be achieved with this jq magic:

```
just --dump --dump-format json | jq -c '[.recipes[] | select(.attributes[] | .group? == "deploy") | .name]'
```
But to my taste it's too much of jq, which make use of Just a bit less smooth in CI pipelines/workflows.

The primary usecase for this command would be forming a matrix for deployment (`matrix: $(just --list --group deploy --format space)`).  Additional jq here makes things crufty and hard to read.

terror:
Resolves https://github.com/casey/just/issues/3109, https://github.com/casey/just/issues/2662, https://github.com/casey/just/issues/2195, https://github.com/casey/just/issues/2922

This diff adds a `--group <GROUP>` flag to the `--list` subcommand that restricts output to recipes belonging to the specified group. The flag can be repeated to show multiple groups. For instance, `just --list --group deploy` shows only recipes annotated with `[group('deploy')]`, and `just --list --group deploy --group test` shows recipes in either group, each under its own group heading.


src/config.rs
```diff
@@ -26,6 +26,7 @@ pub(crate) struct Config {
   pub(crate) explain: bool,
   pub(crate) highlight: bool,
   pub(crate) invocation_directory: PathBuf,
+  pub(crate) list_groups: Vec<String>,
```
casey: I think I would call this `groups`, to more closely mirror the name of the flag.

```suggestion
  pub(crate) groups: Vec<String>,
```

src/config.rs
```diff
@@ -110,6 +111,7 @@ mod arg {
   pub(crate) const GLOBAL_JUSTFILE: &str = "GLOBAL-JUSTFILE";
   pub(crate) const HIGHLIGHT: &str = "HIGHLIGHT";
   pub(crate) const JUSTFILE: &str = "JUSTFILE";
+  pub(crate) const LIST_GROUP: &str = "LIST-GROUP";
```
```suggestion
  pub(crate) const GROUP: &str = "GROUP";
```

src/config.rs
```diff
@@ -316,6 +318,15 @@ impl Config {
           .action(ArgAction::SetTrue)
           .requires(cmd::LIST),
       )
+      .arg(
+        Arg::new(arg::LIST_GROUP)
```
```suggestion
        Arg::new(arg::GROUP)
```

src/config.rs
```diff
@@ -316,6 +318,15 @@ impl Config {
...
+          .long("group")
+          .env("JUST_GROUP")
+          .help("Only list recipes in <GROUP>")
+          .value_name("GROUP")
```
```suggestion
```

casey (changes requested):
If an argument to `--group` isn't actually a group, it should result in an error.

Currently, it actually prints out the group but with no recipes:

```console
: just --list --group barasdfasdf
Available recipes:
    [barasdfasdf]
```

src/subcommand.rs
```diff
@@ -478,12 +478,12 @@ impl Subcommand {
         })?;
     }
 
-    Self::list_module(config, module, 0);
+    Self::list_module(config, module, 0)?;
```
casey: I would pass in `config.groups` here at the top level, instead of doing the `depth==0` check, and then recursive calls pass in `&[]`.

casey (changes requested):

casey (approved):
LGTM!

PR #3118, opened 2026-03-12
Exclude omitted option defaults from positional arguments

AlexanderTheGrey:
This fixes an interaction between `set positional-arguments` and omitted defaulted option parameters.

Before this change, if a recipe option had a default value and the option was omitted, that default value was still injected into the shell positional argument list. This caused omitted option defaults to appear in `$@` for linewise and shebang recipes.

This change keeps omitted defaulted options available for parameter evaluation, but excludes them from positional shell arguments. Explicitly passed options already worked, so their behavior is unchanged.

Tests:
- Omitted defaulted options are not passed to positional arguments

### DIFF
--- a/src/evaluator.rs
+++ b/src/evaluator.rs
@@ -530,7 +530,9 @@ impl<'src, 'run> Evaluator<'src, 'run> {
       let values = if group.is_empty() {
         if let Some(ref default) = parameter.default {
           let value = evaluator.evaluate_expression(default)?;
-          positional.push(value.clone());
+          if !parameter.is_option() {
+            positional.push(value.clone());
+          }
           vec![value]
         } else if parameter.kind == ParameterKind::Star {
           Vec::new()
--- a/tests/positional_arguments.rs
+++ b/tests/positional_arguments.rs
@@ -162,6 +162,68 @@ fn default_arguments() {
 }
 
 #[test]
+fn omitted_default_long_option_is_not_passed_to_linewise_positional_arguments() {
+  Test::new()
+    .arg("foo")
+    .justfile(
+      r#"
+    set positional-arguments
+
+    [arg('bar', long='bar')]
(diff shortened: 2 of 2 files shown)


casey:
Thanks for the PR! It's honestly not clear to me that this is the right behavior. In the example test, `bar` has the value `BAR`, and it's surprising to me that it would be omitted from positional arguments. It's included in positional arguments when it isn't an option and has a default, and it's not clear to me that omitting it when it's an option is the right behavior.

In any case, this is a backwards-incompatible change, so at the very least it would need strong motivation.

AlexanderTheGrey:
You're right, I jumped the gun here. Sorry about that.

I re-tested and the inconsistency I thought I saw was caused by my own testing: I had accidentally left a `shift` when explicitly setting an option, then took it out when leaving the option set to a default value. I mistakenly thought the behavior was inconsistent (a shift happening internal to `just` in the explicitly-set case but not the default-set one).

Part of my motivation was that if `just` consumes an option itself, it feels natural to expect that option not to still appear in the positional arguments, especially when my use case is for the `just` recipe to handle that option and then forward only the remaining arguments to another script. But I can see why that behavior may not be desirable in other use cases.

casey:
No worries at all!

I definitely think that there is some weirdness here, and I'm not entirely sure how to resolve it. When a recipe takes arguments by position, it makes sense to then pass them by position. But when a recipe takes arguments by option/flag, then it could be ambiguous to pass them by position, because they're then missing the flags, and I can see why someone would want only the positional argument to be passed by position. But then you get into the weirdness I mentioned, where adding a option to an argument makes it disappear from the positional arguments array.

Probably the way to deal with this would be to add a key to the `arg` attribute that removes an argument from being passed on to positional arguments, but I would definitely want to understand the use-case before adding it.

PR #3123, opened 2026-03-14
Forward variables

### LINKED ISSUE #1762: Add a function to call `just` executable that preserves all CLI-passed parameters
I often have to run `just_executable()` while also passing all of the parameters passed by the user. As you can see in this example, forwarding parameters to sub-task is highly error-prone and cumbersome. Instead, I think it would be better to have a `just_forward()` function that automatically handles this param-escaping logic:

```justfile
param1 := "default for param1"
param2 := "default for param2"

nested_just := just_executable() + " param1=" + quote(param1) + " param2=" + quote(param2)

foo:
  # ...
  {{ nested_just }} bar ...
  # ...

bar:
  # ...
```

m9810223:
Closes #1762.

## Summary

Add two new built-in functions for forwarding CLI variable overrides:

- **`forward_variables(names...)`** — Expands command-line variable overrides into shell-safe `key='value'` pairs joined by spaces. Optionally accepts variable names to forward only specific overrides.

- **`forward_variables_with(sep, prefix, kvsep, names...)`** — Like `forward_variables`, but with customizable separator, prefix, and key-value separator. Useful for tools that expect specific flag formats (e.g., `psql -v`, `docker --build-arg`, `cmake -D`).

## Motivation

When a recipe invokes a child `just` process, there is currently no way to forward CLI variable overrides. For example:

```just
v1 := "x"
a:
    @echo {{ v1 }}
b: a
c:
    @{{ just_executable() }} a

# `just v1=y b` → y  (override applies within the same process)
# `just v1=y c` → x  (child process doesn't receive the override)
```

### Existing workarounds

**Manual `--set`** — tedious and not dynamic; must hard-code each variable:

```just
v1 := "x"
a:
    @echo {{ v1 }}
c:
    @{{ just_executable() }} --set v1 {{ v1 }} a

# `just v1=y c` → y  (works, but every variable must be listed manually)
```

**`set export` + `env_var_or_default()`** — requires modifying variable declarations in **both** parent and child justfiles:

```just
set export
v1 := env_var_or_default('v1', 'x')
a:
    @echo {{ v1 }}
c:
    @{{ just_executable() }} a

# `just v1=y c` → y  (works, but both justfiles must use env_var_or_default)
```

Neither approach can dynamically forward "whichever variables were overridden" without knowing them in advance. Variable resolution order (scope > assignment > env) means a child `just`'s `v1 := "x"` always wins over an exported environment variable.

### Solution

`forward_variables()` solves this cleanly — no child justfile changes needed:

```just
v1 := "x"
a:
    @echo {{ v1 }}
c:
    @{{ just_executable() }} {{ forward_variables() }} a

# `just v1=y c` → y  (override forwarded to child process)
```

## Examples

Forward all overrides to a child `just`:

```just
mode := 'dev'
build:
    @echo "Building in {{ mode }} mode"
run:
    {{ just_executable() }} {{ forward_variables() }} build

# just mode=release run
# → Building in release mode
```

Forward selected variables with a custom format:

```just
table_name := 'users'
uid := '42'
query:
    echo "psql {{ forward_variables_with(' ', '-v ', '=', 'table_name', 'uid') }} -f query.sql"

# just table_name=orders uid=100 query
# → psql -v table_name='orders' -v uid='100' -f query.sql
```

## Implementation

- Introduce **`NullaryPlus`** (0+ args) and **`TernaryPlus`** (3 required + variadic) function variants in `src/thunk.rs`.
- Implement `forward_variables()` and `forward_variables_with()` in `src/function.rs`, reading overrides from evaluator context.
- Add `overrides()` accessor to `Evaluator` via `src/evaluator.rs` and `src/variables.rs`.
- Comprehensive test coverage in `tests/forward_variables.rs` (531 lines): empty values, equals signs, special shell characters, single quote escaping, selective forwarding, and end-to-end child process test.
- Documentation added to both `README.md` and `README.中文.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1825,15 +1825,11 @@ foo := env('FOO', '') || 'DEFAULT_VALUE'
 
 #### Invocation Information
 
-- `is_dependency()` - Returns the string `true` if the current recipe is being
-  run as a dependency of another recipe, rather than being run directly,
-  otherwise returns the string `false`.
+- `is_dependency()` — Returns the string `true` if the current recipe is being run as a dependency of another recipe, rather than being run directly, otherwise returns the string `false`.
 
 #### Variable Forwarding
 
-- `forward_variables(names...)` - Returns command-line variable overrides as
-  shell-safe `key='value'` pairs, joined by spaces. If `names` are given, only
-  matching keys are included. Single quotes in values are properly escaped.
+- `forward_variables(names...)` — Returns command-line variable overrides as shell-safe `key='value'` pairs, joined by spaces. If `names` are given, only matching keys are included. Single quotes in values are properly escaped.
 
 Forward all overrides to a child `just` invocation:
 
@@ -1844,7 +1840,7 @@ build:
   @echo "Building in {{mode}} mode"
 
 run:
-  {{ just_executable() }} --justfile {{ justfile() }} {{ forward_variables() }} build
+  {{ just_executable() }} {{ forward_variables() }} build
 ```
 
 ```console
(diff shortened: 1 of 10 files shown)


m9810223:
Hi [@casey](https://github.com/casey), this PR adds `forward_variables()` and `forward_variables_with()` to enable forwarding CLI variable overrides to child `just` processes. Would love your feedback when you have a chance. Thanks!

PR #3132, opened 2026-03-15
Add alternative installation method via `taiki-e/install-action`

Danil42Russia:
Added the method first described in the documentation: https://github.com/taiki-e/install-action?tab=readme-ov-file#example-workflow

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,6 +416,16 @@ With [extractions/setup-just](https://github.com/extractions/setup-just):
 
 Or with [taiki-e/install-action](https://github.com/taiki-e/install-action):
 
+Use the generic installer interface in newer workflows:
+
+```yaml
+- uses: taiki-e/install-action@v2
+  with:
+    tool: just
+```
+
+Or use the dedicated `just` shortcut action:
+
 ```yaml
 - uses: taiki-e/install-action@just
 ```


casey:
Thanks for the PR! I don't really see the advantage of the longer version, assuming that they're equivalent.

Danil42Russia:
I’ve only just seen your reply :(

This is a safer way to install `just`. Especially now that the number of supply-chain attacks has increased. I actually forgot to mention that in my PR, but it can be included.

There’s even a thread on Zizmor suggesting that `unpinned-tools` should display a warning: https://github.com/zizmorcore/zizmor/issues/1775

taiki-e:
> This is a safer way to install `just`. Especially now that the number of supply-chain attacks has increased. I actually forgot to mention that in my PR, but it can be included.

The way used in this PR is equivalent to shorthand. As stated in the zizmor issue and the install-action README, fully pinning the action is necessary to enhance security against supply chain attacks.

https://github.com/taiki-e/install-action?tab=readme-ov-file#security

> The `@v<major>` and `@<tool_name>` tags are updated with each release. If you want to enhance workflow stability and security against supply chain attacks, consider using the `@v<major>.<minor>.<patch>` tag or their hash to pin the version and regularly updating with [dependency cooldown](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns). Since all releases are immutable, pinning the version in either way should have the same effect. Pinning `@<tool_name>` tags by hash is strongly discouraged, as it causes the workflow to reference a [commit that is not present on the repository](https://docs.zizmor.sh/audits/#impostor-commit) when a new version is released.
> 
> When installing the tool from GitHub Releases, the tool version that install-action installs with `tool: <tool_name>@latest` or `tool: <tool_name>@<omitted_version>` is associated with the install-action version, so pinning install-action version with the above ways also pins the version of the tool being installed. This also means that if a [dependency cooldown](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns) applies to the action itself, a cooldown of the same duration or a few days longer will apply to the tools installed by that action.

https://github.com/zizmorcore/zizmor/issues/1775#issuecomment-4148000423

> As for `taiki-e/install-action@v2`, if the tool is directly installed from GitHub releases ([list](https://github.com/taiki-e/install-action/blob/HEAD/TOOLS.md)), pinning the action version (`@v<major>.<minor>.<patch>`[^1] or `@<hash>` of it) will basically also pin the tool version. This is because ["latest" actually refers to the latest version in the manifest included in the action for checksum verification](https://github.com/taiki-e/install-action/blob/a164de717a0ee9284c2d9db1c6016a4c339cd333/manifests/zizmor.json#L21-L23). (This also means that if a cooldown applies to the action itself, a cooldown of the same duration or a few days longer will apply to the tools installed by that action.)

PR #3146, opened 2026-03-17
fix(fish-completion): stop treating command-f... as force flag

0xdeafbeef:
Fixes 
```
 just update-fast-selecerror: Failed to read justfile at `/home/<path>/ast-selec`: No such file or directory (os error 2)
 ```
 for recipes with -f in name


0xdeafbeef:
ok, 've seen same pr and i guess https://github.com/casey/just/pull/3154 will fix it

casey:
Thanks for the PR! Hopefully I can land #3154 soon, the patched completion scripts have been a nightmare to maintain.

PR #3154, opened 2026-03-19
Switch to clap's dynamic completion engine

casey:
This PR switches to clap's dynamic rust-native completion engine.

To do before merging:
- [ ] Avoid splitting arguments containing `:` in bash. https://github.com/clap-rs/clap/pull/6311
- [ ] Intelligent file fallback: We should only complete recipes for the first argument, but then complete both recipes and files afterwards.
- [ ] Are there any places where we should hint not to complete files?
- [ ] Add the completion scripts back to `completions/` in release packages
- [ ] Write tests for the custom functionality we're adding
- [ ] Make sure readme changes are correct

After merging:
- Close all PRs and issues related to the old completion scripts
- Create a new issue tracking improvements to the new completion scripts

Later:
- When completing `--set VARIABLE VALUE`, we will complete variable names for both VARIABLE and VALUE. We should only complete variable names for `VARIABLE` and complete files for `VALUE`.
- The patched `zsh` completion script has a nice feature where we should show `just --dump recipe` when tab completing an argument after the recipe, which was nice. This might be zsh specific functionality, but it would be nice if we could replicate it. 
- Should we show parameters on recipes with no docs?
- When completing a variable `VAR=`, don't put a space after `=` so the user can type the value without backspacing (`(-qS "=")` in zsh)
- When completing a recipe argument and the user presses tab, can we provide the parameter name as a hint?


casey:
Closing in favor of #3167.

PR #3185, opened 2026-03-24
Add relative_path built-in function

hunter-hongg:
Adds a new built-in function `relative_path` that computes a relative path from one path to another using `Path::strip_prefix`. This provides an OS-agnostic alternative to the `relpath` bash utility, which breaks on macOS.


src/function.rs
```diff
@@ -502,6 +503,22 @@ fn read(context: Context, filename: &str) -> FunctionResult {
     .map_err(|err| format!("I/O error reading `{filename}`: {err}"))
 }
 
+fn relpath(_context: Context, target: &str, base: &str) -> FunctionResult {
+  let base = match Utf8Path::new(base).canonicalize(){
```
casey, line 507: I'm not a fan of canonicalize unless necessary. I would join them both to the current working directory, lexiclean them, and then compute the relative path.

src/function.rs
```diff
@@ -502,6 +503,22 @@ fn read(context: Context, filename: &str) -> FunctionResult {
...
+    Ok(b) => b,
+    Err(e) => return Err(format!("Canonicalize path {base} failed: {e}")),
+  };
+  let target = match Utf8Path::new(target).canonicalize(){
+    Ok(t) => t,
+    Err(e) => return Err(format!("Canonicalize path {target} failed: {e}")),
+  };
+  if let Some(rel) = pathdiff::diff_paths(&target, &base) {
```
casey, line 515: I would prefer not to add a dependency for a single simple function.

src/function.rs
```diff
@@ -502,6 +503,22 @@ fn read(context: Context, filename: &str) -> FunctionResult {
...
+    Ok(rel.display().to_string())
+  } else {
+    Ok(target.display().to_string()) // fallback to absolute path
```
casey, line 518: It's not clear to me that it's useful to fall back to an absolute path. Is this desirable? Would an error be better?

.gitignore
```diff
@@ -15,3 +15,4 @@
 /test-utilities/Cargo.lock
 /test-utilities/target
 /tmp
+/3104issue.json
```
casey, line 18:
```suggestion
```

Cargo.toml
```diff
@@ -34,8 +33,9 @@ is_executable = "1.0.4"
 lexiclean = "0.0.1"
 libc = "0.2.0"
 num_cpus = "1.15.0"
+pathdiff = "0.2.3"
```
casey, line 36:
```suggestion
```

I'd rather not add a dependency for a single function, it should just be implemented in this PR.

casey (changes requested):

casey:
Also, I couldn't find which `relpath` you're referring to.

casey:

Issue #3188, opened 2026-03-24
Dynamic completion engine relies on completions from `just` binary in `$PATH`


tulilirockz:
I've been noticing on my system that every time I ran `sudo -s`, I got a weird `just` call. Turns out, it was because I had a newer version of just (1.48.0) installed via <https://brew.sh> than my system just (from [fedora linux](https://src.fedoraproject.org/rpms/rust-just/tree/f44)), this lead to the bash completions from the newer version calling out to the first version found in `$PATH`, which did not have the capabilities for generating the just completions automitically, leading to the broken `just` call every time I opened my shell.

I'm not particularly sure how to solve this issue, but heres a way to reproduce it:

```bash
#!/usr/bin/env bash

mkdir -p "oldbin" "newbin"
curl -fsSL https://github.com/casey/just/releases/download/1.48.0/just-1.48.0-x86_64-unknown-linux-musl.tar.gz | tar xzvf - -C newbin
curl -fsSL https://github.com/casey/just/releases/download/1.46.0/just-1.46.0-x86_64-unknown-linux-musl.tar.gz | tar xzvf - -C oldbin

export PATH="$PWD/oldbin:$PWD/newbin:$PATH"

chmod +x ./newbin/completions/just.bash
# This will fail and just call "`just` without any other thing and... not work!"
source <(./newbin/completions/just.bash)

# New binary is now on top
export PATH="$PWD/newbin:$PATH"

chmod +x ./newbin/completions/just.bash
# This will work now!
source <(./newbin/completions/just.bash)
```

Also a PR that would mitigate this. https://github.com/casey/just/pull/3189 - Unsure if this is wanted, though.

casey:
Thanks for reporting this, and sorry about the breakage!

#3189 will not I think work. Using `bash` as an example, with the alternative completion script:

```bash
source <(just --completions bash)
```

Once a new version of `just` is released, `source <(just --completions bash)` will then print out `source <(just --completions bash)` (that version will have the new completion script compiled into it) which will recursively continue to print out and source the same completion script.

We could make this work by doing:

```bash
source <(JUST_COMPLETE=bash just --completions bash)
```

Old versions would ignore `JUST_COMPLETE` and print out the old completion script, and new versions will print out the new completion script (since the hook that prints the new script when it sees `JUST_COMPLETE` happens before arguments are parse).

[...]

PR #3189, opened 2026-03-24
fix(completions): use `--completions` flag for compatibility with older just versions

tulilirockz:
Mitigates: https://github.com/casey/just/issues/3188


casey:
Thanks for the PR! This is a stopgap, but will break once a new version of `just` is released, see my comment here https://github.com/casey/just/issues/3188#issuecomment-4122105632.

PR #3193, opened 2026-03-25
Add checksum validation to install.sh

markusa380:
To be able to use just in production workflows without the risk of supply-chain attacks, a checksum verification of the downloaded archive is necessary.

### DIFF
--- a/www/install.sh
+++ b/www/install.sh
@@ -21,6 +21,7 @@ OPTIONS:
     --tag TAG       Tag (version) of the crate to install, defaults to latest release
     --to LOCATION   Where to install the binary [default: ~/bin]
     --target TARGET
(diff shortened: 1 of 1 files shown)


casey:
Can you be more specific about the use case? How are you planning on using this? Would you save the install.sh script locally? Or retrieve it using curl on-demand?

casey:
I'm going to close this. I don't fully understand the attack that this is trying to prevent. `install.sh` is downloaded from GitHub, and then downloads binaries from GitHub. So even if you pass a checksum when invoking `install.sh`, you're already trusting a script that you got from GitHub. You could download install.sh, inspect it, and use it with a checksum in the future, but it's unclear if that's what you're suggesting.

PR #3198, opened 2026-03-26
Fix `--set` missing value error message path interpolation

cuiweixie:
When `--set` was used without a value, the internal error string contained a literal `{path}` instead of the actual config path because the message was not formatted with `format!`.

This change interpolates `path` so the error is actionable.


casey:
Thanks! This fix is correct. However, in the normal case, it's impossible to see this error message. Did this error message surface somehow?

PR #3201, opened 2026-03-26
Support expressions in confirm attribute

begoon:
The confirmation prompt accepts any expression, so you can use variables,
concatenation, and function calls:

```just
target := "production"
[confirm("Deploy to " + target + "?")]
deploy:
  echo 'Deploying...'
```

Recipe parameters can also be used in confirmation prompts:

```just
[confirm("Deploy to " + env + "?")]
deploy env:
  echo 'Deploying to {{env}}...'
```

### DIFF
--- a/tests/directories.rs
+++ b/tests/directories.rs
@@ -5,7 +5,7 @@ fn cache_directory() {
   Test::new()
     .justfile("x := cache_directory()")
     .args(["--evaluate", "x"])
-    .stdout(dirs::cache_dir().unwrap_or_default().to_string_lossy())
+    .stdout(dirs::cache_dir().unwrap_or_default().to_str().unwrap())
     .success();
 }
 
@@ -14,7 +14,7 @@ fn config_directory() {
   Test::new()
     .justfile("x := config_directory()")
     .args(["--evaluate", "x"])
-    .stdout(dirs::config_dir().unwrap_or_default().to_string_lossy())
+    .stdout(dirs::config_dir().unwrap_or_default().to_str().unwrap())
     .success();
 }
 
@@ -26,7 +26,8 @@ fn config_local_directory() {
     .stdout(
       dirs::config_local_dir()
         .unwrap_or_default()
-        .to_string_lossy(),
+        .to_str()
+        .unwrap(),
     )
     .success();
 }
@@ -36,7 +37,7 @@ fn data_directory() {
   Test::new()
     .justfile("x := data_directory()")
     .args(["--evaluate", "x"])
-    .stdout(dirs::data_dir().unwrap_or_default().to_string_lossy())
+    .stdout(dirs::data_dir().unwrap_or_default().to_str().unwrap())
     .success();
 }
 
@@ -45,7 +46,7 @@ fn data_local_directory() {
   Test::new()
     .justfile("x := data_local_directory()")
     .args(["--evaluate", "x"])
-    .stdout(dirs::data_local_dir().unwrap_or_default().to_string_lossy())
+    .stdout(dirs::data_local_dir().unwrap_or_default().to_str().unwrap())
     .success();
 }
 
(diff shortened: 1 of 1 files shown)


begoon:
@casey

src/parser.rs
```diff
@@ -1437,45 +1437,62 @@ impl<'run, 'src> Parser<'run, 'src> {
       loop {
         let name = self.parse_name()?;
 
-        let mut arguments = Vec::new();
```
casey, line 1440: Parsing attributes is complex enough that I'd like to avoid this duplication. Can we always parse an expression, and then just complain if the expression is not a string literal unless it's the `confirm` attribute? That would avoid the duplication, and additionally make it easy to start accepting expressions in other attributes in the future.

src/justfile.rs
```diff
@@ -380,6 +374,12 @@ impl<'src> Justfile<'src> {
 
     let mut evaluator = Evaluator::new(&context, BTreeMap::new(), true, &scope);
 
+    if !config.yes && !recipe.confirm(&mut evaluator)? {
```
casey: Hmmmmmmm, I kind of don't like that we now have to evaluate the parameters before we confirm that we want to run the recipe. This is a change in behavior. I think it's acceptable though, since there is probably not much going on in parameters expressions.

src/attribute.rs
```diff
@@ -343,6 +346,36 @@ impl Display for Attribute<'_> {
   }
 }
 
+impl Eq for Attribute<'_> {}
```
casey: Why can't these be derived?

casey (changes requested):
Thanks for the PR! Left some comments.

src/parser.rs
begoon: @casey Addressed

src/attribute.rs
begoon: @casey Addressed

casey (changes requested):
It's possible to trigger an internal error by using a variable in the confirm expression which doesn't exist:

```
$ cat justfile
[confirm(x)]
foo:
$ cargo run foo
   Compiling just v1.48.1 (/Users/rodarmor/src/just)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.06s
     Running `target/debug/just foo`
error: Internal runtime error, this may indicate a bug in just: attempted to evaluate undefined variable `x` consider filing an issue: https://github.com/casey/just/issues/new
```

Confirm expressions should be analyzed for missing variables, see `UnresolvedRecipe::resolve`.

src/parser.rs
```diff
@@ -1437,45 +1437,65 @@ impl<'run, 'src> Parser<'run, 'src> {
...
-
-              if !keyword_arguments.is_empty() {
-                return Err(
-                  literal
-                    .token
-                    .error(CompileErrorKind::AttributePositionalFollowsKeyword),
-                );
+          if !self.next_is(ParenR) {
+            loop {
+              if !accepts_expressions
```
casey: Is this check necessary? Can we accept expression keyword arguments (distinguished by lookahead for an identifier followed by an equals sign), and then complain at the end if we got expression keyword arguments for attributes which don't accept them? (At this point all of them.)

begoon: @casey Addressed

casey (changes requested):
See this sketch for how attributes can accept expressions with less disruption to the parser code:

https://github.com/casey/just/commit/303b39b902e4c5fdad5753868bb9903fca5d2ce5

Inside of `Attribute::new`, you can then return an error if any argument is not a string literal, if the attribute is not `Arg`.

casey:
I fucked up T_T The branch was out of date so I attempted to update it by pushing, but I pushed the wrong branch, which auto-closed the PR. Could you re-open the PR? If possible, from a different branch than `master`, since that often causes problems with maintainers editing the PR. (For example, if a PR's base branch is `master`, then I can't use the "update branch" button.) Sorry for the hassle!

casey:
Also, on the off chance that you don't have the PR locally, I pushed a copy here: https://github.com/casey/just/tree/begoon/master

begoon:
After your push to "begoon/master", I cannot find any traces of my changes. I synced down locally, so locally files are "just/master", without my stuff.

begoon:
@casey I recovered from the reflog, so here is a new [PR](https://github.com/casey/just/pull/3238).

PR #3213, opened 2026-03-30
fix: Include module name in verbose messages

behnam-oneschema:
For a recipe `baz` under the module path of `foo::bar`, the current behavior of the verbose mode is to print lines like this:

```
===> Running recipe baz...
```

This change puts the full module path in those outputs, to improve visibility, as well as make it easy to re-trigger the same recipe with a copy-paste:

```
===> Running recipe foo::bar::baz...
```

### DIFF
--- a/src/recipe.rs
+++ b/src/recipe.rs
@@ -207,7 +207,7 @@ impl<'src> Recipe<'src> {
     let suffix = color.suffix();
 
     if context.config.verbosity.loquacious() {
-      eprintln!("{prefix}===> Running recipe `{self.recipe_path()}`...{suffix}");
+      eprintln!("{prefix}===> Running recipe `{}`...{suffix}", self.recipe_path());
     }
 
     if context.config.explain {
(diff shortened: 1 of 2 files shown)


casey (changes requested):
Thanks for the PR!

`recipe_path` is populated at the time this is printed, so you may call `self.recipe_path()` which unwraps internally.

Also, only the first test is necessary.

casey:
This branch is out-of-date, also, the formatting is wrong.

src/recipe.rs
```diff
@@ -207,7 +207,10 @@ impl<'src> Recipe<'src> {
     let suffix = color.suffix();
 
     if context.config.verbosity.loquacious() {
-      eprintln!("{prefix}===> Running recipe `{}`...{suffix}", self.name);
+      eprintln!(
+        "{prefix}===> Running recipe `{}`...{suffix}",
+        self.recipe_path()
```
```suggestion
        self.recipe_path(),
```

casey (changes requested):

src/recipe.rs
behnam-oneschema: Oh, it looks like we have different cargo settings. (I think mine is empty, though.)

Anyways, that's for making the change already!

PR #3221, opened 2026-03-31
Add `--evaluate-format just|shell`

### LINKED ISSUE #2990: Feature request: print shell-compatible variable evaluations
Hi there,

When running `just --evaluate`, or e.g. on a specific file `just --evaluate --file env.just`, to get the variables printed in such a way that they can be sourced / evaluated in the context of a shell.

For a justfile:
```
VAR := "foo"
export BAR := "bar"
```

Current behaviour:
```
$ just --evaluate
VAR := "foo"
BAR := "bar"
```

Desired behaviour (perhaps with an optional argument like `--shell`, or an entirely different command `--shell-eval`)
```
$ just --evaluate
VAR="foo"
export BAR="bar"
```

This allows me to run:
```
$ eval `just --evaluate`
```

casey:
Fixes #2990.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1477,6 +1477,23 @@ $ just --evaluate bob::bar::y
 hello
 ```
 
+The format of exported variables may be controlled with
+`--evaluate-format`<sup>master</sup>:
+
+```console
+$ just --evaluate --evaluate-format shell
+bar="world"
+foo="hello"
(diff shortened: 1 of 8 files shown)


jabalsad:
Amazing! Thank you.

PR #3223, opened 2026-03-31
docs: add navigation hint to Installation section

### LINKED ISSUE #2890: Main readme needs clarification for installation
It's very unclear from reading the Installation/Prerequisites section that we have to read ahead to get info for installing `just`.

The Installation/Prequisites section should lead in with something like "Scroll down to to the Packages section for links to the installer package for your environment."

Then, at the beginning of the Installation/Prerequisites/Windows section, lead in with something like "After installing `just`..."

Thank you.

akhilk27:
Closes #2890

New users reading the Installation section expect to find install commands right away, but the actual commands are under the Packages subsection further down. 

This adds two short sentences at the top of Installation pointing readers directly to Packages (and to 
Prerequisites for Windows users).

No content was moved or removed — this is a two-sentence additive change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,11 @@ always welcome!
 Installation
 ------------
 
+To install `just`, see the [Packages](#packages) section below for one-line
+install commands for your platform. On Windows, check
+[Prerequisites](#prerequisites) first for shell setup notes.
+
+
 ### Prerequisites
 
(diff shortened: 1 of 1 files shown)


casey:
Closing, this has LLM-voice.

PR #3225, opened 2026-04-01
Add glob wildcard support for import paths

### LINKED ISSUE #1885: support wildcard imports in justfiles

xuedi:
Closes #1885

## Summary

- Import paths may now contain glob wildcards (`*`, `?`, `[...]`)
- All matching files are imported in sorted (alphabetical) order
- A non-optional wildcard import that matches no files is an error
- An optional wildcard import (`import?`) that matches no files is silently ignored

## Example

```justfile
import 'plugins/*/justfile'
```

## Changes

- Added `glob` crate dependency
- Changed `Item::Import.absolute` to `Vec<PathBuf>` to hold multiple resolved paths
- Added glob detection and expansion in the compiler
- Added `GlobImportNoMatches` and `GlobPattern` error variants
- Added documentation and changelog entry
- Added integration tests

## Thoughts

I am not a hobby rust dev, i hope i got the types right, i know adding dependencies are never good, please give feedback, i just needed that for my own projects and thought i do a PR since maybe someone else think it useful.

No idea how this **workflow awaiting approval** works i would assume some runners got green after list tests and so on, but they dont, so i keep it in draft state. Feel free to delete or change :-)


casey:
This winds up being pretty tricky, given that we potentially need to match against non-unicode paths. See #2376,

PR #3226, opened 2026-04-01
Add runtime directory function

barskern:
Maybe there has been a conscious choice to leave this function out, if so, please do feel free to close this PR. I came across a usecase where I needed this function today so I thought it apt to do the PR myself as it is a very small and easy addition. My current workaround is to read the `XDG_RUNTIME_DIR` folder manually with a fallback as prescribed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2219,6 +2219,7 @@ details.
 - `data_local_directory()` - The local user-specific data directory.
 - `executable_directory()` - The user-specific executable directory.
 - `home_directory()` - The user's home directory.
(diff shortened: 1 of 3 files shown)


casey:
Nope, not a conscious choice at all! This was failing on MacOS and Windows, since the runtime directory is only defined on Linux. I added a note that it's only defined on Linux, and added a non-linux test which expects the function to fail.

casey:
Thanks for the PR!

PR #3241, opened 2026-04-03
Add just skill for agents

### LINKED ISSUE #2967: Add a skill for coding agents

casey:
Fixes #2967.

### DIFF
--- a/skills/just/SKILL.md
+++ b/skills/just/SKILL.md
@@ -1,0 +1,68 @@
+---
+name: just
+description: >
+  Reference for `just`, the command runner. Use when working in a project with
+  a `justfile` or when the user mentions `just` or `justfile`.
+---
+
+Just
+====
+
+Discovery
+---------
+
+- `just --dump` Print justfile
+- `just --evaluate` Print variable values
+- `just --help` Print detailed command-line syntax help
+- `just --list` Print recipes with descriptions
+- `just --show <RECIPE>` Print recipe source
+- `just --summary` Print recipes without descriptiosn
+
+Execution
+---------
+
+- `just` Run default recipe
+- `just <RECIPE>` Run specific recipe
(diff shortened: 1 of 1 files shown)


glaforge:
I'm wondering if a skill to also author `justfile`s could be possible?
That would help agents edit, extend, existing files, and know how to use `just`.
What do you think?

casey:
What would it help agents do that they either currently don't do, or do badly? LLMs are familiar with just from their pre-training, so they already know how to use it, and basic syntax.

glaforge:
From my limited attempts, indeed, it seems LLMs know `just` quite well. 
I was searching for an existing skill, but couldn't find one. But maybe it's not needed after all.
Please disregard my comment. 
And big thanks for `just`, I ❤️ it and it's awesome!

casey:
You are most welcome! I'm definitely open to adding skills or improving the existing skill, but I want to make sure I do it thoughtfully. It seems like a lot of the time, people add a bunch of random stuff to their prompts or AGENTS.md, and it's very hard to tell what works and what doesn't, and what problem is actually being solved.

Issue #3227, opened 2026-04-01
Pull requests are disabled


casey:
Pull requests on this repository have been disabled.

There are a bunch of reasons.

I've been getting a bunch of pull requests recently and finding it hard to keep up. Even when I am able to keep up with them, they have the effect of overriding my own priorities for what I'd like to work on. I'm not working on the things that I think are most important, I'm working on the thing that the author of the pull request thinks is most important.

And, of course, coding agents have changed the landscape enormously.
 
I've been getting a lot of agent-coded pull requests, which are, by and large, a net negative. When a pull request is simple enough to be coded by an agent, it's faster, easier, and more fun for me to prompt the agent myself. I can review and modify the code as I go and don't need to round-trip reviews with the contributor.

It also isn't clear to me if reviewing and merging pull requests by agents helps accomplish the meta goals of accepting pull requests in the first place, namely, giving people the opportunity to become better programmers, and helping people learn the codebase so they can make better contributions. Accepting pull requests written by agents probably mostly leads to minimal personal growth, and more pull requests written by agents.

There are also some reasons which are more specific to me. I'm very picky about code quality, and have very strong ideas about style and readability. Nontrivial pull requests often require substantial amounts of refactoring and review, even if they are functionally correct, for me to feel comfortable merging them. I also often don't know how the code should be until I try to write it myself and experiment, so this review in practice often means that I check the pull request out locally and iterate on the code until I understand the tradeoffs, which reduces the benefit of pull requests in the first place.

Since, at least at the moment, pull requests are a net negative, I've disabled them.

bluefing:
this makes complete sense and echos what I've been hearing about other open source projects. it's a sad state of affairs but completely understand your position. thanks for continuing to make just better.

starfishpatkhoo:
Wow, love your message! I'm going to start echoing what you just said to others! Ha ha ha!

For me, this especially, resonates:

> ... it's faster, easier, and more fun for me to prompt the agent myself. I can review and modify the code as I go ...

When I do this, often, I learn something new also.

PR #3289, opened 2026-04-11
Note that pull requests are not currently being accepted

casey:


jsskeen-utah:
@casey just curious (for posterity) - is this because of AI slop PRs or some other reason? It would be good to elaborate why, as this could be a deciding factor for others whether or not to use this tool.

casey:
See #3227, I laid out my reasoning there. I should probably either link to that issue in the readme, or include the explanation in the readme.

PR #3318, opened 2026-04-26
Allow expressions in `[working-directory]` attribute

casey:


casey:
This is complicated and has lots of tradeoffs, so closing for now.

PR #3550, opened 2026-07-06
Allow escaping CRLF in strings

casey:


laniakea64:
Not quite able to wrap my head around what this change is.  Might syntax highlighters like `vim-just` need to do anything to either adapt to this change or ensure it's being handled correctly?

casey:
This fixes a bug where you couldn't use line-continuations in a string in files with Windows CRLF line endings.

So this would fail, since you were doing `\` followed by `\r`, which before this change was not a valid escape character:

```
foo := "hello\
goodbye"
```

So this is a very minor behavior change. Syntax highlighters probably don't need to do anything.

